PR #7, opened 2021-10-13
Add `--compat-level` option to conda index

chenghlee:
This CEP would add an option to conda index to get it to generate repodata.json files that are compatible with older conda releases.


awwad:
Withdrawn

PR #8, opened 2021-10-14
Prepare menuinst for multiplatform support

jaimergp:
Standardizing the changes added in `menuinst@cep-devel` for other implementations. 

Please [refer to the rendered version of the proposal](https://github.com/jaimergp/ceps/blob/menuinst/cep-9999.md) to read comfortably :D


ericpre:
Thanks @jaimergp, this looks very promising.
From working on https://github.com/conda-forge/conda-standalone-feedstock/pull/18, your summary in https://gist.github.com/jaimergp/7de5843421d63fa4a408ac5c8712c3c9 and the current suggested actions of this PR looks very good to me, as this would disentangle the menuinst/conda-standalone/constructor situation and simplify adding new feature and making release!

Two suggestions to consider:
- Clarify the concept of "distribution name", which in the context of `menuinst` is relevant to define where to put the shortcuts in the start menu on windows. https://github.com/conda/menuinst/pull/77 allow to use the name of the folder where the distribution is installed, as a proxy but maybe it would be better to have the distribution name saved somewhere. `constructor` already have an [entry for the distribution name](https://github.com/conda/constructor/blob/master/CONSTRUCT.md#name).
- Add [pkg_names](https://github.com/conda/constructor/blob/f14fa91e1e363a029fffccb74c81618d27285bfb/constructor/nsis/_nsis.py#L107) argument to `menuinst`, which is necessary for [constructor](https://github.com/conda/constructor/blob/f14fa91e1e363a029fffccb74c81618d27285bfb/constructor/nsis/main.nsi.tmpl#L904) but I am not sure, if this is working as expected, i.e. if the [constructor menu_packages](https://github.com/conda/constructor/blob/master/CONSTRUCT.md#menu_packages) is honoured.

jaimergp:
Question: How to do cross-platform activation?

* Currently, on Windows, `menuinst` has `cwp.py`, which pseudo-activates the environment. It massages `PATH` and then launches specified command. 
  * Users call it like `PREFIX/python PREFIX/cwp.py PREFIX command args ...`
* On Unix-like platforms, this might not be enough because we also need LD_LIBRARY_PATH :/ We can use the same Python-depending approach and massage env vars and then relaunch in a subprocess.

However, this is not true activation at all. No post-activation scripts are executed, for example.


I was thinking of moving `cwp.py` to the `menuinst` namespace, maybe as `menuinst.activate`. So people can do `PREFIX/python -m menuinst.activate -- command args`. However, it requires Python and some projects might not use it for other stuff.

We can also expose this in the `conda-standalone` CLI so it fakes activation in each platform. But actually, `conda run` already does this for us! The only problem is that other self-contained implementations like micromamba would need to implement this as well to be constructor-compliant.

Thoughts? Maybe I am overthinking this, but the whole bootstrapping issue is hard to reason about 😬 

Cc @ericpre @wolfv

ericpre:
Sorry, I can't help on this, but I would be interested in the answer!

On a slightly different topic and maybe this out of the scope of this CEP: is it worth adding an option to define if the menu shortcut is for all users? This could useful for example to address https://github.com/mamba-org/mamba/issues/923.

wolfv:
@jaimergp is it feasible to create launch scripts (like python entrypoints)?

That could hard code the activation for a given command. E.g. the output of `micromamba shell activate`

Then the script could look like:

menuinst-napari.sh
```
#/usr/bin/env sh
. "/Users/wolfvollprecht/micromamba/etc/conda/deactivate.d/deactivate_clangxx_osx-arm64.sh"
. "/Users/wolfvollprecht/micromamba/etc/conda/deactivate.d/deactivate_clang_osx-arm64.sh"
. "/Users/wolfvollprecht/micromamba/etc/conda/deactivate.d/deactivate-gfortran_osx-arm64.sh"
PS1='(base) '
export PATH='/Users/wolfvollprecht/micromamba/bin:/Users/wolfvollprecht/micromamba/condabin:/Users/wolfvollprecht/.gem/ruby/3.0.0/bin:/opt/homebrew/opt/ruby/bin:/opt/homebrew/opt/ruby/gems/3.0.0/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/share/dotnet:~/.dotnet/tools:/Library/Apple/usr/bin:/Library/Frameworks/Mono.framework/Versions/Current/Commands'
export CONDA_SHLVL='1'
export CONDA_PROMPT_MODIFIER='(base) '
. "/Users/wolfvollprecht/micromamba/etc/conda/activate.d/activate-gfortran_osx-arm64.sh"
. "/Users/wolfvollprecht/micromamba/etc/conda/activate.d/activate_clang_osx-arm64.sh"
. "/Users/wolfvollprecht/micromamba/etc/conda/activate.d/activate_clangxx_osx-arm64.sh"

$CONDA_PREFIX/bin/napari
```

?

wolfv:
We would need to rewrite all these scripts for every modification of hte environment to take any additional post-activation scripts into account.

wolfv:
Or we write the equivalent bash / zsh function to call all activation scripts ...

jaimergp:
On MacOS we are using bash scripts to launch the programs, so we can add as much logic as needed. We didn't need that yet because we were relying on `python.app` to do the activation logic. See [here](https://github.com/jaimergp/menuinst/blob/fc9281e76de8163915ef9028f32910b5cf85fb0b/menuinst/platforms/osx.py#L93). On [Linux](https://github.com/jaimergp/menuinst/blob/cep/menuinst/platforms/linux.py#L174), it's currently a line on the `.desktop` file but I believe that's still executed by the shell directly, so we can do that as well.

Basically, I'd like to be able to source or include shell code to initialize the environment before the requested command is run. Whether the code comes from a file ad-hoc we create on the environment or we inject it in the shortcut file is more of an implementation detail.

On Windows, there's no script file or shell. The shortcut will run that command directly (I don't know the exact mechanism). We pseudoactivate with `python cwp.py`. cwp.py is provided by menuinst so it looks that this is being injected by constructor regardless, along with the whole python stack? In other words, on Windows, Python was already expected to be part of the target environment. It's a pity because `conda-standalone` _is_ a Python stack already, but we can't assume that's going to be always the case for other implementations like micromamba.

If we want to undo that requirement, we could use a `cmd`  call to add more logic. However, that will have a hanging CMD console in the background for no reason, unless there's a flag we can set in the shortcut file to prevent that? For example, subprocess on Windows allows you to add a [`CREATE_NO_WINDOW` flag](https://github.com/jaimergp/menuinst/blob/noconsole/cwp.py#L43). Maybe the shortcut can also be decorated with that flag? No clue.

jaimergp:
> is it worth adding an option to define if the menu shortcut is for all users or single user only?

I think that should depend on the installation we are creating a shortcut for.  I think there are `.nonadmin` files in Windows envs to mark this. We need to investigate how to detect this, but in my opinion it shouldn't be part of the metadata. At least that's how I am implementing it for now. Definitely worth investigating at a later stage, maybe after figuring out activation.

goanpeca:
This is looking great @jaimergp ! 🥳 thanks for working on this. 

What would be the next steps to get this moving forward?

Cheers!

wolfv:
I just read the bylaws and it sounds like you can start having a vote on the CEP:

> Enhancement Proposal Approval: When ready, the proposer may call for a vote on an existing conda enhancement proposal (EP). This requires a super-majority (60%) to pass so that the decision to accept the EP is unequivocal and we have ensured that consensus has been reached.
> Standard
> 60% Majority to pass

https://github.com/conda-incubator/governance

However, I am unsure which the mentioned `core` meetings are (maybe the conda community sync? maybe they don't exist right now?)

jaimergp:
Hey team, I have updated the CEP with the latest changes, as well as the PR description with links to _all_ the PRs involved in implementing the ideas.

cep-9999.md
```diff
@@ -0,0 +1,348 @@
...
+  `{{ PREFIX }}\Library\bin` on Windows.
+* `{{ SP_DIR }}`: `{{ PREFIX }}/lib/pythonX.Y/site-packages` on Unix. `{{ PREFIX }}\Lib\site-packages` on
+  Windows.
+* `{{ PY_VER }}`: The Python major version only. This is taken from the root prefix.
+* `{{ HOME }}`: Personal directory of the user.
+* `{{ ICON_EXT }}`: `ico` on Windows, `png` on Linux, `icns` on MacOS.
+
+Windows only:
+
+* `{{ SCRIPTS_DIR }}`: Scripts folder in Python environment, `{{ PREFIX }}/Scripts`.
```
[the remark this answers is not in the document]
jaimergp: fwiw, the current implementation allows shadowing global values with platform specific overrides.

jaimergp:
A bunch of updates to catch up with the latest changes in the implementation, and simplifying the messaging quite a bit. I think this is easier to read now and goes straight to the point.

cc @wolfv, in case you have comments for the mamba implementation.

jaimergp:
For the record, I plan to call a vote for this in two weeks (next conda community call), so if you have any comments before the vote, this is the time. Thanks! 🙏

jaimergp:
@conda-incubator/steering, I would like to call a vote for this CEP.

This vote falls under the **"Enhancement Proposal Approval"** policy of the [conda governance policy](https://github.com/conda-incubator/governance),

**This vote will end on August 3rd, 2023 End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

goanpeca (approved):
yes

jezdez:
yes

wolfv:
yes

marcelotrevisani (approved):

marcelotrevisani:
yes

xhochy (approved):
Yes

jaimergp:
yes

cep-9999.md
```diff
@@ -0,0 +1,380 @@
...
+`PYTHONW` | Path to the `pythonw.exe` executable in `PREFIX`.
+
+## Packaging guidelines
+
+`conda` packages that wish to create a shortcut at install time MUST provide a JSON file such that:
+
+- The JSON contents MUST pass schema validation.
+- The JSON file MUST be placed under `$PREFIX/Menu`.
+- The JSON file MUST include the package name in its filename. We recommend sticking to
+  `<pkg-name>.json` for simplicity.
```
chenghlee, line 157: Not going to make this a blocker, but is there a reason not to make this `MUST BE <pkg-name>.json`? Having to implement some regex or `*${pkg-name}*.json` glob seems unnecessarily complicated and error prone; e.g., what if `pkg-name` is a substring of some other package?

jaimergp: Current logic in `conda` [doesn't check the filename](https://github.com/conda/conda/blob/1b53c20dab60a262a87d21ed3609430117ff7e18/conda/core/path_actions.py#L569-L577). This is only used for filtering which packages should be processed from menuinst, which indeed [does check strictly for a full match](https://github.com/conda/constructor/blob/b6de413b2c40bf3caf7a809a1fbb08a8f1ee03a3/constructor/nsis/_nsis.py#L122-L136).

So you are right, this is better encoded as your suggestion. I'll fix it.

mariusvniekerk (approved):
Yes

msarahan (approved):
Yes

chenghlee (approved):
Yes

ocefpaf:
Yes

mbargull (approved):
yes

jakirkham (approved):
Yes

(guessing the filename will be updated with the CEP #)

jaimergp:
Hello team! The vote concluded last week on Aug 3rd at 23:59 AOE (Aug 4th 11:59 UTC). I have now reviewed the votes and these are the results:

* Yes: 12
* No: 0
* Abstain: 0
* Available quorum: [14](https://github.com/conda-incubator/governance/blob/9d40caa6a6a9ca889fbecce3a2f5b57653578662/steering.csv)
* Criteria: 60% of quorum
* 12/14 = 85.7%
* ✅ The vote has **passed**!

--

I will now update the PR to reflect the status and mint a CEP number. Since the vote for this PR was started earlier than #51, this has priority, and will be minted the **number 10**.

PR #10, opened 2021-12-01
Add cep for a new history format

wolfv:


chenghlee:
> Additionally we should formalize that a conda environment does _not_ require a history file to exist.

If we're going to do that, then we need to also define what makes some directory a conda environment. In the absence of any other standard, I've been using "if `foo/conda-meta/history` exists, then `foo` is by definition a conda environment". (Possibly one that's been tampered with, but a conda environment nonetheless.)

PR #15, opened 2022-01-19
Improve .conda support in conda-build

CheeseMochi:
CEP proposal to improve the support and documentation for creating .conda packages with conda-build


dholth:
I would suggest building version 2 `.conda` exclusively as a new default.

PR #19, opened 2022-04-25
CEP 6: Add Channel Notifications to conda

travishathaway:
This pull request contains a draft CEP for the implementation of a new "Channel Notifications" feature to the `conda` package manager. Please see the CEP itself for more information about the proposal.

Please use this pull request as a public forum for discussing the feature proposal as well as suggesting improvements.

### DIFF
--- a/cep-6.md
+++ b/cep-6.md
@@ -11,10 +11,10 @@
 
 ## Abstract
 
-In order to better facilitate better communicate between users of `conda` and channel owners, we wish to 
+In order to facilitate better communication between users of `conda` and channel owners, we want to 
 implement a notification feature.
 Specifically, this feature will give channel owners the ability to send users a small message (about the size 
-of a single Twitter post). How often this message displays will be set by channel owners,
+of a single Twitter post). How often this message displays will be set by channel owners
 and will be displayed for each channel a user is actively using (i.e. everything under "channels" in a
 user's `.condarc`).
 
@@ -28,9 +28,11 @@ The messages may contain but are not limited to the following use cases:
 
 ## Specification
 
-### When will show this message?
+The specification is described below via a question-answer format.
 
-The notification message will be limited to only appearing on the following commands:
+### When will this message be shown?
+
+The notification message will appear while running the following commands:
 
 - `create`
 - `env create`
@@ -41,7 +43,7 @@ The notification message will be limited to only appearing on the following comm
 
 The reasoning behind this decision is that the above commands all retrieve `repodata.json` from the configured
 channels. Simply adding another file request (which is very small and immediately cached) would not add much more
-overhead to the execution of these commands. Additionally, these are commands where the user already expects network
+overhead to the execution of these commands. Additionally, these are commands where our users already expect network
 traffic to occur, so the requirement for having an active internet connection is assumed.
 
 Here's an example of how a notification message may appear while running the `conda create` command:
@@ -67,12 +69,13 @@ Notice [info]:
 
 ```
 
-### How else can the user access this message?
+### How else can our users access this message?
 
-Additionally, because users may wish to see this message on demand, we will add a new sub-command called `motd`
-for doing just that. The following are a couple examples to show exactly how it would function:
+Additionally, because our users may wish to see this message on demand, we will add a new sub-command called `motd` or
+`notices` or `alerts` (exact name is yet to be determined) The following are a couple examples to show exactly how 
+it would function:
 
-**Basic usage:** grabs the MOTD for all current channels:
+**Basic usage:** grabs notifications for all current channels:
 
 ```
 $ conda (alerts|motd|notices)
@@ -90,7 +93,17 @@ Here is another message. It could have info about the latest happenings or blog 
 https://conda-forge.org/
 ```
 
-### What file format will this message have and what will it contain?
+**Show a single channel:** grabs notifications for a single channel:
+
+```
+$ conda (alerts|motd|notices) -c default
+
+Notice [info]:
+This is a test message. It is not very long, and could have a link to a longer post:
+https://example.com/short-link
+```
+
+### What file format will this message have, and what will it contain?
 
 The notification message will be in the JSON file format. This will allow us to not only store the message itself but 
 also metadata about the message, including information about how often the client should display the message (more on 
@@ -134,7 +147,7 @@ The motivation for this CEP came about as channel owners (Anaconda specifically)
 users of their channels. These messages may contain specific notices for particular users (e.g. identified by
 IP address) or general messages for the wider audience of a particular channel.
 
-Additionally, this new notification space can also provide a spot for us to relocate "conda update conda" reminders 
+Additionally, this new notification space can also provide a place for us to relocate `conda update conda` reminders 
 to a more visible spot (at the end of command output versus in the middle of the output). On top of this, other channels
 can use these notifications as a way to share news with their users or "calls for help" in maintaining their channels.
 
@@ -151,7 +164,7 @@ We do not expect any backwards compatibility issues for this new feature.
 ## Alternatives
 
 - **Show MOTD at the beginning of environment activation** This was deemed a little too intrusive/annoying.
-- **Show MOTD at the beginning of command output** Users may miss this if place here, especially for commands
+- **Show MOTD at the beginning of command output** Users may miss this if placed here, especially for commands
   with lots of output
 
 
--- a/cep-6.md
+++ b/cep-6.md
@@ -78,7 +78,7 @@ it would function:
 **Basic usage:** grabs notifications for all current channels:
 
 ```
-$ conda (alerts|motd|notices)
+$ conda alerts
 
 Channel: defaults
 
@@ -96,7 +96,7 @@ https://conda-forge.org/
 **Show a single channel:** grabs notifications for a single channel:
 
 ```
-$ conda (alerts|motd|notices) -c default
+$ conda alerts -c default
 
 Notice [info]:
 This is a test message. It is not very long, and could have a link to a longer post:
@@ -109,33 +109,38 @@ The notification message will be in the JSON file format. This will allow us to 
 also metadata about the message, including information about how often the client should display the message (more on 
 this in the next section).
 
-Here's an example of the `notifications.json` file which will be stored alongside files such as `repodata.json` on the
-channel servers:
+Here's an example of the `notifications.json` file which will be stored in the root of the channel directory structure.
 
 ```json
 {
   "notifications": [
     {
+      "id": "1cd1d8e5-d96c-42d1-9c29-e8120ad80823",
       "message": "Here is an example message that will be displayed to users",
       "level": "info",
-      "created_at": "2022-05-01 00:00:00",
-      "show_again": "7 days"
+      "created_at": "2022-04-26T11:50:34+00:00",
+      "expiry": 604800
     }
   ]
 }
 ```
 
-### How often will this message appear?
+Detailed overview of the JSON fields:
 
-How often the message appears will be configurable by the channel owners. This will be accomplished by several fields
-within the `notifications.json` file itself.  They are as follows:
+- **notifications** [Array] holds zero or more notifications that will be displayed to the client.
+  - **id** [String] unique ID for the message itself. UUIDs are preferred, but there is no required format.
+  - **message** [String] message that gets displayed to users.
+  - **level** [String] one of (info|warning|error). These will let our users know the category of the message
+    and will also allow the client to apply different formatting rules (e.g. text color).
+  - **created_at** [String] ISO 8601 formatted timestamp showing the creation time of the message.
+  - **expiry** [Number] a number specifying how long in seconds the message is valid for.
 
-- **created_at** Lets `conda` know how the old the message is
-- **show_again** plain english explanation of when to check for a new message; combining this with the previous field
-  allows us to know when to refresh our cache.
+### How often will these messages appear?
 
-The client will also have ultimate say over whether this message is displayed. We will provide clients with a setting
-to permanently disable these messages in their `.condarc` files:
+How often the messages appears will be configurable by the channel owners and the client. This will be accomplished by 
+the expiry field in the `notifications.json` file itself, but the client will the have ultimate say over whether this
+message is displayed. We will provide clients with a setting to permanently disable these messages in their `.condarc`
+files:
 
 ```yaml
 display_notifications: false
--- a/cep-6.md
+++ b/cep-6.md
@@ -160,7 +160,16 @@ can use these notices as a way to share news with their users or requests for he
 
 ## Rationale
 
-TBD
+In order to keep this feature as a simple as possible and reduce the number of moving parts, we choose to serve
+messages via a flat JSON file on the server. The client is responsible for caching these messages which helps
+to reduce the number of HTTP requests that it has to make. In designing such a system, it is our goal to keep
+the code complexity to a minimum while still affording as much flexibility as possible. In the future, certain
+channels may choose to dynamically serve this JSON file and this specification allows for that.
+
+The other consideration was keeping these messages to a minimum for our users and allowing them to easily disable
+these messages if they do not want to see them while running commands such as "install" or "create". Although
+we feel these messages are important for users to see, we also do not want to clutter their terminal output. 
+Ultimately, this is why we are choosing to turn this on by default but are also providing a way to disable it.
 
 
 ## Backwards Compatibility
--- a/cep-6.md
+++ b/cep-6.md
@@ -1,0 +1,197 @@
+<table>
+    <tr><td> Title </td><td>Add Channel Notifications to conda</td>
+    <tr><td> Status </td><td>Draft</td></tr>
+    <tr><td> Author(s) </td><td>Travis Hathaway &lt;thathaway@anaconda.com&gt;</td></tr>
+    <tr><td> Created </td><td> Apr 21, 2022</td></tr>
+    <tr><td> Updated </td><td> Aug 21, 2022</td></tr>
+    <tr><td> Discussion </td><td>NA</td></tr>
+    <tr><td> Implementation </td><td>NA</td></tr>
+</table>
+
+
+## Abstract
+
+In order to better facilitate better communicate between users of `conda` and channel owners, we wish to 
+implement a notification feature.
+Specifically, this feature will give channel owners the ability to send users a small message (about the size 
+of a single Twitter post). How often this message displays will be set by channel owners,
+and will be displayed for each channel a user is actively using (i.e. everything under "channels" in a
+user's `.condarc`).
+
+The messages may contain but are not limited to the following use cases:
+
+- News about the latest package additions to the channel
+- How to sponsor, volunteer or help out a particular channel
+- Warnings about a breach of TOS while using a particular channel
+- Status updates about the stability of the channel
+
+
+## Specification
+
+### When will show this message?
+
+The notification message will be limited to only appearing on the following commands:
+
+- `create`
+- `env create`
+- `install`
+- `update`
+- `env update`
+- `search`
+
+The reasoning behind this decision is that the above commands all retrieve `repodata.json` from the configured
+channels. Simply adding another file request (which is very small and immediately cached) would not add much more
+overhead to the execution of these commands. Additionally, these are commands where the user already expects network
+traffic to occur, so the requirement for having an active internet connection is assumed.
+
+Here's an example of how a notification message may appear while running the `conda create` command:
+
+```
+$ conda create -n geopandas geopandas
+Collecting package metadata (repodata.json): done
+Solving environment: done
+
+## Package Plan ##
+
+  environment location: /Users/username/opt/conda_x86_64/envs/geopandas
+
+  added / updated specs:
+    - geopandas
+
+... (output truncated)
+
+Notice [info]:
+> Here is a message to the user
+> Here is a link they could click: https://example.com/link-name
+> To see the message again, run `conda motd`
+
+```
+
+### How else can the user access this message?
+
+Additionally, because users may wish to see this message on demand, we will add a new sub-command called `motd`
+for doing just that. The following are a couple examples to show exactly how it would function:
+
+**Basic usage:** grabs the MOTD for all current channels:
+
+```
+$ conda (alerts|motd|notices)
+
+Channel: defaults
+
+Notice [info]:
+This is a test message. It is not very long, and could have a link to a longer post:
+https://example.com/short-link
+
+Channel: conda-forge
+
+Notice [info]:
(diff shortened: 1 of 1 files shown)


kenodegard (commented):
You offered several subcommand ideas `conda (alerts|motd|notices)`. For the sake of the `.condarc` config argument, we should probably avoid any aliasing and just pick one so we can keep consistent terminology throughout.

I vote for `alerts` since those are a well-defined concept from the web with understood severity tiers (e.g. [bootstrap](https://getbootstrap.com/docs/5.0/components/alerts/#examples) and [tailwindUI](https://tailwindui.com/components/application-ui/feedback/alerts)).

cep-6.md
```diff
@@ -0,0 +1,210 @@
...
+https://example.com/short-link
+```
+
+### What file format will this message have, and what will it contain?
+
+The notification message will be in the JSON file format. This will allow us to not only store the message itself but 
+also metadata about the message, including information about how often the client should display the message (more on 
+this in the next section).
+
+Here's an example of the `notifications.json` file which will be stored alongside files such as `repodata.json` on the
```
wolfv, line 112: Note that `repodata.json` is architecture specific.

wolfv:
In generally I think it could be interesting for channels to be able to contain more metadata. 

For example, if the `robostack` channel could easily advertise that it relies on the `conda-forge` channel to work, that could be cool.
A channel metadata could further have some fields for description etc.

cep-6.md
travishathaway: Thanks for pointing this out. I will correct this as we don't want this to architecture specific.

cep-6.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+### How often will this message appear?
+
+How often the message appears will be configurable by the channel owners. This will be accomplished by several fields
+within the `notifications.json` file itself.  They are as follows:
+
+- **created_at** Lets `conda` know how the old the message is
+- **show_again** plain english explanation of when to check for a new message; combining this with the previous field
```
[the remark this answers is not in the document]
travishathaway: So, this is really just an expiry like you point out. I can re-work this to reflect that.

Some kind of `id` also makes a lot of sense. I will update the CEP to include that and elaborate a little more on exactly how that would work client side.

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+## Backwards Compatibility
+
+We do not expect any backwards compatibility issues for this new feature.
+
+
+## Alternatives
+
+- **Show notifications at the beginning of environment activation** This was deemed a little too intrusive/annoying.
+- **Show notifications at the beginning of command output** Users may miss this if placed here, especially for commands
+  with lots of output
```
seibert, line 173: I'm not sure how many alternatives for message delivery we want to enumerate here, but for completeness, other options include:

* _Message in an HTTP header when retrieving any file from the repository_: This would be a better option for some kinds of messages, like download rate limiting or other blocks due to abuse, since it could be turned on by a rule in a CDN.  However, this use case is probably better addressed by having a standard way for conda to display errors on the console from HTTP status codes like 429 (Too Many Requests) and 403 (Forbidden).  Additionally, serving custom headers is challenging unless the repository owner has more control over the web server than is usually given by services like Github Pages.
* _Add notifications to the repodata.json file_: The repodata.json file is already being fetched, so adding some notifications would reduce the number of requests.  However, the repodata.json file is way too big already, so it would require clients to download a fairly large file before even being able to show the notification.
* _Create a generic metadata.json file, which contains a notifications key_: This is @wolfv 's suggestion, I think, and I do find it appealing for creating a space for other kinds of channel metadata.  The file should remain small, so it hopefully would not have the problems of putting notifications into repodata.json.  However, putting notifications into a generic metadata file would complicate dynamic generation of notifications (perhaps based on some client attribute, like IP address or something in the user agent) because the notifications would need to be combined with the other metadata.  For this reason, I think I still lean toward notifications being a separate file.

seibert:
I think another useful thing to specify in this CEP is the expectation that the client should fetch and show notifications *before* fetching or parsing the repodata.json file.  The reason for this is that if the repository is corrupted in some way, it would be nice for admins to quickly put up a notifications.json file explaining why and where to get more information while it is being fixed.  Could be similarly useful if a backward incompatible change to the repo format ever needs to be made. A clever web server configuration could serve a special notifications.json file to clients before a certain version in order to notify the user of the compatibility issue.

cep-6.md
LtDan33: Good thoughts. FYI there is a separate issue to make a better response to things like 403/429s. It was deemed that that (HTTP status codes) is more akin to better error reporting and something like this (channel notifications) would be better served under a different discussion.

LtDan33:
1. What is the thought on priority for channels? In theory if you are pulling from many channels you could have quite a few. Is there any limit to the number of alerts? 
3. It was also pointed out that the “expiry” naming could perhaps be a little clearer.  
5. Also I think I like the idea of having a dedicated notification json, do we want any qualifiers for specific client/conda metadata? For ex:

- Having an “arch” field where the alert would only show if its a windows user in the case for a specific issue with users on that platform for example
- Alert users of a conda version becoming EOL. We could add “conda_version” to target that. 
- A critical vulnerability for a user using a specific package.

There are always “what about…” so I’m all for keeping it simple and then expanding, but trying to think through other common and straightforward use cases that currently would add a lot of value. There are some implementation details on if there would be any parsing of the user agent to only return the “correct” alerts.json to the client, or if the server would throw back everything and let the client figure out what it wants to show.

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+and will be displayed for each channel a user is actively using (i.e. everything under "channels" in a
+user's `.condarc`).
+
+The messages may contain but are not limited to the following use cases:
+
+- News about the latest package additions to the channel
+- How to sponsor, volunteer or help out a particular channel
+- Warnings about a breach of TOS while using a particular channel
+- Status updates about the stability of the channel
+
```
jezdez, line 27: Maybe add "announcements for planned security releases"?

jezdez, line 27:
```suggestion
- Status updates about the stability of the channel
- Announcements for planned security releases
```

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+### When will this message be shown?
+
+The notification message will appear while running the following commands:
+
+- `create`
+- `env create`
+- `install`
+- `update`
+- `env update`
+- `search`
```
jezdez, line 42:
```suggestion
- `create`
- `env create`
- `env update`
- `install`
- `search`
- `update`
```

jezdez: I wonder if we should also do this for `conda remove` and `conda env remove` since technically they also more like install/update than anything else.

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+## Package Plan ##
+
+  environment location: /Users/username/opt/conda_x86_64/envs/geopandas
+
+  added / updated specs:
+    - geopandas
+
+... (output truncated)
+
+Notice [info]:
```
jezdez, line 65:
```suggestion
Channel notification (info):
```

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+> Here is a message to the user
+> Here is a link they could click: https://example.com/link-name
+> To see the message again, run `conda alert`
```
jezdez, line 68:
```suggestion

  Here is a message to the user
  Here is a link they could click: https://example.com/link-name
  To see the message again, run `conda alert`

End of channel notification.
```

Let's not use characters that may be a problem for screen readers for example

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+
+```
+
+### How else can our users access this message?
+
+Additionally, because our users may wish to see this message on demand, we will add a new sub-command called `alerts`.
```
jezdez, line 74:
```suggestion
Additionally, because our users may wish to see this message on demand, we will add a new sub-command called `notices`. 
```

I'll comment on this later more, but I think given the wide range of use cases for channel notifications we shouldn't call this subcommand `alerts` since it predefines a level of alarm that may not actually be matching the messages intentions.

I suggest to instead use (and rename the proposal to): `notices`

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+      "created_at": "2022-04-26T11:50:34+00:00",
+      "expiry": 604800
+    }
+  ]
+}
+```
+
+Detailed overview of the JSON fields:
+
+- **notifications** [Array] holds zero or more notifications that will be displayed to the client.
```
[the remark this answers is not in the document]
jezdez: I disagree that `alerts` is appropriate as a name since it predefines the messages as something to react to, something alarming, while the use cases above are by design left to the channel authors to decide whether they should be alarming or not. I'd suggest to use the neutral `notices`.

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+  - **id** [String] unique ID for the message itself. UUIDs are preferred, but there is no required format.
+  - **message** [String] message that gets displayed to users.
+  - **level** [String] one of (info|warning|error). These will let our users know the category of the message
```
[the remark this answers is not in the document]
jezdez: Hmm, sorry for nitpick, can we just call it `critical` since it's less alarming?

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+    and will also allow the client to apply different formatting rules (e.g. text color).
+  - **created_at** [String] ISO 8601 formatted timestamp showing the creation time of the message.
+  - **expiry** [Number] a number specifying how long in seconds the message is valid for.
```
jezdez, line 135:
```suggestion
  - **expiry** [Number] starting at `created_at`, a number specifying how long in seconds the message is valid for.
```

cep-6.md
```diff
@@ -0,0 +1,215 @@
...
+
+## Motivation
+
+The motivation for this CEP came about as channel owners (Anaconda specifically) needed a way to broadcast messages to
+users of their channels. These messages may contain specific notices for particular users (e.g. identified by
+IP address) or general messages for the wider audience of a particular channel.
+
+Additionally, this new notification space can also provide a place for us to relocate `conda update conda` reminders 
+to a more visible spot (at the end of command output versus in the middle of the output). On top of this, other channels
+can use these notifications as a way to share news with their users or "calls for help" in maintaining their channels.
```
jezdez, line 156:
```suggestion
can use these notifications as a way to share news with their users or requests for help in maintaining their channels.
```

jezdez (changes requested):
I like where this is going, and just have some wording advice: 
We need to be careful about framing this system as emergency-use only and use a more neutral word to communicate its diverse use cases.

As such I would recommend to (from the list of terms used in this text: "notifications", "alerts", "notices", "messages", "news") to stick to one that is a good compromise between length (because of the use as a subcommand) and neutrality. IMO that's "notices".

cep-6.md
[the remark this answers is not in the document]
travishathaway: @jezdez , I like your proposal. Let's go with that.

cep-6.md
travishathaway: I added these to the CEP (I changed the wording a bit on the last one though).

cep-6.md
```diff
@@ -0,0 +1,237 @@
...
+## Package Plan ##
+
+  environment location: /Users/username/opt/conda_x86_64/envs/geopandas
+
+  added / updated specs:
+    - geopandas
+
+... (output truncated)
+
+Channel notices (info):
```
kenodegard, line 65:
```suggestion
Channel (defaults) notices [info]:
```

cep-6.md
```diff
@@ -0,0 +1,237 @@
...
+The following are a couple examples to show exactly how it would function:
+
+**Basic usage:** grabs notices for all current channels:
+
+```
+$ conda notices
+
+Channel: defaults
+
+Notices [info]:
```
kenodegard, line 86:
```suggestion
Channel (defaults) notices [info]:
```

cep-6.md
```diff
@@ -0,0 +1,237 @@
...
+This is a test message. It is not very long, and could have a link to a longer post:
+https://example.com/short-link
+
+Channel: conda-forge
+
+Notices [info]:
```
kenodegard, line 92:
```suggestion
Channel (conda-forge) notices [info]:
```

cep-6.md
```diff
@@ -0,0 +1,237 @@
...
+Here is another message. It could have info about the latest happenings or blog posts from conda-forge:
+https://conda-forge.org/
+```
+
+**Show a single channel:** grabs notices for a single channel:
+
+```
+$ conda notices -c defaults
+
+Notices [info]:
```
kenodegard, line 102:
```suggestion
Channel (defaults) notices [info]:
```

kenodegard (commented):
Minor suggestion to use the same header whether a `display_notices` notice or if its `conda notices` notices. This should also make it easier to identify the notice's channel.

kenodegard:
> I suggest we rely on channel priority like you suggested. For example, we could only display the top five channel notifications and instruct our users to read the full output by running `conda notices`.

Let's assume a user has 5 channels defined (A-E). Channel A defines 2 critical notices and one info notice, channel B/C both define one info notice, and D/E both define one critical notice. What notices do we prioritize if we show the top 5 notices?

1. Display the most severe (and oldest) notice for the first 5 channels. Only one notice per channel.
  - A shows one **critical** notice
  - B shows one **info** notice
  - C shows one **info** notice
  - D shows one **critical** notice
  - E shows one **critical** notice
3. Sort all notices by severity followed by channel priority. Display the top 5 notices.
  - A shows two **critical** notice
  - D shows one **critical** notice
  - E shows one **critical** notice
  - B shows one **info** notice

I don't think either of these is better than the other so it's probably best to only display one notice? That way there's no confusion as to what notice would be displayed. We could just include a statement that there are additional notices, and to run `conda notices` to see all of these notices.

travishathaway:
@kenodegard You bring up an interesting point. I would err on the side of sorting by "priority", "channel order". We can play around with this in the initial implementation too.

In any case, we should definitely show a message instructing users if there are more unread messages. Something like:

```
5 of 12 new messages run `conda notices` to view all
```

LtDan33:
@kenodegard Good points. @travishathaway I like the thinking around making it explicit that there is value and a reason to run the command. Wordsmithing aside, I think we should make sure to call out any critical notices that aren't shown. Something like "X notices not shown, including y critical."  

I'd also make sure we are consistent with the naming throughout around messages/notices/etc and just use one everywhere. Ubiquitous language ftw 😄

travishathaway:
Hi all,

This is a recent video I have made demonstrating what this feature _could_ look like once is ready:

https://user-images.githubusercontent.com/718726/168361745-73adb527-8a13-4553-a906-1a51ae69b070.mp4

Looking forward to comments. If you have code specific comments, please take it over to the open PR I have current have for this: https://github.com/conda/conda/pull/11462

cep-6.md
```diff
@@ -0,0 +1,233 @@
...
+```json
+{
+  "notices": [
+    {
+      "id": "1cd1d8e5-d96c-42d1-9c29-e8120ad80823",
+      "message": "Here is an example message that will be displayed to users",
+      "level": "info",
+      "created_at": "2022-04-26T11:50:34+00:00",
+      "expiry": 604800,
+      "interval": 604800
```
jezdez, line 120:
```suggestion
      "expiry": 604800,
      "interval": 604800,
      
      "lang": "en"
```

Do you think it would be appropriate to add a `lang` attribute here to denote the language of the message in preparation for a future in which a server could respond in a user's preferred terminal language?

jezdez, line 120:
```suggestion
      "expiry": 604800,
      "interval": 604800,
      
      "lang": "en"
```

Do you think it would be appropriate to add a `lang` attribute here to denote the language of the message in preparation for a future in which a server could respond in a user's preferred terminal language?

cep-6.md
```diff
@@ -0,0 +1,233 @@
...
+IP address) or general messages for the wider audience of a particular channel.
+
+Additionally, this new notification space can also provide a place for us to relocate `conda update conda` reminders 
+to a more visible spot (at the end of command output versus in the middle of the output). On top of this, other channels
+can use these notices as a way to share news with their users or requests for help in maintaining their channels.
+
+
+## Rationale
+
+TBD
```
jezdez, line 166: It'd be good to fill in reasons why this was implemented in a client/server model or something like this. If there are no added details for design decisions you can also just skip it, of course.

jezdez, line 166: It'd be good to fill in reasons why this was implemented in a client/server model or something like this. If there are no added details for design decisions you can also just skip it, of course.

cep-6.md
travishathaway: Hi @jezdez, I think this would be appropriate to have, but I think this should be added in the future. Perhaps we could later make an update to this CEP when we wish to add multi-lingual support?

cep-6.md
travishathaway: Thanks for bringing this to my attention. I am going to add a couple paragraphs about why we made the decision to serve this via a static file on the channel servers and why we chose to add an option to disable notices from appearing while using conda.

jezdez:
@conda-incubator/steering

This PR falls under the **[Enhancement Proposal Approval](https://github.com/conda-incubator/governance#enhancement-proposal-approval)** policy of the [conda governance policy](https://github.com/conda-incubator/governance), please vote and/or comment on this PR.

This PR **needs 60% of the Steering Council** to vote yea to pass.

To vote please leave **Approve (yea)** or **Request Changes (nay)** pull request reviews.

If you would like changes to the current language please leave a comment ([in the PR](https://github.com/conda/ceps/pull/19)) or push to this branch.

**This vote will end on 2022-06-15.**

jezdez (approved):

cep-6.md
```diff
@@ -0,0 +1,252 @@
...
+## References
+
+This implementation is similar to the way that `npm` handles version update reminders:
+
+```
+npm install                                                                                                                   travishathaway@thath-work-laptop
+
+up to date, audited 365 packages in 3s
+
+34 packages are looking for funding
```
msarahan, line 209: we need a funding nag, too!

msarahan (approved):

ocefpaf (approved):

beckermr (approved):

cjmartian (approved):

cep-6.md
jezdez: Sounds good to me!

csoja (approved):

mariusvniekerk (approved):

cep-6.md
```diff
@@ -0,0 +1,252 @@
...
+
+Detailed overview of the JSON fields:
+
+- **notices** [Array] holds zero or more notices that will be displayed to the client.
+  - **id** [String] unique ID for the message itself. UUIDs are preferred, but there is no required format.
+  - **message** [String] message that gets displayed to users.
+  - **level** [String] one of (info|warning|critical). These will let our users know the category of the message
+    and will also allow the client to apply different formatting rules (e.g. text color).
+  - **created_at** [String] ISO 8601 formatted timestamp showing the creation time of the message.
+  - **expiry** [Number] starting at `created_at`, a number specifying how long in seconds the message is valid for.
```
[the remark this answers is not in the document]
travishathaway: Hi @chenghlee,

I believe this field initially started out as an "interval" field that would display the message every "x" seconds (seconds being the most granular interval). It then evolved into an `expiry` field, and I thought it would be acceptable to keep it as a number in seconds. On the Python side, I can just create this as a `timedelta` object and then easily calculate expiry time. Implementation here: https://github.com/travishathaway/conda/blob/1ab9deb123ae0d1987850345ea22175f344e31fc/conda/notices/cache.py#L55-L59

Now that I think about it though, creating this as a ISO 8601 would further simplify the logic for determining whether this message is expired....

What do you think? Is it worth implementing this as a timestamp field? The underlying code wouldn't be too difficult to update.

Thanks for your input!

mcg1969 (approved):

wolfv (approved):

xhochy (approved):

cep-6.md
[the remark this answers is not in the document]
awwad: > Personally, I find having both `created_at` and `expires_at` as ISO-8601 timestamps much easier to deal with. As an API consumer, having the expiry field as an actual timestamp makes it "obvious" what it is, whereas an integer forces me to look into the spec. Further, a timestamp would make it easier to consume the messages with tooling that doesn't have extensive support for datetime math (e.g., `curl url/to/notices.json | jq '.notices|.[]|[.message, .expires_at]' | ...`).

I also think expiry using ISO 8601 would be far preferable for simplicity reasons.  That's my only real qualm at the moment, so I guess that's a conditional yes from me. :)

awwad (changes requested):
(See comments above -- ISO 8601 expiry.  Otherwise, looks good to me.)

cep-6.md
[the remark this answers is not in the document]
travishathaway: Thanks for the feedback. I have updated the CEP accordingly 👍

travishathaway:
There were recent changes to this CEP. We have dropped the `expiry` field on individual message in exchange for an `expired_at` field. This new field will be set as an ISO 8601 timestamp string just like `created_at`.

ericdill (approved):

chenghlee (approved):

goanpeca (approved):

cep-6.md
```diff
@@ -0,0 +1,254 @@
...
+
+### Is this opt-in or opt-out?
+
+Users will automatically be opted in to these feature. They will have the ability to turn it off via
+a configuration parameter.
+
+
+## Resolution
+
+TBD
```
jakirkham, line 250: Is this meant to be replaced?

jakirkham (approved):
Generally approve. Though saw one leftover bit below that might need updating

cep-6.md
jezdez: This _might_ be replaced by the voting result, but it's optional, since the [governance policy only explicitly mentions](https://github.com/conda-incubator/governance/tree/69f3ce97b3b3e7001bf6bdece7d544f2dea3a633#posting-results) documenting the results in the pull request where the vote was conducted.

I would suggest to use the section if there would have been a controversial debate prior or during the vote. In this case I think it's fine to simply remove it when marking the CEP's status as "Accepted". Just my opinion though.

jakirkham: Interesting maybe we should only include this section when it is needed then?

jezdez: Yeah, that's not a bad idea, [CEP 1](https://github.com/conda-incubator/ceps/blob/main/cep-1.md#cep-content) uses "Some section that may be included if appropriate for the proposal include." so I mostly see this as part of the cleanup process before merging the proposal, in case the CEP author(s) have added it.

marcelotrevisani (approved):

PR #20, opened 2022-04-27
`.jlap` JSON Patch Incremental Updates for repodata.json

dholth:
Describes a system to save client bandwith on `conda install` when fetching `repodata.json` updates, based on patch series from specific earlier versions of the file.


dholth:
If we generate the patch file before it hits the CDN e.g. in `conda index`, then it can't include cache headers. The idea is that a paranoid client checks that `repodata.json`'s Last-Modified or ETag is within a certain range compared to the patches, in case a mirror stops providing the `.jlap` file but does not delete it.

cep-incremental-repodata.md
```diff
@@ -0,0 +1,289 @@
...
+appending new patches, a new metadata line and a new trailing checksum.
+
+When the client wants new data, it issues a single HTTP Range request from the
+bytes offset of the beginning of the penultimate `"metadata"` line, to the end
+of the file (`Range: bytes=<offset>-`), and re-verifies the trailing checksum.
+If the trailing checksum does not match the computed checksum then it must
+re-fetch the entire file; otherwise, it may apply the new patches.
+
+If the `.jlap` file represents part of a stream (earlier lines have been
+discarded) then the leading checksum is an intermediate checksum from that
```
kathatherine, line 279: Comma after ending parenthesis

cep-incremental-repodata.md
```diff
@@ -0,0 +1,289 @@
...
+stream. Otherwise the leading checksum is all `0`'s.
+
+## Reference
+
+* JSON Patch https://datatracker.ietf.org/doc/html/rfc6902
+* Server and client implementation in Python https://github.com/dholth/repodata-fly
```
kathatherine, line 285: The methods are all slightly beyond me, but the idea sounds really great. Being able to download only part of the repodata.json would be a great way to make updates more efficient.

kathatherine (changes requested):
Some minor punctuation edits. Thanks!

cep-incremental-repodata.md
```diff
@@ -73,10 +73,10 @@ this ~170MB (23MB compressed) file:
 The penultimate metadata line may include response headers for its corresponding
 `repodata.json`.
 
-When downloading new patches, a local cache service logs the number of lines,
-and number of bytes added, which is greater than the `Content-Encoding: gzip`
-compressed bytes transferred. (Ideally the feature would be integrated into
-`conda` itself instead of going through a local cache server):
+When downloading new patches, a local cache service logs the number of lines and
+number of bytes added (not the `Content-Encoding: gzip` compressed bytes
+transferred). Eventually the feature would be integrated into `conda` instead of
```
kathatherine, line 78: Comma after "Eventually" would also be appropriate.

cep-incremental-repodata.md
```diff
@@ -0,0 +1,289 @@
...
+37e1ba4f3bd8942a… → c9e035690378ca90…, 4 steps
+c9e035690378ca90… → 8c62fec64752f733…, 5 steps
+8c62fec64752f733… → 13604558bc1a5d65…, 1 steps
+13604558bc1a5d65… → 8d2b5e57d1c897b7…, 1 steps
+8d2b5e57d1c897b7… → 558651fbacc6e178…, 23 steps
+558651fbacc6e178… → a5661fb43183b877…, 4 steps
+a5661fb43183b877… → 300dc9b5cd8801d0…, 8 steps
+300dc9b5cd8801d0… → cd9fbab6f5af29da…, 24 steps
+cd9fbab6f5af29da… → 0aa315a3c31dd468…, 24 steps
+0aa315a3c31dd468… → bf2456c2e8ff30d5…, 31 steps
```
barabo, line 219: This series of transitions is probably longer than it needs to be to get the point across.  I would abbreviate it like this:

```suggestion
54506a213ba69d8f… → ...
                … → bf2456c2e8ff30d5…, 31 steps
```

cep-incremental-repodata.md
```diff
@@ -0,0 +1,289 @@
...
...{"url": "https://conda.anaconda.org/conda-forge/linux-64/repodata.json", "latest": "ea8a695e21baa8090691656bbe4a4d2240ee62cdc2539dc5940077ccf2a44efc", "headers": {"date": "Tue, 26 Apr 2022 18:14:12 GMT", "content-type": "application/json", "transfer-encoding": "chunked", "connection": "keep-alive", "cf-ray": "70214bf9bd722293-MIA", "age": "670", "cache-control": "public, max-age=1200", "etag": "W/\"30e1be6bf1b73889a41eaa4846a59c56-23\"", "expires": "Tue, 26 Apr 2022 18:34:12 GMT", "last-modified": "Tue, 26 Apr 2022 17:38:23 GMT", "vary": "Accept-Encoding", "cf-cache-status": "HIT", "expect-ct": "max-age=604800, report-uri=\"https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\"", "x-amz-id-2": "G2rn+4SD+sHXNxisSHtEbxHLulnvMh/Bj+KotOHzI0zxZDval2V8My839uHEfIC7LjtQRcnTOFA=", "x-amz-request-id": "7X74ZXZB2DMZA92F", "x-amz-version-id": "null", "set-cookie": "__cf_bm=.XMuRKJKwivDP9dELOOpA8NcS__lncLVi06qA6hKELM-1650996852-0-AaceO99Yc9N2lSPDz67hzvxp+BIrIYJE9Wj7HrfdzxL1xiTCx3fG6Cj/d0kODkOelFlXUp+eRnJ0uZE/oLxFysUZVVIu4GUx7gtisLdg5L6i; path=/; expires=Tue, 26-Apr-22 18:44:12 GMT; domain=.anaconda.org; HttpOnly; Secure; SameSite=None", "server": "cloudflare", "content-encoding": "gzip"}}
```
barabo, line 69: This would be easier for me to read for the first time if I didn't have to scroll to the right and try to mentally parse what's happening.  I would show this json beautified but add a comment elsewhere explaining that each update is listed on a line by itself, but is shown here expanded for clarity.

```suggestion
{
  "to": "ea8a695e21baa8090691656bbe4a4d2240ee62cdc2539dc5940077ccf2a44efc",
  "from": "fda691c5ff4e265b580b0dc43cbb0bd95af9920a10bacf621b51e5cfd72baa03",
  "patch": [
    {
      "op": "add",
      "path": "/packages/mapbox_earcut-1.0.0-py310hbf28c38_2.tar.bz2",
      "value": {
        "build": "py310hbf28c38_2",
        "build_number": 2,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "numpy",
          "python >=3.10,<3.11.0a0",
          "python_abi 3.10.* *_cp310"
        ],
        "license": "ISC AND BSD-3-Clause",
        "md5": "93fc1b1a1e34d2b21fa2136d06aa6c89",
        "name": "mapbox_earcut",
        "sha256": "acaca6408c049b0575666614aa9a50f628f345a0b3b7d36ace8d55b2f005ff4d",
        "size": 80917,
        "subdir": "linux-64",
        "timestamp": 1650989008805,
        "version": "1.0.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/mapbox_earcut-1.0.0-py37h506a98e_2.tar.bz2",
      "value": {
        "build": "py37h506a98e_2",
        "build_number": 2,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "numpy",
          "pypy3.7 >=7.3.7",
          "python >=3.7,<3.8.0a0",
          "python_abi 3.7 *_pypy37_pp73"
        ],
        "license": "ISC AND BSD-3-Clause",
        "md5": "b486d2bf0b9cb6f9fc9caf29aced70cf",
        "name": "mapbox_earcut",
        "sha256": "50242c2aff0fa007135e18b32c540b57460c53ca225f3ee182c13d1e13c7ed09",
        "size": 80007,
        "subdir": "linux-64",
        "timestamp": 1650988961306,
        "version": "1.0.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/mapbox_earcut-1.0.0-py37h7cecad7_2.tar.bz2",
      "value": {
        "build": "py37h7cecad7_2",
        "build_number": 2,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "numpy",
          "python >=3.7,<3.8.0a0",
          "python_abi 3.7.* *_cp37m"
        ],
        "license": "ISC AND BSD-3-Clause",
        "md5": "a6227836fd9dc731a61ac0ed57589790",
        "name": "mapbox_earcut",
        "sha256": "050b3d269a0964c8bd814b2bdcf9f5c523225c3f602a18e4c04e1049438d4826",
        "size": 79206,
        "subdir": "linux-64",
        "timestamp": 1650988985131,
        "version": "1.0.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/mapbox_earcut-1.0.0-py38h43d8883_2.tar.bz2",
      "value": {
        "build": "py38h43d8883_2",
        "build_number": 2,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "numpy",
          "python >=3.8,<3.9.0a0",
          "python_abi 3.8.* *_cp38"
        ],
        "license": "ISC AND BSD-3-Clause",
        "md5": "26ff06f1842618cd17ab7c99dc7a8a95",
        "name": "mapbox_earcut",
        "sha256": "d5e81828bf1cdc6e2105362cf867cb217d7fa38519628936d1a254b0be154bfc",
        "size": 80431,
        "subdir": "linux-64",
        "timestamp": 1650988985479,
        "version": "1.0.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/mapbox_earcut-1.0.0-py39hf939315_2.tar.bz2",
      "value": {
        "build": "py39hf939315_2",
        "build_number": 2,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "numpy",
          "python >=3.9,<3.10.0a0",
          "python_abi 3.9.* *_cp39"
        ],
        "license": "ISC AND BSD-3-Clause",
        "md5": "4770c3dfb3b34f390e57d1fa3cf208f3",
        "name": "mapbox_earcut",
        "sha256": "6941e8d18d283b35e84596014c7ef8bca317ec320ae765eded8ad67fe9d9d0c7",
        "size": 80689,
        "subdir": "linux-64",
        "timestamp": 1650989007616,
        "version": "1.0.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/poco-1.11.2-h08a2579_0.tar.bz2",
      "value": {
        "build": "h08a2579_0",
        "build_number": 0,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "openssl >=3.0.2,<4.0a0",
          "unixodbc >=2.3.9,<2.4.0a0"
        ],
        "license": "BSL-1.0",
        "license_family": "OTHER",
        "md5": "e1a33760d40c42bb21f10ebbf79fbd24",
        "name": "poco",
        "sha256": "6525ab8d6089481938e298dc50f8ca79dbd39ad3e616cbbd3daa9a9fb2cec295",
        "size": 5332156,
        "subdir": "linux-64",
        "timestamp": 1650989286660,
        "version": "1.11.2"
      }
    },
    {
      "op": "add",
      "path": "/packages/poco-1.11.2-h3790be6_0.tar.bz2",
      "value": {
        "build": "h3790be6_0",
        "build_number": 0,
        "depends": [
          "libgcc-ng >=10.3.0",
          "libstdcxx-ng >=10.3.0",
          "openssl >=1.1.1n,<1.1.2a",
          "unixodbc >=2.3.9,<2.4.0a0"
        ],
        "license": "BSL-1.0",
        "license_family": "OTHER",
        "md5": "5af5066afe0a90be9a2f95d92f3b225e",
        "name": "poco",
        "sha256": "bc86251c0a0f86146423996b8b8db30358a61d7f911c2322fe70ca2df7d752a8",
        "size": 5225400,
        "subdir": "linux-64",
        "timestamp": 1650989278591,
        "version": "1.11.2"
      }
    },
    {
      "op": "add",
      "path": "/packages/poppler-22.04.0-h1434ded_0.tar.bz2",
      "value": {
        "build": "h1434ded_0",
        "build_number": 0,
        "depends": [
          "boost-cpp >=1.74.0,<1.74.1.0a0",
          "cairo >=1.16.0,<1.17.0a0",
          "fontconfig >=2.13.96,<3.0a0",
          "fonts-conda-ecosystem",
          "freetype >=2.10.4,<3.0a0",
          "gettext >=0.19.8.1,<1.0a0",
          "jpeg >=9e,<10a",
          "lcms2 >=2.12,<3.0a0",
          "libcurl >=7.82.0,<8.0a0",
          "libgcc-ng >=10.3.0",
          "libglib >=2.70.2,<3.0a0",
          "libiconv >=1.16,<1.17.0a0",
          "libpng >=1.6.37,<1.7.0a0",
          "libstdcxx-ng >=10.3.0",
          "libtiff >=4.3.0,<5.0a0",
          "libzlib >=1.2.11,<1.3.0a0",
          "nss >=3.77,<4.0a0",
          "openjpeg >=2.4.0,<2.5.0a0",
          "poppler-data"
        ],
        "license": "GPL-2.0-only",
        "license_family": "GPL",
        "md5": "864687071a99df6f24dd587b77e4c753",
        "name": "poppler",
        "sha256": "53593388bb16933c2c711d0fb8e46b6cf6f429acb19a204b6e1cd6eaf1646288",
        "size": 20191738,
        "subdir": "linux-64",
        "timestamp": 1650986618544,
        "version": "22.04.0"
      }
    },
    {
      "op": "add",
      "path": "/packages/poppler-qt-22.04.0-hf358c63_0.tar.bz2",
      "value": {
        "build": "hf358c63_0",
        "build_number": 0,
        "depends": [
          "cairo >=1.16.0,<1.17.0a0",
          "fontconfig >=2.13.96,<3.0a0",
          "fonts-conda-ecosystem",
          "freetype >=2.10.4,<3.0a0",
          "gettext >=0.19.8.1,<1.0a0",
          "jpeg >=9e,<10a",
          "lcms2 >=2.12,<3.0a0",
          "libcurl >=7.82.0,<8.0a0",
          "libgcc-ng >=10.3.0",
          "libglib >=2.70.2,<3.0a0",
          "libiconv >=1.16,<1.17.0a0",
          "libpng >=1.6.37,<1.7.0a0",
          "libstdcxx-ng >=10.3.0",
          "libtiff >=4.3.0,<5.0a0",
          "libzlib >=1.2.11,<1.3.0a0",
          "nss >=3.77,<4.0a0",
          "openjpeg >=2.4.0,<2.5.0a0",
          "poppler 22.04.0 h1434ded_0",
          "qt >=5.12.9,<5.13.0a0"
        ],
        "license": "GPL-2.0-only",
        "license_family": "GPL",
        "md5": "33b3a474e4ecc5ab4e21c2f7912caf3d",
        "name": "poppler-qt",
        "sha256": "aa2c23a362c13837edf0763c77c98c97a3412640339924fa17c24f577d63b0ec",
        "size": 5553975,
        "subdir": "linux-64",
        "timestamp": 1650986746308,
        "version": "22.04.0"
      }
    }
  ]
}
{
  "url": "https://conda.anaconda.org/conda-forge/linux-64/repodata.json",
  "latest": "ea8a695e21baa8090691656bbe4a4d2240ee62cdc2539dc5940077ccf2a44efc",
  "headers": {
    "date": "Tue, 26 Apr 2022 18:14:12 GMT",
    "content-type": "application/json",
    "transfer-encoding": "chunked",
    "connection": "keep-alive",
    "cf-ray": "70214bf9bd722293-MIA",
    "age": "670",
    "cache-control": "public, max-age=1200",
    "etag": "W/\"30e1be6bf1b73889a41eaa4846a59c56-23\"",
    "expires": "Tue, 26 Apr 2022 18:34:12 GMT",
    "last-modified": "Tue, 26 Apr 2022 17:38:23 GMT",
    "vary": "Accept-Encoding",
    "cf-cache-status": "HIT",
    "expect-ct": "max-age=604800, report-uri=\"https://report-uri.cloudflare.com/cdn-cgi/beacon/expect-ct\"",
    "x-amz-id-2": "G2rn+4SD+sHXNxisSHtEbxHLulnvMh/Bj+KotOHzI0zxZDval2V8My839uHEfIC7LjtQRcnTOFA=",
    "x-amz-request-id": "7X74ZXZB2DMZA92F",
    "x-amz-version-id": "null",
    "set-cookie": "__cf_bm=.XMuRKJKwivDP9dELOOpA8NcS__lncLVi06qA6hKELM-1650996852-0-AaceO99Yc9N2lSPDz67hzvxp+BIrIYJE9Wj7HrfdzxL1xiTCx3fG6Cj/d0kODkOelFlXUp+eRnJ0uZE/oLxFysUZVVIu4GUx7gtisLdg5L6i; path=/; expires=Tue, 26-Apr-22 18:44:12 GMT; domain=.anaconda.org; HttpOnly; Secure; SameSite=None",
    "server": "cloudflare",
    "content-encoding": "gzip"
  }
}
```

barabo (commented):
I haven't given this a thorough read-though yet, but here are a couple comments right away to consider.  I would also suggest you think about providing a smaller / simpler example of the jlap patch - maybe include 1 or 2 patches rather than 9.

If there are properties in the patch that are extraneous, consider cutting them out, too.

cep-incremental-repodata.md
dholth: Thanks. The point was supposed to be "it is pretty fast". Maybe we can just say that.

cep-incremental-repodata.md
```diff
@@ -279,6 +133,100 @@ If the `.jlap` file represents part of a stream (earlier lines have been
...
+
+{
+    "to": current_digest.hex()
+    "from": previous_digest.hex(),
+    "patch": [] # RFC 6902 JSON Patch
+}
+```
+
+The metadata object gives the digest of the latest version, and any other
+metadata especially the `url`.
```
kathatherine, line 156:
```suggestion
metadata, especially the `url`.
```

kathatherine (changes requested):
Just a few punctuation/grammar clarifications.

ocefpaf:
@dholth I think this is a great idea and I recall Phil Elson wanting to implement something like this way back in 2018 but never got to it. I'm definitely+1.

The only confusion I have is: should this be a CEP? It is a desired improvement but does it have any operational impacts that warrants voting? I may be missing something (read the doc on my mobile phone) but I did not see any reason for this to be a CEP. I'd love to see the PR that implements this merged though.

beckermr (changes requested):
If we're voting now, my answer is no. I don't think we've reached a consensus on how this format might interact with the power loader stuff from Wolf and possible optimizations for repodata sharding and patching. 

I think if we work some of those details out, and if we have we should put that in the CEP (or I missed it!, then we could and should reconsider this.

dholth:
I like the idea of being able to implement this in conda without worrying about CEPs. It does not change the format of repodata.json at all.

I was able to meet Wolf in July. The C++ json library they are using implements jsonpatch. Mamba still wants to use the same on-disk http cache as conda, which is a question of paths. The CEP does mention other patching schemes, like the one RPM uses that is generic on bytes instead of generic on json, but is only a C library and is not as optimized for HTTP round trips.

We will be able to get fantastic network performance in Python conda without powerloader.

beckermr:
That's all great and I support it. Let's try it live in conda and with anaconda.org. We'll learn a lot. 

I simply don't want to see us declaring standards on this for conda just yet.

wolfv:
I've started the implementation of the jlap format in C++ as well (https://github.com/mamba-org/mamba/pull/2029) – I don't think it's too difficult and I think we'd be happy to support it with mamba (post-1.0) as well.

- Did the conda project decide on a new way to store HTTP metadata for `repodata.json` / `repodata.jlap` downloads? We've discussed to store etag / last-modified headers and such in a SQLite database which I was (and still am) slightly against because that would mean an additional dependency for `micromamba`. Even though this is only tangentially related to this CEP, it might be a byproduct of the implementation since you probably want to be able to directly compute the hash of the repodata.json file without modifying its contents. I'd be curious to hear your plans for this.
- I am not sure if the tooling to generate `jlap` files is open source / available?

dholth:
There are multiple jlap generators in the Implementation link in the CEP. `json2jlap.py`

In the conda branch, there is a separate `(hash).state.json` which stores the mtime, actual hash, and upstream hash of repodata.json, as well as headers from the last fetch, an offset for how far we are in the jlap etc. I don't like this because I'm trying to implement "update two files transactionally" without sqlite but it works. Of course mtime and etag go in the state file and not in repodata.json.

Conda should provide an API for retrieving repodata.json instead of multiple software looking inside its cache. We have this problem with Anaconda Navigator as well.

@wolv I read your comment about having a state machine, and I feel it too. My approach is to handle the common cases but fall back to downloading the entire repodata.json, even if a more complicated implementation could avoid the full download.

wolfv:
@dholth thanks for the answer. Should we spec out the metadata format and URL hash scheme as well? Should that be in this CEP or another?

I think an API to access these repodata from the Python side would be nice, indeed, however, `micromamba` also shares the cache and is pure C++ so it wouldn't be useful in that case.

We can also decide that we create different repodata caches for mamba & conda but I think the shared cache has worked fine in the past.

wolfv:
Also was the switch to 5 base32 encoded bytes instead of 4 base16 encoded bytes random or deliberate? Is there any other reasoning behind it except to make it not collide with the previous hashes / format?

dholth:
It's random and only to make it not collide. Of course the on-disk format is not part of the CEP.

wolfv:
Would gzip buy much vs. just pickle?

I guess the downside of the pickle format is that json-patch won't work as straight forwardly (if there isn't a fast way to restore the original JSON from it).

I've also thought about storing `jlap-availability` in the additional metadata.

Not sure I like "versioning" repodata storage by randomly changing hashes. We could do something more deliberate, like `<hash>.v2.json` or something.

dholth:
Pickle is the conda objects like SubdirData which do not translate back into the exact server-sent json.

It would be interesting to pickle the unmodified json. Of course that would not do much for micromamba.

mariusvniekerk:
Please use faster compression formats if you're using any of them. Gzip and bzip2 are laughably slow at decompression compared to zstd.  
Our files are big enough that decompression speed matters.

dholth:
@mariusvniekerk we will benchmark the whole thing, and we have zstandard available because of `.conda` https://dholth.github.io/conda-benchmarks/#conda_install.TimeInstall.time_explicit_install?conda-package-handling=1.9.0&conda-package-handling=2.0.0a2&p-threads=1&p-threads=5

dholth:
This feature is available in conda 23.3.1 with the `"--experimental=jlap"` flag. It is the most impressive after the third cache expiration - on an empty cache, we fetch repodata.json; the first jlap fetch, fetches the entire remote jlap file; the third fetches only the end of the remote jlap file.

dholth:
@wolfv @ocefpaf @beckermr @mariusvniekerk Maybe you are not all still voting members, but how do you feel about this CEP now that there is a complete implementation?

cep-jlap.md
```diff
@@ -0,0 +1,247 @@
+<table>
+<tr><td> Title </td><td> JSON Patch Incremental Updates for repodata.json </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Daniel Holth &lt;dholth@gmail.com | dholth@anaconda.com&gt;</td></tr>
+<tr><td> Created </td><td> Mar 30, 2022</td></tr>
+<tr><td> Updated </td><td> Oct 11, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> https://github.com/dholth/repodata-fly </td></tr>
```
dholth, line 8: Might as well link to the implementations in conda and librattler. conda's jlap core is pretty elegant. The network error handling less so.

dholth:
@baszalmstra @jaimergp ping

cep-jlap.py
jaimergp, line 1: I haven't seen attachments to CEPs before, but in case that's not desired, we can attach this as an addendum in triple-backtick fences.

baszalmstra:
@travishathaway implemented this CEP in rattler. There it is enabled by default which means pixi is using this by default and we have been using this as such for a while.

This may be due to the implementation in rattler but we have not been particularly happy with it. As far as I can tell the algorithm provides a tradeoff between bandwidth vs compute. You transfer significantly less data but computing the full repodata requires some CPU and (in our case) a lot of memory (in the order of GBs). People with low memory (embedded) device often run into out- of-memory issues because of this. 

With the defaulting of the zstd compressed files the download overhead has also become significantly less. But I also realize that this is not a scalable solution.

But! I have not yet actively profiled this, or really tried to improve the current situation. I will make some time for that in the next few weeks and report back here with some actual numbers. 

Im very curious about how other people experience this. @wolfv, @jaimergp Have you tried using JLAP? Its not implemented in (micro)mamba right?

@travishathaway Do you think there are more avenues we can explore to optimize the performance of the implementation in rattler?

dholth:
On my machine, memray shows a peak of about 420MB for `repodata = json.load(open("/Users/dholth/miniconda3/pkgs/cache/09cdf8bf.json"))` which is the conda-forge noarch repodata.json.

dholth:
In Python, it might be possible to have a sqlite-backed repodata.json that returns proxy objects for `"packages"` and `"packages.conda"`, and apply typical `"patch": [{"op": "add", "path": "/packages/pre_commit-3.4.0-hd3eb1b0_1.tar.bz2" ...` without loading the entire file. (I've timed repodata.json-to-sqlite that processes the full repodata.json each time and it is too much slower)

travishathaway:
@baszalmstra,

I thought about this for a bit and revisited the Rust implementation. Perhaps one thing that would help is removing the re-ordering of the document that is done in order for the checksums to match at the end. That's just a hunch though.

No matter what, we will always have to at least load the full repodata.json into memory in order to successfully apply the patches. I think that's just the sad reality we must face when choosing to use a JSON file as our database 😭.

Perhaps we could come up with a way to partition repodata into separate files, but that seems like it would only yield an unacceptably complex solution.

dholth:
@travishathaway It's designed so that the checksum doesn't match at the end. We don't specify the server's json formatting. But we keep track of the on-disk versus putative "logically equivalent to server-side repodata.json with this checksum".

baszalmstra:
> No matter what, we will always have to at least load the full repodata.json into memory in order to successfully apply the patches. I think that's just the sad reality we must face when choosing to use a JSON file as our database 😭.

Yeah but I dont think thats strictly required. Its because we use JSON as the level of abstraction at which we patch the content. We dont have to parse and load the entire JSON to query information from it in rattler by sparsely reading the content.

I haven't though about this enough yet but we could also patch on the byte level using zchunk for instance. The CEP also mentions that zchunk. Has that been evaluated? Im very curious about any issues encountered there.

dholth:
If we look at https://gist.github.com/dholth/cc76ce07f1c6ff099f440fc901bea35b, which are the 'op' and 'path' properties of individual json patch elements, it shows that almost all of the patch operations in the current https://conda.anaconda.org/conda-forge/linux-64/repodata.jlap look like ` ('add', '/packages.conda/pymeep-1.27.0-mpi_mpich_py310h1234567_2.conda'),` and operate on a single package at a time.

In Python, it would be easy to author a lazy object that looked like `repodata.json` but loaded individual packages from storage by hooking the dictionary index `[]` operator on `packages` and `packages.conda`.

With knowledge of the patch format, it might be reasonable to load objects (individual package records) based on the first part of the patch path, then send the rest of the path to the json patch library.

`zchunk` is neat. It's hard to implement in pure Python. Running it against the possibly-formatted `repodata.json` and preserving whitespace is a waste. It would make more individual requests to the server to get the necessary patch information.

dholth:
I've add a two-file scheme to this CEP. Instead of trying to update repodata.json each time, we accumulate patches in a simple overlay. It will be a while before (re-)writing those additions approaches the time needed to reserialize a 220MB conda-forge/linux-64/repodata.json. There is an implementation in conda+libmamba that shows writing and reading patches. It eliminates about half of the time needed (the time spent calling json.dumps() on repodata.json). It will eliminate all `repodata.json` parsing on the Python side for patches that only add complete new packages. A bit more work is needed to integrate conda with the new two-file API for the solver.

wolfv:
I think this can be closed.

PR #21, opened 2022-05-11
CEP-3 updates

jaimergp:
@jezdez - I couldn't push to your branch, so here it is.

I added the basic technical info to understand _how_ and _why_ we architected the integrations this way, but didn't go into the details of the package structure and modules.

Let me know if you need something else expanded!


cep-3.md
```diff
@@ -42,27 +47,56 @@ solver is to be considered but the first step to affect lasting change
...
+tree.
+
+`conda.core.solve` sits high up in the conda logic, right below the
+command-line interface for the `install`, `update`, `remove` and
+`create` commands. There, a class named `Solver` is in charge of
+collecting the index metadata, prefix state and other variables before
+sending the list of required `MatchSpec` objects to the underlying
+`conda.resolve.Resolve` instance.
+
+The `Resolve` class its too tied to the `Clauses` class, and some of its
```
jezdez, line 72:
```suggestion
The `Resolve` class is too tied to the `Clauses` class, and some of its
```

cep-3.md
```diff
@@ -42,27 +47,56 @@ solver is to be considered but the first step to affect lasting change
...
+abstractions do not hold true for `libsolv`. As a result, we decided to
+operate at the `Solver` level. We subclassed `Solver` and replaced its
+`Solver.solve_final_state()` method, which only requires us to return an
+iterable of `PackageRecord` objects that is then diff'd with the initial
+environment state and turned into a transaction by the other methods in
+the base class.
```
jezdez, line 78:
```suggestion
abstractions do not hold true for `libsolv`. As a result, we need to
operate at the `Solver` level. We have to subclass `Solver` and replace its
`Solver.solve_final_state()` method, which only requires us to return an
iterable of `PackageRecord` objects that is then diffed with the initial
environment state and turned into a transaction by the other methods in
the base class.
```

cep-3.md
```diff
@@ -42,27 +47,56 @@ solver is to be considered but the first step to affect lasting change
...
+
+The resulting subclass lives now in a separate project
+[conda-libmamba-solver] and only depends on `libmambapy` (Mamba's
+library) and `conda` itself. To enable the experimental solver, we added
+a new key to the `Context` class (`experimental_solver`) and a
+command-line flag (`--experimental-solver`), which can take any of the
+values added to a new `Enum` in `conda.base.constants`
+(`ExperimentalSolverChoice`). The adequate solver class is returned by a
+helper function added to `conda.core.solve` (`_get_solver_class()`) once
+the `context` has been initialized.
```
jezdez, line 88:
```suggestion
The resulting subclass would live in a separate project
[conda-libmamba-solver] and only depends on `libmambapy` (Mamba's
Python wrapper) and `conda` itself. To enable the experimental solver,
we add a new key to the `Context` class (`experimental_solver`) and a
command-line flag (`--experimental-solver`), which can take any of the
values added to a new `Enum` in `conda.base.constants`
(`ExperimentalSolverChoice`). The adequate solver class would be
returned by a helper function added to `conda.core.solve`
(`_get_solver_class()`) once the `context` has been initialized.
```

cep-3.md
```diff
@@ -76,24 +110,32 @@ already supported solvers, it's to be expected that Mamba's solver will
...
 set of dependencies.
 
-One prerequisite to implementing this CEP is therefor to provide a
-list of essential dependency solutions in the form of an automatic test
-suite of expected solutions and a prose documentation of the minimum
-viable solving behavior. Both will be the basis to determine the integration
-success of the Mamba solver to secure backward compatibility at a high degree.
+To guarantee an acceptable degree of backwards compatibility with
+`conda`'s behavior, we did our best to make the experimental solver
+comply with the existing `conda` integration test suite.
```
jezdez, line 115:
```suggestion
To guarantee an acceptable degree of backwards compatibility with
`conda`'s behavior, we need to make the experimental solver
comply with the existing `conda` integration test suite.
```

jezdez (commented):
Just a few corrections to make sure the CEP is written in a future tense because - while we know that we've done the work already - the intentions of CEPs like this is to describe a plan and not the work already done.

PR #24, opened 2022-05-18
[WIP/Help Wanted] CEP-7 CPython Version Support

travishathaway:
This PR proposes an official policy and schedule that conda organization can follow regarding its projects and their support for various versions of CPython.

*Note: this is still a work in progress and community input and help is very much welcomed*

### DIFF
--- a/cep-7.md
+++ b/cep-7.md
@@ -1,0 +1,69 @@
+<table>
+<tr><td> Title </td><td> CPython Version Support </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Travis Hathaway &lt;thathawa@anaconda.com&gt;</td></tr>
+<tr><td> Created </td><td> May 18th, 2022</td></tr>
+<tr><td> Updated </td><td> May 18th, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+Decisions for which Python versions we support are currently made adhoc and in a manner
+which is not easily predicted by the conda community. In order for this to be a more organized
+process and to provide greater transparency, this document outlines the procedures we take to
+add support for newer Python versions and how we drop support for older versions. Additionally,
+this document outlines our official schedule for when we add support and begin testing for new
+versions as well as when we drop support and stop testing for older version.
+
+## Specification
+
+Questions this section should answer:
+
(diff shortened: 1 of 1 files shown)


cep-7.md
```diff
@@ -0,0 +1,69 @@
...
+</table>
+
+## Abstract
+
+Decisions for which Python versions we support are currently made adhoc and in a manner
+which is not easily predicted by the conda community. In order for this to be a more organized
+process and to provide greater transparency, this document outlines the procedures we take to
+add support for newer CPython versions and how we drop support for older versions. Additionally,
+this document outlines our official schedule for when we add support and begin testing for new
+versions as well as when we drop support and stop testing for older version.
```
beeankha, line 18:
```suggestion
Decisions for which Python versions we support are currently made ad hoc and in a manner
which is not easily predicted by the conda community. In order for this to be a more organized
process and to provide greater transparency, this document outlines the procedures we take to
add support for newer CPython versions and how we drop support for older versions. Additionally,
this document outlines our official schedule for when we add support and begin testing for new
versions as well as when we drop support and stop testing for older versions.
```

EwoutH:
Personally I would suggest staying closer to [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) than to the Python [End of Life](https://endoflife.date/python) dates. This allows keeping a leaner and meaner codebase, profiting earlier from new features and [performance improvements](https://docs.python.org/3.11/whatsnew/3.11.html#faster-cpython) while still providing a reasonably long migration time of 42 months. Most scientific libraries already follow this schedule.

As for new versions, I would suggest starting up testing, CI and CD workflows right when a new version reaches beta. This way there will be a fair chance that the new minor Python version could be supported relatively soon, instead of the current situation where the latest conda release is still Python 3.9 while Python 3.10 is out for over 7 months.

jakirkham:
It's worth stating (though maybe this is obvious) that the version support here is for Conda itself (and related tooling). While these tools are consumed by users of NumPy, etc., they are not dependent on NumPy themselves. They are also consumed by users outside of the NumPy (or even Python) space. So establishing support for Python-based tooling to use Python's EOL scheme is reasonable. It's also worth noting that Python itself has sped up its own release cycle; so, it is more nimble than it was in the past.

EwoutH:
Would it be possible to get some movement this? Python 3.11 isn't supported by conda yet, and Python 3.12 beta is coming up. I would be nice if 3.12 support won't take as long as 3.11 is currently taking.

PR #29, opened 2022-06-10
Technical specification for creation, modification and deletion of conda environments

jaimergp:
(WIP)

### DIFF
--- a/cep-techspec.md
+++ b/cep-techspec.md
@@ -1,0 +1,61 @@
+<table>
+<tr><td> Title </td><td> Technical specification for creation, modification and deletion of conda environments </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jrodriguez@quansight.com&gt;</td></tr>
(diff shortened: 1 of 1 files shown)


cep-techspec.md
```diff
@@ -0,0 +1,339 @@
...
+packages installed.
+
+The only technical requirement enforced by `conda` activation mechanisms is that the environment
+must contain a `conda-meta` directory with a `history` file. Other than that, the environment can
+contain anything. It must be noted, though, that there are other important locations which might be
+present in the directory tree. These are optional, but have a role in the conda environment
+lifecycle:
+
+- `bin/`: added to PATH on Unix
+- `lib/`: searched for dynamic libraries on Unix
```
ocefpaf, line 49: @jaimergp we try our best to force packages to use `lib` but sometimes `lib64` passes without our knowledge. Maybe, if we write this down in the spec, we should start warning/breaking packages that deviates from this at build time.

jaimergp: Agreed! I've been thinking about resurrecting `conda-verify` (maybe a subset of the checks), and including some of the stuff mentioned in https://github.com/conda-forge/conda-forge.github.io/pull/1784 to make it easier for the reviewer.

PR #31, opened 2022-06-28
CEP 7: Conda Capitalization Standard

kathatherine:
This pull request contains a draft CEP for a final official decision of conda capitalization within the documentation, so that it can be enforced for consistency in PRs and any other written materials. Please see the CEP itself for more information about the proposal.

### DIFF
--- a/cep-7.md
+++ b/cep-7.md
@@ -1,0 +1,35 @@
+<table>
+<tr><td> Title </td><td> Conda Capitalization Standard </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Katherine Kinnaman &lt;kkinnaman@anaconda.com&gt;</td></tr>
+<tr><td> Created </td><td> Jun 28, 2022</td></tr>
+<tr><td> Updated </td><td> Jun 28, 2022</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+The capitalization conventions for conda need to be decided and written down within the conda docs so that they can be enforced during PRs, for consistency within written materials about and related to conda.
+
+## Specification
+
+This change will only affect the text of the conda docs, so shouldn't need any special technical considerations. 
+
+## Motivation
+
+The capitalization of C/conda can be inconsistent. It should be formally decided and written down so that it can be enforced for consistency in any documentation or written materials about conda.
+
+## Alternatives
+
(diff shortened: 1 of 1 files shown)


cep-7.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Motivation
+
+The capitalization of C/conda can be inconsistent. It should be formally decided and written down so that it can be enforced for consistency in any documentation or written materials about conda.
+
+## Alternatives
+
+- Conda should be lowercase except when used at the beginning of a sentence. When speaking about the command, code formatting should be used (i.e. `conda`)
+
+- Conda should be capitalized when speaking about the organization/product and lowercase, with code formatting, when discussing the command
```
kenodegard, line 27: Move the choice you are advocating for up to the "Specification".

"Alternatives" should indicate other options that were considered but rejected. Include any reasoning for why alternative ideas were rejected.

cep-7.md
```diff
@@ -0,0 +1,33 @@
...
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+The capitalization conventions for conda need to be decided and written down within the conda docs so that they can be enforced during PRs, for consistency within written materials about and related to conda.
+
+## Specification
+
+Conda should be lowercase except when used at the beginning of a sentence. When speaking about the command, code formatting should be used (i.e. `conda`)
```
kenodegard, line 17: And what about the rare oddball cases when the sentence starts with the command? E.g.:

> `conda create ...` will create a new environment.

Otherwise, it would simply be incorrect:

> `Conda create ...` will not create a new environment.

Are we simply disallowing `conda` commands from occurring at the beginning or do we allow this exception?

kathatherine: Great edge case to consider. I think in that case, we would not have the capitalization apply to the command.

So maybe a rewrite of "Conda should be written in lowercase. References to the `conda` command should use code formatting. If the use of conda is not a command and if conda is at the beginning of a sentence, conda should be uppercase."

jaimergp: Are we only referring to conda as the tool? Or also extended meanings like "the conda ecosystem", "conda packages", "conda channels"...?

travishathaway (commented):
@kathatherine I think that this CEP should definitely contain an "Examples" section. This will be really helpful for those referencing it later.

cep-7.md
kathatherine: Yeah, I want to make this as simple to follow as possible, so yes, it would extend. Basically, anything that isn't a reference to the command itself should just be regular text.

kathatherine:
Closing this because we solved it with the PRs linked above.

PR #38, opened 2022-08-17
Renaming Packages

ForgottenProgramme:
This was written by Michael Sarahan and has been taken from the conda wiki.


beckermr:
Replacing with a new package is very liable to break environments. I think instead conda should refuse to do any operations on environments that have EOL packages in them. They should be frozen but usable at run time. Conda should be able to make EOL environments from specific metadata through some mechanism as well. For example one could provide a date and conda could pretend it is solving at that date so that stuff that is already EOL is not used but stuff that is not EOL yet as of that date is used.  These changes would make things reproducible without getting rid of the good things here related to freeing namespaces and being able to restructure the ecosystem as a whole.

PR #44, opened 2022-12-19
Add plugin hook for overriding network requests in conda

travishathaway:
This CEP proposed a new plugin hook that will allow plugin authors to completely customize the way that network requests are made. Please check what is currently on the CEP and use this pull request as a forum to discuss changes or oppositions to it.

### DIFF
--- a/new-cep.md
+++ b/new-cep.md
@@ -2,10 +2,17 @@
     <tr><td> Title </td><td> Conda Session Plugin Hook </td>
     <tr><td> Status </td><td> Draft </td></tr>
     <tr><td> Author(s) </td><td>Travis Hathaway&lt;thathaway@anaconda.com&gt;</td></tr>
-    <tr><td> Created </td><td>TBD</td></tr>
-    <tr><td> Updated </td><td>TBD</td></tr>
-    <tr><td> Discussion </td><td>TBD</td></tr>
-    <tr><td> Implementation </td><td>TBD</td></tr>
+    <tr><td> Created </td><td>December 20, 2022</td></tr>
+    <tr><td> Updated </td><td>December 20, 2022</td></tr>
+    <tr><td> Discussion </td><td>https://github.com/conda-incubator/ceps/pull/44</td></tr>
+    <tr>
+        <td> Implementation </td>
+        <td>
+            <ul>
+                <li><b>Prototype:</b> https://github.com/conda/conda/pull/12139</li>
+            </ul>
+        </td>
+    </tr>
 </table>
 
 [requests-session]: https://requests.readthedocs.io/en/latest/api/#requests.Session
@@ -16,19 +23,22 @@ This CEP introduces a new a plugin hook that will enable customization
 of all network traffic in conda. The way we intend to do this is by
 allowing plugin authors to essentially replace the `CondaSession`
 class. Plugin authors will either subclass our existing session or
-write a new class that conforms to the `requests.Session` API. The primary
+write a new class that conforms to the `requests.Session` API. 
+Overriding this session class can either be done for all network 
+requests or on a channel-by-channel basis. The primary
 use case driving this forward is support for better authentication
-handling. In this CEP, we also provide an example in the form of basic 
-HTTP authentication to make our case for such a plugin hook more 
-compelling. We believe that this hook not only serves for better support 
-various authentication schemes but is flexible enough to handle other use 
-cases in the future.
+handling. In this CEP, we also provide an example implementation 
+in the form of basic HTTP authentication to make our case for such 
+a plugin hook more  compelling. We believe that this hook not only
+serves for better support  various authentication schemes but is 
+flexible enough to handle other use cases in the future.
 
 ## Specification
 
 This new plugin hook will allow plugin authors to completely replace the
 `CondaSession` class currently defined in conda to handle all network 
-traffic. Plugin authors will have the ability to either define a completely
+traffic. This can either be done on a per channel basis or for all
+network. Plugin authors will have the ability to either define a completely
 new class as long as it conforms to the `requests.Session` API or simply
 subclass the existing `CondaSession` class to make any desired 
 customizations.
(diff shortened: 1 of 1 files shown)


new-cep.md
```diff
@@ -0,0 +1,101 @@
...
+This CEP introduces a new a plugin hook that will enable customization
+of all network traffic in conda. The way we intend to do this is by
+allowing plugin authors to essentially replace the `CondaSession`
+class. Plugin authors will either subclass our existing session or
+write a new class that conforms to the `requests.Session` API. The primary
+use case driving this forward is support for better authentication
+handling. In this CEP, we also provide an example in the form of basic 
+HTTP authentication to make our case for such a plugin hook more 
+compelling. We believe that this hook not only serves for better support 
+various authentication schemes but is flexible enough to handle other use
```
beeankha, line 24:
```suggestion
compelling. We believe that this hook not only serves for better support of
various authentication schemes but is flexible enough to handle other use 
```

new-cep.md
```diff
@@ -0,0 +1,101 @@
...
+
+This new plugin hook will allow plugin authors to completely replace the
+`CondaSession` class currently defined in conda to handle all network 
+traffic. Plugin authors will have the ability to either define a completely
+new class as long as it conforms to the `requests.Session` API or simply
+subclass the existing `CondaSession` class to make any desired 
+customizations.
+
+Below is an example of how a plugin author would use this new hook
+to create a their own custom `CondaSession` class:
```
beeankha, line 37:
```suggestion
to create their own custom `CondaSession` class:
```

jezdez (commented):
The name of the hook has me worried here, it's a little Russian doll style naming confusion IMO and visible in the example: `CondaSessionClass` has a `session_class` attribute of subclasses of `CondaSession`.

I wonder if we should use "session class" for the plugin name, as it's an implementation detail of the requests library and not a feature of conda itself. It feels too overused for a plugin hook and I think precludes what this could be used for.

I think a few things could work to balance this out:

- `CondaSession` should be renamed to `CondaRequestsSession` since that's what it is and would prevent 3rd party devs to accidentally try to mimic that class instead of the `requests.Session` class.
- Use "requester" or "transmitter" as the domain name of this new plugin hook.
  So that `conda install --transmitter=curl scipy` works

The example would look like this:

```python
class CustomCondaRequestsSession(CondaRequestsSession):
     """
     Our custom CondaSession class which defines an additional class
     property
     """
     def __init__(self):
         super().__init__()
         self.custom_property = 'custom_property'


 @hookimpl
 def conda_transmitters(): 
     """
     Register our custom CondaSession class
     """
     yield CondaTransmitter(
         name=PLUGIN_NAME, 
         session=CustomCondaRequestsSession
     )
```

The UX would be:

```
$ CONDA_TRANSMITTER=curl conda install scipy
```

```
$ conda install --transmitter=curl scipy
```

jezdez:
BTW as some inspiration for people creating requests-style session classes: https://github.com/dcoles/pycurl-requests

travishathaway:
@jezdez,

I really like your "transmitter" idea. You pointed something out that I felt more and more awkward about as I was writing the plugin prototype, which is the naming of the `CondaSession` class. I'd also like to rename that class in any implementation we go with too.

Also, thank you for adding those high level UX examples. I will definitely make room for those in the CEP too (under "Specification").

travishathaway:
@jezdez,

I switched this back to `fetch` for now 😅 . It's shorter and easier to type than `transmitter`. It also aligns with the original "spirit" of this work (whatever that means 😂).

new-cep.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Motivation
+
+### Better handling for authentication
+
+The primary motivation behind this new plugin hook is
+giving plugin authors the ability to support a variety of authentication
+schemes. Currently, conda only supports HTTP Basic Authentication
+and a custom form of token based authentication that is highly coupled
+with [anaconda.org][https://anaconda.org]. By creating this plugin hook,
```
jaimergp, line 136:
```suggestion
with [anaconda.org](https://anaconda.org). By creating this plugin hook,
```

new-cep.md
```diff
@@ -0,0 +1,226 @@
...
+fetch: custom_fetch
+```
+
+or on a per channel basis:
+
+```yaml
+channels:
+    - https://my-custom-conda-packages.com:
+        fetch: custom_fetch
+```
```
jaimergp, line 107: How can this be achieved via the CLI or env vars?

travishathaway: @jaimergp,

I am not sure there is a clean way to do this that would be the most flexible.

Here's an example of a method one could use on the command line to effectively achieve the the configuration defined above:

```bash
conda install python \
  --channels "https://my-custom-conda-packages.com" \
  --fetch custom_fetch \
  --override-channels
```

and with environment variables:

```bash
export CONDA_CHANNELS="https://my-custom-conda-packages.com"
export CONDA_OVERRIDE_CHANNELS=1
export CONDA_FETCH=custom_fetch
conda install new_package
```

The only problem is that this approach breaks as soon as you want to add another channel in there (e.g. "conda-forge"). For the most flexible option, I think passing in JSON as the value for channel makes the most sense:

```bash
conda install python \
  --channels '{"https://my-custom-conda-packages.com": {"fetch": "custom_fetch"}}'
```

and as an environment variable:

```bash
export CONDA_CHANNELS='{"https://my-custom-conda-packages.com": {"fetch": "custom_fetch"}}'
```

That is definitely not the prettiest approach, but it Works™️.

Do you have any other ideas?

travishathaway:
I wonder if we should be including support for adding additional adapters here too:

- https://github.com/conda/conda/pull/10667

Perhaps it makes sense to be able to expose just these as hooks too.

milljm:
I would be interested in this. We ended up creating a mock object for our session purposes. It would be nice to move to something more supported by the community.

https://github.com/idaholab/ncrc-client/blob/2ea96d8ac56d3f504531a9858b657ff75f6f160b/ncrc/__main__.py#L386

travishathaway:
We recently had a new feature request in conda repository that could also be supported by this new plugin hook:

- https://github.com/conda/conda/issues/12553

This would also be a good case for allowing plugin authors for adding additional adapters.

new-cep.md
jaimergp: Yea, after some thinking I think we would need JSON support for maps. I don't know if that can become a problem by being too flexible, though. Nothing Pydantic validation and schemas cannot help with :D

PR #45, opened 2022-12-22
Conda Generic Plugin Hooks

travishathaway:
This is a CEP for implementing a set of generic plugin hooks for conda. Included so far are the following hooks:

- **pre_run**: allows plugin authors to execute code before a command is run in the conda context
- **post_run**: allows plugin authors to execute code after a command is run in the conda context
- **on_execption**: allows plugin authors to execute code when an exception is thrown in conda

☝️ This list is not final, and we welcome a healthy discussion to debate exactly what belongs here.

### DIFF
--- a/new-cep.md
+++ b/new-cep.md
@@ -11,9 +11,11 @@
 ## Abstract
 
 In order to support a variety use cases and extensions to conda's default
-behavior, we propose a set of generic plugin hooks in this CEP. Included will be pre-run and post-run hooks that will allow
+behavior, we propose a set of generic plugin hooks in this CEP. Included will be `pre_run` and `post_run` hooks that will allow
 authors to execute their plugin code before or after conda commands
-run, respectively. We outline several example cases in this CEP and
+run, respectively. Additionally, we introduce the `on_exception` plugin 
+hook that will allow authors to execute when an exception is thrown. 
+For each hook, we outline example cases and
 show exactly how plugin authors will define these new hooks.
 
 ## Specification
--- a/new-cep.md
+++ b/new-cep.md
@@ -1,0 +1,39 @@
+<table>
+  <tr><td> Title </td><td> Conda Generic Plugin Hooks</td>
+  <tr><td> Status </td><td> Draft </td></tr>
+  <tr><td> Author(s) </td><td> Full Name &lt;thathaway@anaconda.com&gt;</td></tr>
+  <tr><td> Created </td><td> Dec 22, 2022</td></tr>
+  <tr><td> Updated </td><td> Dec 22, 2022</td></tr>
+  <tr><td> Discussion </td><td> _TBD_ </td></tr>
+  <tr><td> Implementation </td><td> _TBD_ </td></tr>
+</table>
+
+## Abstract
+
+In order to support a variety use cases and extensions to conda's default
+behavior, we propose a set of generic plugin hooks in this CEP. Included will be pre-run and post-run hooks that will allow
+authors to execute their plugin code before or after conda commands
+run, respectively. We outline several example cases in this CEP and
+show exactly how plugin authors will define these new hooks.
+
+## Specification
+
+## Motivation
+
+## Rationale
+
+## Sample Implementation
+
+_Point to the basic authentication example pull request._
+
+## FAQ
+
+## Resolution
+
+_TBD_
+
+## References
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).


kenodegard:
would it be better to reuse existing terminology (e.g., `*_subcommand`)?

travishathaway:
> would it be better to reuse existing terminology (e.g., `*_subcommand`)?

@kenodegard , I'm not exactly sure what that means. Can you please clarify a little?

kenodegard:
@travishathaway as in use `pre_subcommand`/`post_subcommand` instead. It isn't clear what "run" means in `pre_run`/`post_run`. Where does it occur? Is it only applicable for the `conda run` command? Etc.

new-cep.md
```diff
@@ -0,0 +1,173 @@
...
+
+The above use cases are just a few possibilities out there. We know that our users and community will
+have many more ideas for how best to utilize these generic plugin hooks.
+
+## Rationale
+
+By introducing a set of generic hooks like the above, we grant plugin authors with quite a bit of
+flexibility for customizing how conda behaves. We believe starting out with a small set of generic
+plugin hooks is the best so we do not overwhelm would be developers at first. Depending on how much
+these are used and whether there is a demand, we may choose add more in the future with a new CEP.
```
beeankha, line 159:
```suggestion
plugin hooks is the best so that we do not overwhelm potential conda contributors. Depending on how
much these are used and whether there is a demand, we may choose to add more in the future with a new CEP.
```

new-cep.md
```diff
@@ -0,0 +1,173 @@
...
+  <tr><td> Author(s) </td><td> Full Name &lt;thathaway@anaconda.com&gt;</td></tr>
+  <tr><td> Created </td><td> Dec 22, 2022</td></tr>
+  <tr><td> Updated </td><td> Dec 22, 2022</td></tr>
+  <tr><td> Discussion </td><td> _TBD_ </td></tr>
+  <tr><td> Implementation </td><td> _TBD_ </td></tr>
+</table>
+
+## Abstract
+
+In order to support a variety use cases and extensions to conda's default
```
beeankha, line 13:
```suggestion
In order to support a variety of use cases and extensions to conda's default
```

new-cep.md
```diff
@@ -0,0 +1,173 @@
...
+be `pre_run` and `post_run` hooks that will allow
+plugin authors to execute their plugin code before or after conda commands
+run, respectively. Additionally, we introduce the `on_exception` plugin 
+hook that will allow plugin authors to execute custom code when an exception is thrown. 
+For each hook, we outline example use cases and
+show exactly how plugin authors will define these new hooks.
+
+## Specification
+
+This specific calls for the creation of the following three new plugin hooks:
```
beeankha, line 24:
```suggestion
This CEP specifically calls for the creation of the following three new plugin hooks:
```

new-cep.md
```diff
@@ -0,0 +1,173 @@
...
+    """
+    print("on_exception action")
+
+
+@hookimpl
+def conda_on_exception():
+    """
+    Returns our CondaOnException class which attaches our ``custom_plugin_on_exception_action``.
+    """
+    yield CondaPreRun(
```
beeankha, line 122:
```suggestion
    yield CondaOnException(
```

new-cep.md
```diff
@@ -0,0 +1,173 @@
...
+    print("post-run action")
+
+
+@hookimpl
+def conda_post_run():
+    """
+    Returns our CondaPreRun class which attaches our ``custom_plugin_post_run_action`` to
+    the "install" and "create" command.
+    """
+    yield CondaPreRun(
```
beeankha, line 85:
```suggestion
    yield CondaPostRun(
```

travishathaway:
> @travishathaway as in use `pre_subcommand`/`post_subcommand` instead. It isn't clear what "run" means in `pre_run`/`post_run`. Where does it occur? Is it only applicable for the `conda run` command? Etc.

@kenodegard , I think I'm going to just go with `pre_command` and `post_command` to make it just tad bit shorter to type. It conveys the same meaning.

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+
+@hookimpl
+def conda_on_exception():
+    """
+    Returns our CondaOnException class which attaches our ``custom_plugin_on_exception_action``.
+    """
+    yield CondaOnException(
+        name=f"{PLUGIN_NAME}_on_exception",
+        action=custom_plugin_on_exception_action
+    )
```
jezdez, line 128: It might make sense to make sure that the hook spec is set up as [`firstresult=True`](https://pluggy.readthedocs.io/en/stable/#first-result-only) to make sure we can define exception handlers that are final, and those that just inspect the exception and then move to the next registered exception handler.

Secondly, the hook implementation should receive the exception object as an argument and store it in the `CondaExceptionHandler` instance for later use. AND the actual handler callable should _also_ be passed the exception object so that it can do something with the exception.

Lastly, I think the naming is a little off about this, the hook yields handlers, so we should name the domain object `CondaExceptionHandler` and the hook spec `conda_exception_handlers`. 

```suggestion
from conda.plugins import hookimpl, CondaOnException


PLUGIN_NAME = "custom_plugin"


def print_exception(exception):
    """
    Defines our custom on_exception handler which prints a message
    with the provided exception object.
    """
    print(f"Caught exception: {exception}")


@hookimpl
def conda_exception_handlers(exception):
    """
    Returns our CondaExceptionHandler class which attaches our ``print_exception(``.
    """
    yield CondaExceptionHandler(
        name=f"{PLUGIN_NAME}_exception_handler",
        handler=print_exception,
        exception=exception,
    )
```

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+    - All registered `pre_command` hooks will be called before the `do_call` call in the
+      [`conda.cli.main:main_subshell`][conda-pre-invoke-location] function.
+
+- Where exactly will the `post_command` hooks be called?
+    - All registered `post_command` hooks will be called after the `do_call` call in the
+      [`conda.cli.main:main_subshell`][conda-pre-invoke-location] function.
+
+- Where exactly will the `on_exception` hooks be called?
+    - All registered `on_exception` hooks will be called inside the
+      [`conda.exceptions.ExceptionHandler`][conda-on-exception-location] class.
```
jezdez, line 183: I think the `conda.exceptions.ExceptionHandler` should be instead the default exception handler registered ([using the `tryfirst=True` hookimpl parameter](https://pluggy.readthedocs.io/en/stable/index.html?highlight=tryfirst#call-time-order))

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+def conda_pre_command():
+    """
+    Returns our CondaPreCommand class which attaches our ``custom_plugin_pre_command_action``
+    to the "install" and "create" command.
+    """
+    yield CondaPreCommand(
+        name=f"{PLUGIN_NAME}_pre_command",
+        action=custom_plugin_pre_command_action,
+        run_for={"install", "create"}
+    )
```
jezdez, line 72:
```suggestion
@hookimpl
def conda_pre_commands():
    """
    Yields our CondaPreCommand instance which attaches our ``custom_plugin_pre_command_action``
    to the "install" and "create" command.
    """
    yield CondaPreCommand(
        name=f"{PLUGIN_NAME}_pre_command",
        action=custom_plugin_pre_command_action,
        run_for={"install", "create"}
    )
```

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+def conda_post_command():
+    """
+    Returns our CondaPreCommand class which attaches our ``custom_plugin_post_command_action`` to
+    the "install" and "create" command.
+    """
+    yield CondaPostCommand(
+        name=f"{PLUGIN_NAME}_post_command",
+        action=custom_plugin_post_command_action,
+        run_for={"install", "create"}
+    )
```
jezdez, line 92:
```suggestion
@hookimpl
def conda_post_commands():
    """
    Yields our CondaPreCommand instance which attaches our ``custom_plugin_post_command_action`` to
    the "install" and "create" command.
    """
    yield CondaPostCommand(
        name=f"{PLUGIN_NAME}_post_command",
        action=custom_plugin_post_command_action,
        run_for={"install", "create"}
    )
```

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+
+
+PLUGIN_NAME = "custom_plugin"
+
+
+def custom_plugin_pre_command_action():
+    """
+    Defines our custom pre-run action which simply prints a message.
+    """
+    print("pre-run action")
```
jezdez, line 59:
```suggestion
def custom_plugin_pre_command_action(command):
    """
    Defines our custom pre-run action which simply prints a message.
    """
    print(f"pre-run action before {command}")
```

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+        action=custom_plugin_pre_command_action,
+        run_for={"install", "create"}
+    )
+
+
+def custom_plugin_post_command_action():
+    """
+    Defines our custom post-run action which simply prints a message.
+    """
+    print("post-run action")
```
jezdez, line 79:
```suggestion
def custom_plugin_post_command_action(command):
    """
    Defines our custom post-run action which simply prints a message.
    """
    print(f"post-run action after {command}")
```

jezdez (commented):
Leaving some general naming and conceptual items as feedback since the way this is written at the moment doesn't fully use pluggy's abilties. 

Generally speaking it would be useful to pass the actually used command to the pre/post command action, and be a little more detailed about the exception handler implementation since it's a lot different than the pre and post command hooks.

ForgottenProgramme (approved):

new-cep.md
travishathaway: I'm curious why you added the `command` argument for this and the other example. It does raise an interesting question of what, if any arguments, we want to pass in. I'm assuming that plugin authors would simply access the `context` object for most runtime config settings they may need.

travishathaway: Okay... I thought about it a little more. Makes sense 😄

new-cep.md
```diff
@@ -0,0 +1,204 @@
...
+    """
+    Defines our custom post-run action which simply prints a message.
+    """
+    print(f"post-run action after {command}")
+
+
+@hookimpl
+def conda_post_commands():
+    """
+    Yields our CondaPreCommand instance which attaches our ``custom_plugin_post_command_action`` to
```
beeankha, line 85:
```suggestion
    Yields our CondaPostCommand instance which attaches our ``custom_plugin_post_command_action`` to
```

travishathaway:
@jezdez, thanks for the review.

When I created the `on_exception` handler, I was thinking of it more as pseudo event handling system. But, it sounds like what you think may be better is to just allow people to replace or subclass the `conda.exceptions.ExceptionHandler` class instead.

The "exception_handler" hook would then function just like the "solver" hook. You would only be allowed to configure a single exception handler.

I'm thinking this would be easier from a plugin author's perspective. All they would have to do is the following:

```python
from conda.exceptions import ExceptionHandler

# Example for implementing pretty printing exceptions with "rich"
from rich.console import Console
console = Console()

PLUGN_NAME = "rich"

class RichExceptionHandler(ExceptionHandler):
    """
    Custom exception handler that uses rich to print exceptions.
    """

    def handle_exception(self, exc_val, exc_tb):
        console.print_exception(show_locals=True)

@hookimpl
def conda_exception_handlers(exception):
    """
    Returns our CondaExceptionHandler class which attaches our ``print_exception(``.
    """
    yield CondaExceptionHandler(
        name=f"{PLUGIN_NAME}_exception_handler",
        exception_handler=RichExceptionHandler
    )
 ```

We could change the function signature of the `handle_exception` method on the `ExceptionHandler` class to pass in the actual exception object too.

What do you think? Is relying on inheritance here a Good Idea™️?

File: new-cep.md [hunk 13]
```diff
@@ -0,0 +1,218 @@
...
+    the "install" and "create" command.
+    """
+    yield CondaPostCommand(
+        name=f"{PLUGIN_NAME}_post_command",
+        action=custom_plugin_post_command_action,
+        run_for={"install", "create"}
+    )
+```
+
+### `exception_handler`
```
aovasylenko, line 97: can we review, if current implementation of exception classes should be extended to support custom exception handlers? For example - let's say in the code we have (just found the first one)
```
    if args.all and prefix == context.default_prefix:
        msg = "cannot remove current environment. deactivate and run conda remove again"
        raise CondaEnvironmentError(msg)
```

If I want to write a custom exception handler for his exception, I'm missing context - which environment and reason, why it couldn't be removed (or is conda Context the right thing to use?)

If quick question - yes, it worth to review and extend exception handling mechanism - I believe it's another CEP?

jezdez:
@travishathaway Okay, thinking about this a little more, I think we'll have to handle a situation in which multiple plugins implement exception handlers and exist in the runtime environment at the same time.

E.g. imagine if conda-libmamba-solver shipped an exception handler that would enable it to print out additional details of solver errors (e.g. https://github.com/conda/conda-libmamba-solver/issues/102) and the user _in addition_ would like to install a handler that generally renders exceptions with rich (like your example above). The "solver" hook works in that way: there can be multiple solvers installed, and the default setting and the config system determine which to use. In case of exceptions handlers, we probably shouldn't require users to enable them before using them, though. Or is that an indicator to require a `plugins enable` and `plugins disable` command to make this an explicit process?

My question is whether we want to enable multiple exception handlers to coexist and create a kind of exception handling pipeline instead of enabling exactly one exception handler?


```
exception_handlers = plugin_manager.get_hook_results("exception_handlers")
for exception_handler in exception_handlers:
    result = exception_handler.handler(exc_val, exc_tb):
    if result is not None
        return result
```

The API could require exception handlers to return if they've handled all exceptions, e.g.

```
class LibmambaExceptionHandler(ExceptionHandler):
    """
    When encountering a libmamba specific error, print it nicely.
    """
    def handle_exception(self, exc_val, exc_tb):
        if isinstance(exc_val, LibMambaUnsatisfiableError):
            return self.handle_libmamba_unsatisfiable_exception(exc_val, exc_tb)

    def handle_libmamba_unsatisfiable_exception(self, exception, traceback):
        ...

@hookimpl
def conda_exception_handlers(exception):
    """
    Returns our CondaExceptionHandler class specific to libmamba exceptions.
    """
    yield CondaExceptionHandler(
        name=f"{PLUGIN_NAME}_exception_handler",
        exception_handler=LibmambaExceptionHandler
    )

```

@jaimergp You've looked at https://github.com/conda/conda-libmamba-solver/issues/102 a bunch and I think @wolfv has thought about that as well, what's your take on this topic?

new-cep.md
```diff
@@ -0,0 +1,218 @@
...
+
+@hookimpl
+def conda_on_exception():
+    """
+    Returns our ``CondaExceptionHandler`` class which attaches our 
+    registers our ``
+    """
+    yield CondaExceptionHandler(
+        name=f"{PLUGIN_NAME}_exception_handler",
+        handler=CustomExceptionHandler
```
aovasylenko, line 135: will it support `run_for` as well?

new-cep.md
```diff
@@ -0,0 +1,218 @@
...
+
+### `pre_command` and `post_command`
+
+Plugin authors create the `pre_command` and `post_command` hooks by first defining a `conda.plugins.hookimpl`
+decorated function called either `conda_pre_command` or `conda_post_command` which return either a `CondaPreCommand` 
+or `CondaPostCommand` class, respectively. These classes will both receive the following three properties:
+
+- `name`: unique name which identifies this plugin hook
+- `action`: a callable which contains the code to be run
+- `run_for`: a Python `set` of strings representing the commands this will be run on (e.g. `install` and `create`)
```
aovasylenko, line 46: is it optional or mandatory parameter? 
(also asked similar question if exception handler hook should have it as well)

travishathaway: Good question. I think making this argument optional with the default behavior of it running for every command makes the most sense.

new-cep.md
travishathaway: I left that out because I didn't think that this plugin should have that fine grain control. It would also complicate the implementation too.

new-cep.md
travishathaway: @aovasylenko, please see @jaimergp's comment below. I think he did a reasonable job of summarizing what you were talking about.

The scenario you are talking about would require a very extensive refactor to existing code, and I do not think it is realistic anytime soon. In the meantime, it's way better to just rely on the context object like you said.

travishathaway:
@jezdez

I have thought about this some and have decided to take @jaimergp's advice to not include a custom exception handler yet. To use a metaphor, I think we need to clean up the house a bit before inviting guests in 😅. I think that a refactor of the current exception handling mechanism should be done before we revisit this. Once that has been done, we can restart this conversation with a separate CEP.

In the meantime, I will remove exception handling from this CEP and just stick with the `pre_command` and `post_command` hooks. These already enable quite a bit.

Once I've removed the exception handling pieces from this CEP, I will move it out of "draft" status. Hopefully, the vote will follow soon afterwards 👍.

travishathaway:
Example of how our plugin developers could use these hooks:

- https://github.com/conda/conda/pull/10567

XuehaiPan:
Hi, some questions about the hooks. As I post in the description in PR conda/conda#10567:

- conda/conda#10567

A core usage is to define environment variables when activating/deactivating a conda environment. E.g.:

```bash
export CMAKE_PREFIX_PATH="${CONDA_PREFIX}"
```

NOTE: `conda env config vars set` does not support variable substitution. So we need hooks here.

In conda/conda#10567, I propose to create a post-create hook to add `env-vars.sh` to `CONDA_PREFIX/etc/activate.d` and `CONDA_PREFIX/etc/deactivate.d` on each environment creation. That approach is covered by this CEP. Now, I can create a `CondaPostCommand` with `run_for={"create"}`.

I wonder can the hooks affect the current shell environment like `CONDA_PREFIX/etc/activate.d/*.sh` do? (not the process of the `conda` command, which is a subprocess) So I can write `post-activate` hooks directly rather than `etc/activate.d/*.sh`. It would be nice to have an example of this in the documentation too.

```bash
# post-activate hook
export CONDA_CMAKE_PREFIX_PATH_BACKUP="${CMAKE_PREFIX_PATH}"
export CONDA_LD_LIBRARY_PATH_BACKUP="${LD_LIBRARY_PATH}"
export CMAKE_PREFIX_PATH="${CONDA_PREFIX}"
export LD_LIBRARY_PATH="${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH}"
```

```bash
# pre-deactivate hook
export CMAKE_PREFIX_PATH="${CONDA_CMAKE_PREFIX_PATH_BACKUP}"
export LD_LIBRARY_PATH="${CONDA_LD_LIBRARY_PATH_BACKUP}"
unset CONDA_CMAKE_PREFIX_PATH_BACKUP
unset CONDA_LD_LIBRARY_PATH_BACKUP
```

travishathaway:
@XuehaiPan,

Thanks for your feedback. To answer your question, no this not possible. I researched this myself once and came to the conclusion that simply is not possible because of how shells work (sub-process cannot modify the parent process' environment variable values). This is why we use so much shell specific code in conda and why it's necessary for us to modify a user's shell configuration file (e.g. `.zshrc` and `.bashrc`).

PR #46, opened 2023-01-09
initial cep for repodata state

wolfv:
Created a quick CEP for the new repodata state format (cc @dholth as we had some discussions about this. Happy to list you as author!).

Also regarding the spec happy to make changes. My hope is just that mamba and conda can both share the same format.

### DIFF
--- a/cep-repodata-state.md
+++ b/cep-repodata-state.md
@@ -1,0 +1,79 @@
+<table>
+<tr><td> Title </td><td> .state.json files for repodata metadata </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> Jan 09, 2023</td></tr>
+<tr><td> Updated </td><td> Jan 09, 2023</td></tr>
+<tr><td> Discussion </td><td> https://conda.slack.com/archives/C017F7C0VM3/p1672669131100819 </td></tr>
+<tr><td> Implementation </td><td> https://github.com/mamba-org/mamba/pull/2113 </td></tr>
+</table>
+
+## Abstract
+
+Changing how conda and mamba store metadata about repodata.json downloads.
+
+### Motivation
+
+When conda currently downloads `repodata.json` files from the internet, it stores metadata "inside" the file by adding some JSON keys:
+
+- `_url`: The URL that was requested
+- `_etag`: ETag returned from server
+- `_mod`: Last-Modified header from server
+- `_cache_control`: Cache-Control header from server
+
+These are stored as three string values. 
+
+This is not an ideal approach as it modifies the `repodata.json` file and corrupts e.g. the hash of the file. Also, the repodata files have gotten increasingly large, and parsing these state values can require parsing a large `json` file.
+
+Therefore we propose to store the metadata in a secondary file called `.state.json` file next to the repodata.
+
+Another motivating factor is that for the `jlap` proposal we need to (repeatedly) compute the hash value of the `repodata.json` file -- that only gives correct results straight away when the repodata is stored externally.
+
+Both mamba and conda currently use the same cache folder. If both don't implement the same storage strategy but continue to share the same repodata cache, it would lead to frequent cache busting.
+
+### Specification
+
+```json
+{
+    // we ensure that state.json and .json files are in sync by storing the file
+    // last modified time in the state file, as well as the file size
+    "file_mtime": {
+        // seconds and nanoseconds counted from UNIX timestamp (1970-01-01)
+        "seconds": "<integer>",
+        "nanoseconds": "<integer>"
+    },
+    "file_size": "<integer>", // file size in bytes
+
+    // The header values as before
+    "url": "<string>",
+    "etag: "<string>",
+    "mod": "<string>",
+    "cache_control": "<string>",
+
+    // these are alternative encodings of the repodata.json that 
+    // can be used for faster downloading
+    // both `has_zst` and `has_jlap` keys are optional but should be kept
+    // even if the other data times out or `file_mtime` does not match
+    "has_zst": {
+        // UTC RFC3999 timestamp of when we last checked wether the file is available or not
+        // in this case the `repodata.json.zst` file
+        // Note: same format as conda TUF spec
+        "last_checked": "2023-01-08T11:45:44Z",
+        // false = unavailable, true = available
+        "value": false
+    },
+    "has_jlap": {
+        // same format as `has_zst`
+    }
+}
+```
+
+If the `state.json` file_mtime or file_size does not match the `.json` file actual `mtime`, the header values are discarded. However, the `has_zst` or `has_jlap` values are kept as they are independent from the repodata validity on disk.
+
+### Backward compatibility
+
+Older clients that try to reuse the existing cache will not be able to make use of the cached repodata as they do not know about the state (since it's not written to the same location). That means they will redownload the repodata.
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).


cep-repodata-state.md
```diff
@@ -0,0 +1,79 @@
...
+
+Both mamba and conda currently use the same cache folder. If both don't implement the same storage strategy but continue to share the same repodata cache, it would lead to frequent cache busting.
+
+### Specification
+
+```json5
+{
+    // we ensure that state.json and .json files are in sync by storing the file
+    // last modified time in the state file, as well as the file size
+    "file_mtime": {
```
baszalmstra, line 40: Maybe you can store this in an ISO standard DateTime format instead? Similar to the `has_zst.last_checked` date. Or is that not precise enough when dealing with file times?

cep-repodata-state.md
```diff
@@ -0,0 +1,79 @@
...
+        // seconds and nanoseconds counted from UNIX timestamp (1970-01-01)
+        "seconds": INTEGER,
+        "nanoseconds": INTEGER
+    },
+    "file_size": INTEGER, // file size in bytes
+
+    // The header values as before
+    "url": STRING,
+    "etag": STRING,
+    "mod": STRING,
```
baszalmstra, line 50: Maybe change this to `last_modified` similar to the header name.

baszalmstra (commented):
Looks good! Looking forward to it!

cep-repodata-state.md
dholth: IMO the fractional time `os.stat().st_mtime` would not be a problem otherwise `mtime_ns` would be a decent name.

cep-repodata-state.md
dholth: These are the old names without a leading _

dholth:
One thing about this is that when you start using alternative formats (.zst, .jlap) the remote headers last-modified, etag, cache-control come from the alternate file.

cep-repodata-state.md
wolfv: I would have to check how exactly the fractional time is computed though.

cep-repodata-state.md
wolfv: I am open to change them, though.

cep-repodata-state.md
dholth:
```
In [51]: datetime.datetime.fromtimestamp(1666095612162394000/1000000000).isoformat()
Out[51]: '2022-10-18T08:20:12.162394'
```

🤔

dholth:
An example of the current jlap branch's `.state.json`.

`have` is the nominal hash (what the hash of the original repodata.json was according to jlap)

`have_hash` is the actual hash on disk since we don't serialize with exactly the same sorting, formatting as conda-index. Could be used instead of `mtime` (if file on disk doesn't match have_hash, then it doesn't correspond to this state.json)

`jlap` includes too many headers from the jlap request, an intermediate hash `iv` corresponding to `pos-` bytes in the file, and the last line of the jlap file.

```json5
{
 "_url": "https://repo.anaconda.com/pkgs/main/osx-arm64/repodata.json",
 "_mod": "Fri, 06 Jan 2023 20:09:20 GMT",
 "_etag": "W/\"0134da379063a17831ef4ed73d3489dd\"",
 "_cache_control": "public, max-age=30",
 "have": "e45656091705b1be55b72a3b48068520cc3309560ca0d37fb96f2b2ea559c81f",
 "have_hash": "e45656091705b1be55b72a3b48068520cc3309560ca0d37fb96f2b2ea559c81f",
 "mtime": 1673274986.4032788,
 "jlap": {
  "headers": {
   "date": "Mon, 09 Jan 2023 14:36:25 GMT",
   "content-type": "text/plain",
   "transfer-encoding": "chunked",
   "connection": "keep-alive",
   "x-amz-id-2": "81qDgERSlA/bEpQQeL/YBn3BniAaB37uUkbD5ZySYC/h9JWb+8Sbg1ik70ufAvNtzTeGHqiwZHI=",
   "x-amz-request-id": "Q1HHT9KCY69TQXQX",
   "last-modified": "Fri, 06 Jan 2023 20:09:20 GMT",
   "x-amz-version-id": "BaggXYx0RmtOxe4B6PIl3IDX_G8Ryt5X",
   "etag": "W/\"0134da379063a17831ef4ed73d3489dd\"",
   "cf-cache-status": "MISS",
   "expires": "Mon, 09 Jan 2023 14:36:55 GMT",
   "cache-control": "public, max-age=30",
   "set-cookie": "__cf_bm=YfYijeZ0Y8xc_XBZebA1UA1bX9uz47v67b3guqZFfY0-1673274985-0-Ab9B1pIhPfM0fQkGk5rTS9A5vvc3tPD37jnV+pmXSI2C82sgdxdKaBCB3zhj4wQ6P1yVmaYRpioaARDTmt6H5s0=; path=/; expires=Mon, 09-Jan-23 15:06:25 GMT; domain=.anaconda.com; HttpOnly; Secure; SameSite=None",
   "vary": "Accept-Encoding",
   "server": "cloudflare",
   "cf-ray": "786de7b3ffe9244d-ATL",
   "content-encoding": "gzip"
  },
  "iv": "2f598f0587410d455c8370bef19759fd7a25b4aad4d65c3b3b1be7c7422a938c",
  "pos": 1714976,
  "footer": {
   "url": "repodata.json",
   "latest": "e45656091705b1be55b72a3b48068520cc3309560ca0d37fb96f2b2ea559c81f"
  }
 }
}
```

cep-repodata-state.md
dholth: Python's datetime has microsecond resolution. You would always have to convert st_mtime to a datetime, and then compare the two datetimes instead of trying to convert the stored datetime to `.timestamp()` and compare with `st_mtime`.

```
In [39]: datetime.fromtimestamp(Path('Dockerfile').stat().st_mtime, tz=zoneinfo.
    ...: ZoneInfo('UTC')).resolution
Out[39]: datetime.timedelta(microseconds=1)
```

wolfv: I think the fractional time is not great because it makes it appear as if it was nanosecond resolution while it isn't. And then we'd have to match the same behavior in C++ or any other language that wants to share the cache.

It seems more natural to store seconds / nanoseconds since UNIX epoch for the mtime since that is what Python does and we don't need to deal with timezones.

However, I am open to other suggestions as long as we can decide to the same behavior.

dholth: Don't know about why a . is a problem however a single nanoseconds since epoch number would be tolerable

wolfv: The reason is that we loose precision and a comparison will not be straightforward anymore:

https://docs.python.org/3/library/os.html#os.stat_result.st_ctime_ns

https://ciaranm.wordpress.com/2009/11/15/this-week-in-python-stupidity-os-stat-os-utime-and-sub-second-timestamps/

dholth: There's something extremely pedantic to be said about timestamps; on the other hand comparing two `st_mtime_ns` is really simple.

dholth:
In the existing format `"_url": "https://conda.anaconda.org/conda-forge/osx-arm64",` doesn't contain the filename. May want to continue doing that especially since different repodata variants matter. Or ignore the field.

wolfv:
Any chance you have time to make a PR against my branch with your change suggestions? Or I can also try to give you edit rights, if you want :)

dholth:
If we are going to play with nanoseconds, let's go ahead and replace all timestamps (except those that are web server headers) with those numbers. e.g. last_checked.

We will quickly release a conda with the main last_modified, cache_control, etag state but it will take us a few more releases to get to "last checked zstd"

wolfv:
environment locking as in conda-lock or as in filesystem lockfiles to prevent overwriting things?

wolfv:
My reasoning for the different formats is that for the mtime checking it is "precise" since we actually want to match the file on disk.

For the "last time checked zst" we just want a timestamp so we can check that it's been more than 2 weeks and it doesn't need to be precise. We had this function around to create a RFC3339 string representation so I just used that. We could also use nanoseconds but this one doesn't need to be precise.

dholth:
I tried out micromamba's January release, and it downloads repodata.json.zst very quickly. Producing a state file 

In Python it is easier to store nanoseconds as a single number, `time.time_ns().bit_length()` is only 61 bits today.

In the jlap branch I store "jlap_unavailable" as a timestamp, assuming you check alternative formats in a known order of preference unless you know they are 404's.

```
{
    "cache_control": "public, max-age=30",
    "etag": "\"a9c77cc4c9b1a947375d53326f1604de\"",
    "file_mtime": {
        "nanoseconds": 960132000,
        "seconds": 1674678293
    },
    "file_size": 6974249,
    "has_zst": {
        "last_checked": "2023-01-27T02:09:53Z",
        "value": true
    },
    "mod": "Wed, 25 Jan 2023 19:54:35 GMT",
    "url": "https://repo.anaconda.com/pkgs/main/osx-arm64/repodata.json.zst"
}```

File: cep-repodata-state.md (hunk 2 above)
Comment (dholth) [reply cep-repodata-state.md:48]: Have implemented (nicer) non-underscored names in a conda branch.

**Comment by dholth** [comment]
Am working on a "lock byte 21" implementation, where we lock the `.state.json` before doing anything else (even before reading or stat'ing cached `repodata.json`. Then we always try to keep the lock for as short a time as possible, e.g. download repodata.json to a temp file, then stat it, then move it on top of the cache filename.

**Comment by wolfv** [comment]
you might also run into issues with atomic renames if your temporary file is on a different fs. `/tmp` is often a different fs :) So I'd stat it after moving.

**Comment by dholth** [comment]
The temporary file is in the same cache folder.

**Comment by baszalmstra** [comment]
@dholth Does your implementation also produce a `<hash>.lock` file (like mamba) or does the `<hash>.state.json` file also function as a lock file? I guess that would also makes a lot of sense. Would you be able to formalize that in the CEP @wolfv ?

> In the existing format `"_url": "https://conda.anaconda.org/conda-forge/osx-arm64",` doesn't contain the filename. May want to continue doing that especially since different repodata variants matter. Or ignore the field.

I agree with this. I would keep it though to ensure hash collisions don't form an issue. I also like that I can find the original URL from the hash. Can we maybe formalize that in the CEP @wolfv ?

> Have implemented (nicer) non-underscored names in a conda branch.

@dholth What did you end up calling these? I especially think the `mod` could just be named `last_modified`, similar to the HTTP header from where it comes. Also, shouldn't a few of these fields be optional, given that not all HTTP servers return these headers. WDYT @wolfv ?

**Comment by dholth** [comment]
For example I've been working in this branch, link should take you to [conda/gateways/repodata/__init__.py](https://github.com/conda/conda/blob/features/repodata-jlap/conda/gateways/repodata/__init__.py#L411) with code to handle the ISO timestamps. The RepodataState class should clearly show the current format.

From my reading of the mamba code, it locks a certain byte in the repodata.json / repodata.state.json files. I didn't notice it creating a `.lock` file although that is maybe a more old-school technique? and necessary for locking a complete directory - I haven't attempted to lock complete directories in my branch. So far I am only trying to maintain the integrity of the repodata.json cache and don't do anything to prevent e.g. parallel conda's downloading it twice, they simply won't corrupt the cache. Feedback & PR's against PR's welcome.

File: cep-repodata-state.md [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+- `_url`: The URL that was requested
+- `_etag`: ETag returned from server
+- `_mod`: Last-Modified header from server
+- `_cache_control`: Cache-Control header from server
+
+These are stored as three string values.
+
+This is not an ideal approach as it modifies the `repodata.json` file and corrupts e.g. the hash of the file. Also, the repodata files have gotten increasingly large, and parsing these state values can require parsing a large `json` file.
+
+Therefore we propose to store the metadata in a secondary file called `.state.json` file next to the repodata.
```
jezdez, line 28: I've been reviewing the conda implementation in https://github.com/conda/conda/pull/12425 and have stumbled over the `state` term there few times.

I don't think "state" correctly describes what's in the proposed file as its content is mostly derived from `stat`-like calls -- and `stat` for better or worse refers to "status" and not really the ambiguous term "state". This might be confusing for future maintainers.

I know this sounds like nitpicking, but before we're shipping this in conda, I'd rather mention it before we regret it.

cep-repodata-state.md
```diff
@@ -0,0 +1,84 @@
...
+
+Both mamba and conda currently use the same cache folder. If both don't implement the same storage strategy but continue to share the same repodata cache, it would lead to frequent cache busting.
+
+### Specification
+
+```json5
+{
+    // we ensure that state.json and .json files are in sync by storing the file
+    // last modified time in the state file, as well as the file size
+
```
jezdez, line 40: I strongly suggest to add a new `version` entry to the format, to be able to evolve it if needed.
```suggestion

    // version of the repodata status format
    "version": 1,
```

cep-repodata-state.md
dholth: We have `size` and `mtime_ns` from `stat` (short for `status`) calls; last-modified and etag headers for caching; "is some number of alternative remote files available", possibly a content hash of the full file; an intermediate hash and position to start fetching an update to the remote `jlap` file if that is being used; and the time that the remote server was last checked for new content.

cep-repodata-state.md
dholth: If I was releasing a version 2, I would say all files without a "version" are version 1.

dholth:
@baszalmstra our implementation assumes all the keys are optional, same as if the file was missing. We also treat the keys as missing if `state.json` doesn't match the cached `repodata.json` file. We could formally add that to the CEP.

cep-repodata-state.md
baszalmstra: Thats also fine for me!

baszalmstra:
Our state data structure in [rattler ](https://github.com/mamba-org/rattler) looks like [this](https://github.com/mamba-org/rattler/blob/565c3020b28aa1a86f9e7b300c7605dedeee9e61/crates/rattler_repodata_gateway/src/fetch/cache/mod.rs#L10).

You may note that besides checking the timestamp and the size we also check the `repodata.json` against a blake2 hash. If the timestamp and the size won't match but the blake2 hash does still match, we consider the data to be up-to-date.

I like the idea of having all keys be optional. Currently, the `mtime_ns`, `size`, and `url` in our implementation are not optional.

We also create an extra lockfile (`.lock`) that guards both the `repodata.json` and the `state.json` file. I think this is also what mamba is doing by observing its behavior. But looking at the code mamba is [creating](https://github.com/mamba-org/mamba/blob/e549d07b0cdc146352e63890fff62949114f5993/libmamba/src/core/subdirdata.cpp#L512) [several](https://github.com/mamba-org/mamba/blob/e549d07b0cdc146352e63890fff62949114f5993/libmamba/src/core/subdirdata.cpp#L667) [lock](https://github.com/mamba-org/mamba/blob/e549d07b0cdc146352e63890fff62949114f5993/libmamba/src/core/subdirdata.cpp#L682) [files](https://github.com/mamba-org/mamba/blob/e549d07b0cdc146352e63890fff62949114f5993/libmamba/src/core/subdirdata.cpp#L750) throughout the process on several different files.

cep-repodata-state.md
baszalmstra: I'm fine either way. 👍

dholth:
I was preparing to add the hash as well, at least for jlap. I have two hash fields called ```
NOMINAL_HASH = "nominal_hash"
ON_DISK_HASH = "actual_hash"```. `nominal_hash` is the hash according to the `.jlap` file and `actual_hash` is the hash after we `json.dumps` the updated data

cep-repodata-state.md
dholth: I think I agree with Jannis to add version: 1, and an older conda should ignore newer .state.json

cep-repodata-state.md
dholth: IMO we are using 5 meaningless letters "state", we could make it even more meaningless ".s.json" (and nicely shorter),  we could roll the dice and call it `>>> os.urandom(2).hex() 'ec5a'`.json

dholth:
I assume mamba is also trying to lock whole directories.

The development conda implementation also does the trick of writing the new repodata.json to a new file, stat'ing the temporary name, and moving it on top of the old file.

cep-repodata-state.md
baszalmstra: What about something that does makes sense? `.cache-info.json` perhaps?

dholth: plain `info.json` would be short

baszalmstra:
> The development conda implementation also does the trick of writing the new repodata.json to a new file, stat'ing the temporary name, and moving it on top of the old file.
Yeah Rattler does the same. (On windows we use some Win32 API to achieve this).

>We are always using BLAKE2(256) which are the same length as sha-256 hashes, but "by default" blake2 produces an overkill 512-bit hash.
Still, it might be nice to include the algorithm used in either the key or value. It doesn't add overhead but makes it easier for others to deduce whats going on. We could also do `"on_disk_hash" : "blake2:blabla..."`?

dholth:
For anything crypto-adjacent I'd let the version # fix the exact hash used.

cep-repodata-state.md
baszalmstra: https://quotesondesign.com/phil-karlton/

Thats fine by me!

baszalmstra:
In that case, since we can just change keys when changing versions anyway, let's name the key something with `blake2`. At least then it's clear from reading the file. You also have this in the repodata with `sha2` or `md5`.

dholth:
Parametrize important key names https://github.com/conda/conda/pull/12461/files#diff-813ca3bd61f56355fb3ea7c560d18b892d42d62a07fa0e78666dcfa50c5fda13R64

PR #48, opened 2023-03-21
CEP repodata state/info

dholth:
Updates with edits, haven't heard back from @wolfv in a while, so making a PR directly to main branch.

### DIFF
--- a/cep-repodata-state.md
+++ b/cep-repodata-state.md
@@ -42,6 +42,12 @@ Both mamba and conda currently use the same cache folder. If both don't implemen
     "mtime_ns": INTEGER,
     "size": INTEGER, // file size in bytes
 
+    // most recent remote request e.g. "304 Not Modified", instead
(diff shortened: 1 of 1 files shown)


dholth:
Yes. Although we could format repodata.json in exactly the same way as conda-index, that formatting is not guaranteed. Instead, we store the hash that the server found when creating the patches, and the hash of what we wrote to disk. We save time by not reformatting and the server can change its formatting.

For example the anaconda.org channels don't pretty-format (but also don't have a way to provide `jlap` diffs).

We could consider adopting a formal json canonicalization scheme in the future.

wolfv:
Thanks for the clarification. I think it'd be nice if the hashes match up but maybe we can fix that another time? Idk

dholth:
Yes, we could do that in the future but in the meantime we are allowed to change this code https://github.com/conda/conda-index/blob/main/conda_index/index/__init__.py#L825-L833

PR #51, opened 2023-05-04
Serve `run_exports.json` in conda channels

jaimergp:
As discussed in https://github.com/conda/conda-index/issues/102 and [this conda community meeting](https://github.com/conda-incubator/governance/blob/main/meetings/archive/20230426_agenda_and_minutes.md).

cc @chenghlee


jaimergp:
Hi @conda-incubator/steering, any thoughts or feedback here? Thanks 🙏

wolfv:
I think this is a great idea and will also simplify rattler-build!

I am leaning towards adding an empty run exports for _all_ packages (instead of leaving out those that don't have it) so that one knows that these were considered.

wolfv (approved):

marcelotrevisani (approved):

beckermr (commented):
We should state explicitly if we allow patching and if the values in this file taken precedence over the ones in the artifacts.

jezdez (approved):

wolfv:
good point @beckermr! There is also `pin_as_run` in the `conda_build_config.yaml` file. I believe that that one actually takes precedence over any run exports. Might be good to mention that in the CEP as well.

jaimergp:
> We should state explicitly if we allow patching and if the values in this file taken precedence over the ones in the artifacts.

Right now `conda-build` (and I would imagine other implementations) have not been notified of `run_exports.json` existence, so I'd say that initially it won't have an impact on building packages. The file is there as a convenience for automation infrastructure. Taken that into account, I'd be inclined to say that the file is not going to be patched _for now_. 

If the community decides that's convenient, then we can create another (smaller) CEP to change that. I think I have read feedback from some users (cannot recall who, sorry) who were against run_exports patching.

jaimergp:
> There is also pin_as_run in the conda_build_config.yaml file

TIL! I'll check how it works and add it to the CEP. Thanks Wolf!

jaimergp:
> I am leaning towards adding an empty run exports for all packages (instead of leaving out those that don't have it) so that one knows that these were considered.

@wolfv, my thinking here is that the source of truth for all available packages is still `repodata.json`. `run_exports.json` simply extends that with extra fields for _some_. If we enumerate all packages in `run_exports.json` then we also have to account for removals with "marked as broken" and so on. Does that sound reasonable?

jaimergp:
Added my current view regarding precedence and patching to the CEP. Happy to revisit, but for now I think the easiest is to consider `run_exports.json` a mirror for the info already available in the archives. This means it should not be patched or given a special treatment wrt pinning resolution in conda-build tooling.

cc @beckermr @wolfv

jaimergp:
For the record, I plan to call a vote for this in two weeks (next conda community call), so if you have any comments before the vote, this is the time. Thanks! 🙏

cep-9999.md
```diff
@@ -0,0 +1,165 @@
...
+
+We could maintain the status quo and ask downstream infrastructure to maintain their own database. However, this is a burden on them, and it is trivial to generate this data.
+
+We also considered _adding_ the `run_exports` metadata to `repodata.json`, but this has a few shortcomings:
+
+- It would require extending the `repodata` schema, currently not formally standardized.
+- It would increase the size of the already heavy `repodata.json` files.
+- (Typed) repodata parsers would need to be updated to handle the new field.
+
+Finally, we studied whether the `run_exports` metadata already present in the `channeldata.json` metadata would be enough. However, this metadata is only presented per version (not build), so it is insufficient and incomplete.
```
chenghlee, line 143: Should we add a sentence saying extending the `channeldata.json` format to add per-build information risks breaking conda-build's `--use-channeldata` option? (We should have a separate discussion about deprecating the `--use-channeldata` option.)

cep-9999.md
```diff
@@ -0,0 +1,165 @@
...
+
+Since `conda-index` already processes the package metadata to generate `repodata.json`, it would be trivial to also generate `run_exports.json` and serve them together.
+
+## Precedence and role in the pinning resolution process
+
+This file is not meant to replace the `run_exports` metadata already present in the package archives. It merely presents that information in a more convenient way. 
+`conda-build`-like clients are free to use either the `run_exports` metadata in the archives or the one in `run_exports.json`, since they should be equivalent.
+Special keys like `pin_run_as_build` will keep their behavior, since `run_exports.json` does not add a new level of precedence in the pinning resolution process. Again, it's an equivalent source for information already present in the archives.
+
+This means that `run_exports.json` should not be patched (like it is done with `repodata.json`);
```
chenghlee, line 30: I'm wondering if we should adopt [IETF BCP 14 convention](https://datatracker.ietf.org/doc/html/rfc8174#section-2) and use an explicitly capitalized "SHOULD NOT" here. (Or possibly a stronger "MUST NOT".)

cep-9999.md
```diff
@@ -0,0 +1,165 @@
...
+                "weak_constrains": [
+                    "string",
+                ],
+            }
+        }
+    }
+}
+```
+
+If a package does not define `run_exports`, the corresponding entry in `packages` or `packages.conda` will be an empty `run_exports` item:
```
chenghlee, line 103: If we decide to adopt BCP 14 conventions, I suggest replacing "...will be an empty..." with "...MUST be an empty...".

cep-9999.md
```diff
@@ -0,0 +1,165 @@
...
+- It would increase the size of the already heavy `repodata.json` files.
+- (Typed) repodata parsers would need to be updated to handle the new field.
+
+Finally, we studied whether the `run_exports` metadata already present in the `channeldata.json` metadata would be enough. However, this metadata is only presented per version (not build), so it is insufficient and incomplete.
+
+## FAQ
+
+### Which packages should be included?
+
+`run_exports` metadata is optional and not all packages have it. Should we only list artifacts that define it? Or should we list all artifacts, and have an empty `run_exports` field for those that don't define it?
```
chenghlee, line 149: Since we've required empty entries above, I suggest rephrasing this to say something like "all packages in `repodata.json` should be included".

chenghlee:
Suggested some minor tweaks, but I am in favor of this proposal.

jaimergp:
Thank you @chenghlee! Added your points in the new commits.

jaimergp:
@conda-incubator/steering, I would like to call a vote for this CEP.

This vote falls under the "Enhancement Proposal Approval" policy of the [conda governance policy](https://github.com/conda-incubator/governance),

It needs 60% of the Steering Council to vote yes to pass.

To vote, please leave yes, no or abstain as comments below.

This vote will end on August 3rd, 2023 End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).

ocefpaf (approved):

goanpeca (approved):

kkraus14 (approved):

jezdez:
yes

chenghlee (approved):

chenghlee:
Yes

wolfv:
yes

wolfv (approved):

marcelotrevisani (approved):

marcelotrevisani:
yes!

jezdez:
> This vote will end on August 3rd, 2023 End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).

@jaimergp Any reason to expand the vote from 7 to 16 days BTW?

jaimergp:
@jezdez To be able to invoke the timeout quorum rules and have two in-call reminders 😁

xhochy (approved):
Yes!

cep-9999.md
```diff
@@ -0,0 +1,171 @@
...
+Since `conda-index` already processes the package metadata to generate `repodata.json`, it would be trivial to also generate `run_exports.json` and serve them together.
+
+## Precedence and role in the pinning resolution process
+
+This file is not meant to replace the `run_exports` metadata already present in the package archives. It merely presents that information in a more convenient way. 
+`conda-build`-like clients are free to use either the `run_exports` metadata in the archives or the one in `run_exports.json`, since they MUST be equivalent.
+Special keys like `pin_run_as_build` MUST keep their behavior, since `run_exports.json` does not add a new level of precedence in the pinning resolution process. Again, it's an equivalent source for information already present in the archives.
+
+This means that `run_exports.json` MUST NOT be patched (like it is done with `repodata.json`);
+it MUST always reflect the metadata present in the archives.
```
msarahan, line 35: I wasn't sure about this, because it seemed like recipes could rot and become broken without the repo data patches. I think we are ok, because the repo data patches still apply when the build/host/run environments are created. Is it worthwhile to add a note about this eventual patching here, or is that already clear enough to everyone (except maybe me)?

msarahan (approved):

cep-9999.md
jaimergp: I am not sure I follow. This CEP tries to say that `run_exports.json` won't be patched for the sake of simplicity of scope, but we can always submit a new CEP overriding that decision if we ever arrive to a point where we need run_exports patching.

The previous discussion around it seemed to contain mixed opinions about whether patching run-exports was a good idea or not, so for now we are leaving things as they are.

What kind of wording would help clarify the intent here? I am happy to add it, but changing the intent of the CEP is a different thing given how far into the vote we are already.

Thanks for reviewing!

msarahan: I agree with the decision to not patch it right now, and I think it might be fine to never patch it. The only explanation I would add is that this decision is not ignoring patching, it is deferring it to the environment creation step. If this is shown to create invalid environments, then a follow on CEP would be needed to apply patching here. You could probably prove that one way or the other with a thought experiment now, but I think the incremental approach is a good enough start.

jaimergp: Thanks for the explanation! What about adding this sentence:

> Note this only applies to the served `run_exports.json` file. It does not try to regulate what `conda-build`-like tools can do; e.g. they might need to apply modifications to the `run_exports` metadata during the `build` and `host` environments setup.

msarahan: How about:

Note this only applies to the served run_exports.json file. It does not try to regulate what conda-build-like tools can do at environment creation time. They might need to apply modifications analogous to repodata patching to the run_exports metadata during the build, host and test environments setup. If patching run _exports.json is shown to be necessary for correct environment creation, it will be the subject of another CEP.

jaimergp:
Yes.

msarahan (approved):
Yes

jakirkham (approved):
Yes please!

jaimergp:
Hello team! The vote concluded last week on Aug 3rd at 23:59 AOE (Aug 4th 11:59 UTC). I have now reviewed the votes and these are the results:

* Yes: 12
* No: 0
* Abstain: 0
* Available quorum: [14](https://github.com/conda-incubator/governance/blob/9d40caa6a6a9ca889fbecce3a2f5b57653578662/steering.csv)
* Criteria: 60% of quorum
* 12/14 = 85.7%
* ✅ The vote has **passed**!

--

I will now update the PR to reflect the status and mint a CEP number. Since the vote for this PR was started after #8, this CEP will be minted the **number 11**.

jaimergp:
Sorry, small mistake, this CEP will get **number 12**, not 11.

PR #53, opened 2023-05-24
add new build format cep

wolfv:
This is a mashup of the boa / rattler-build and conda-build documentation for the new recipe spec.

I am not sure in how far we want to spec out things here since there are two speccable things:

- allowed inputs
- expected behavior

This is already a bit of both. What's missing is for example a section on run_exports, including the `ignore_run_exports_from` and `ignore_run_exports`. However, when specifying these inputs, we probably also want to specify the behavior a little bit?

I also need to expand more on the expected behavior of `outputs` and how "merging" works with the top-level recipe.

Todo:

- [ ] document all build keys
- [ ] spec out variant config files and variant evaluation
- [ ] document all available selector variables (unix, win, ...)?
- [ ] figure out if documenting package hash computation is part of this spec or not
- [ ] finish jinja function docs incl. all allowed inputs
- [ ] think more about `cmp` function

### DIFF
--- a/cep-20.md
+++ b/cep-20.md
@@ -1,0 +1,876 @@
+# 
+
+<table>
+<tr><td> Title </td><td> A new recipe format </td>
+<tr><td> Status </td><td> Proposed</td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> May 23, 2023</td></tr>
+<tr><td> Updated </td><td> May 23, 2023</td></tr>
+<tr><td> Discussion </td><td>  </td></tr>
+<tr><td> Implementation </td><td>https://github.com/prefix-dev/rattler-build</td></tr>
+</table>
+
+## Abstract
+
+We propose a new recipe format that is heavily inspired by conda-build. The main changes are: pure YAML, no non-intuitive "outputs", no comments-as-selectors.
+
+## Motivation
+
+The conda-build format has grown over the years to become quite complex. Unfortunately it has never been formally "specified" and it has grown some features over time that make it hard to parse as straightforward YAML.
+
+The CEP attempts to introduce a subset of the conda build format that allows for fast parsing and building of recipes.
+
+### History
+
+A discussion was started on what a new recipe spec could or should look like. The fragments of this discussion can be found here: https://github.com/mamba-org/conda-specs/blob/master/proposed_specs/recipe.md
+The reason for a new spec are:
+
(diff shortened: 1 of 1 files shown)


cep-20.md
```diff
@@ -0,0 +1,876 @@
...
+In addition to the runtime requirements, you can specify requirements
+needed during testing. The runtime requirements that you specified in
+the "run" section described above are automatically included during
+testing.
+
+```yaml
+test:
+  requires:
+    - nose
+```
```
dhirschfeld, line 653: Just mentioning one of my personal bugbears; if we're making changes, I think specifying test dependencies under `requirements/test` makes a lot more sense than special-casing them under their own top-level heading.
e.g.
```yaml
requirements:
  build:
    - "{{ compiler('c') }}"
    - sel(linux): "{{ cdt('xorg-x11-proto-devel') }}"
  host:
    - python
  run:
    - python
    - sel(py26): argparse
    - six >=1.8.0
  run_constrained:
    - optional-subpackage =={{ version }}
  test:
    - nose
```

cep-20.md
```diff
@@ -0,0 +1,876 @@
...
+- "3.6.* *cpython" -> true (build strings are ignored when not specified in the version spec)
+```
+
+Preprocessing selectors
+-----------------------
+
+You can add selectors to any item, and the selector is evaluated in
+a preprocessing stage. If a selector evaluates to `true`, the item is
+flattened into the parent element. If a selector evaluates to `false`,
+the item is removed.
```
jaimergp, line 100: Can selectors be nested? Should they?

```yaml
source:
  - sel(not win):
      url: http://path/to/unix/source
      sel(py3k):
        url: http://path/to/python3/bindings/linux/only
  - sel(win):
      url: http://path/to/windows/source
      sel(py3k):
        url: http://path/to/python3/bindings/windows/only

```

wolfv:
I've decided to split the proposal up into multiple sub-proposals, the first one is

1. About the new YAML format incl. selectors, allowed Jinja, ... #54 
2. The allowed keys / values in the new format (excluding `outputs`)
3. The new specification for how `outputs` should behave

If someone thinks that we should split into more parts let me know! :)

cep-20.md
```diff
@@ -0,0 +1,876 @@
...
+build:
+  noarch: python
+```
+
+Noarch Python packages are post-processed as follows:
+
+- `site-packages`  folder is moved to the root of the package
+- `bin/` and `Scripts/` files (that are _not_ entrypoints) are moved to `python-bin` folder in the root of the package
+
+These are moved to the appropriate Python-version specific locations when installing.
```
tdejager, line 483:
```suggestion
These can be moved to the appropriate Python-version specific locations when installing.
```

As this spec does not dictate anything about installing right?

wolfv:
Closing because we've split this CEP into multiple smaller ones :)

PR #54, opened 2023-06-02
A new YAML based format for "conda-build" files

wolfv:
✏️ Rendered version here: https://github.com/wolfv/ceps/blob/cep_20.1/cep-20.1.md

### DIFF
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -44,7 +44,7 @@ The reason for a new spec are:
 
 - no full Jinja2 support: no conditional or `{% set ...` support, only string
   interpolation. Variables can be set in the toplevel "context" which is valid
-  YAML
+  YAML (all new features should be native to YAML specs)
 - Jinja string interpolation needs to be quoted at the beginning of a string,
   e.g. `- "{{ version }}"` in order for it to be valid YAML
 - Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `-
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -164,3 +164,72 @@ requirements:
   build:
     - "{{ compiler('cxx') }}"
 ```
+
+## Examples
+
+### xtensor
+
+Original recipe [found here](https://github.com/conda-forge/xtensor-feedstock/blob/feaa4fd8015f96038168a9d67d69eaf06a36d63f/recipe/meta.yaml).
+
+```yaml
+context:
+  name: xtensor
+  version: 0.24.6
+  sha256: f87259b51aabafdd1183947747edfff4cff75d55375334f2e81cee6dc68ef655
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  fn: "{{ name }}-{{ version }}.tar.gz"
+  url: https://github.com/xtensor-stack/xtensor/archive/{{ version }}.tar.gz
+  sha256: "{{ sha256 }}"
+
+build:
+  number: 0
+  sel(win and vc<14):
+    skip: true
+
+requirements:
+  build:
+    - "{{ compiler('cxx') }}"
+    - cmake
+    - sel(unix): make
+  host:
+    - xtl >=0.7,<0.8
+  run:
+    - xtl >=0.7,<0.8
+  run_constrained:
+    - xsimd >=8.0.3,<10
+
+test:
+  commands:
+    sel(unix):
+      - test -d ${PREFIX}/include/xtensor
+      - test -f ${PREFIX}/include/xtensor/xarray.hpp
+      - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfig.cmake
+      - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfigVersion.cmake
+    sel(win):
+      - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
+      - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfig.cmake (exit 1)
+      - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
+
+about:
+  home: https://github.com/xtensor-stack/xtensor
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE
+  summary: The C++ tensor algebra library
+  description: Multi dimensional arrays with broadcasting and lazy computing
+  doc_url: https://xtensor.readthedocs.io
+  dev_url: https://github.com/xtensor-stack/xtensor
+
+extra:
+  recipe-maintainers:
+    - SylvainCorlay
+    - JohanMabille
+    - wolfv
+    - davidbrochart
+```
+
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -2,11 +2,11 @@
 
 <table>
 <tr><td> Title </td><td> A new recipe format </td>
-<tr><td> Status </td><td> Proposed</td></tr>
+<tr><td> Status </td><td> Proposed </td></tr>
 <tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
 <tr><td> Created </td><td> May 23, 2023</td></tr>
-<tr><td> Updated </td><td> May 23, 2023</td></tr>
-<tr><td> Discussion </td><td>  </td></tr>
+<tr><td> Updated </td><td> October 20, 2023</td></tr>
+<tr><td> Discussion </td><td>https://github.com/conda-incubator/ceps/pull/54</td></tr>
 <tr><td> Implementation </td><td>https://github.com/prefix-dev/rattler-build</td></tr>
 </table>
 
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -12,28 +12,42 @@
 
 ## Abstract
 
-We propose a new recipe format that is heavily inspired by conda-build. The main change is a pure YAML format without arbitrary Jinja or comments with semantic meaning.
+We propose a new recipe format that is heavily inspired by conda-build. The main
+change is a pure YAML format without arbitrary Jinja or comments with semantic
+meaning.
 
 ## Motivation
 
-The conda-build format has grown over the years to become quite complex. Unfortunately it has never been formally "specified" and it has grown some features over time that make it hard to parse as straightforward YAML.
+The conda-build format has grown over the years to become quite complex.
+Unfortunately it has never been formally "specified" and it has grown some
+features over time that make it hard to parse as straightforward YAML.
 
-The CEP attempts to introduce a subset of the conda build format that allows for fast parsing and building of recipes.
+The CEP attempts to introduce a subset of the conda build format that allows for
+fast parsing and building of recipes.
 
 ### History
 
-A discussion was started on what a new recipe spec could or should look like. The fragments of this discussion can be found here: https://github.com/mamba-org/conda-specs/blob/master/proposed_specs/recipe.md
+A discussion was started on what a new recipe spec could or should look like.
+The fragments of this discussion can be found here:
+https://github.com/mamba-org/conda-specs/blob/master/proposed_specs/recipe.md
 The reason for a new spec are:
 
-- Make it easier to parse ("pure yaml"). conda-build uses a mix of comments and jinja to achieve a great deal of flexibility, but it's hard to parse the recipe with a computer
-- iron out some inconsistencies around multiple outputs (build vs. build/script and more)
+- Make it easier to parse ("pure yaml"). conda-build uses a mix of comments and
+  jinja to achieve a great deal of flexibility, but it's hard to parse the
+  recipe with a computer
+- iron out some inconsistencies around multiple outputs (build vs. build/script
+  and more)
 - remove any need for recursive parsing & solving
 
 ## Major differences with conda-build
 
-- no full Jinja2 support: no conditional or `{% set ...` support, only string interpolation. Variables can be set in the toplevel "context" which is valid YAML
-- Jinja string interpolation needs to be quoted at the beginning of a string, e.g. `- "{{ version }}"` in order for it to be valid YAML
-- Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `- sel(osx): somepkg` instead of `- somepkg  # [osx]`
+- no full Jinja2 support: no conditional or `{% set ...` support, only string
+  interpolation. Variables can be set in the toplevel "context" which is valid
+  YAML
+- Jinja string interpolation needs to be quoted at the beginning of a string,
+  e.g. `- "{{ version }}"` in order for it to be valid YAML
+- Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `-
+  sel(osx): somepkg` instead of `- somepkg  # [osx]`
 
 ## Selectors
 
@@ -41,7 +55,8 @@ Selectors in the new spec take the following format:
 
 `sel(unix): selected_value`
 
-This is a valid YAML dictionary. Selector contents are simple boolean expressions and follow Python syntax. The following selectors are all valid:
+This is a valid YAML dictionary. Selector contents are simple boolean
+expressions and follow Python syntax. The following selectors are all valid:
 
 ```
 win and arm64
@@ -51,7 +66,8 @@ something == "test"
 
 ### The cmp function for variant selection
 
-Furthermore, we have a special "cmp" function that can be used to run a check against a selected variant version. The `cmp` function looks like the following:
+Furthermore, we have a special "cmp" function that can be used to run a check
+against a selected variant version. The `cmp` function looks like the following:
 
 ```
 cmp(python, "3.6")
@@ -68,14 +84,16 @@ requirements:
     - sel(cmp(python, ">=3.6,<3.10")): dataclasses
 ```
 
-This functionality generalizes and replaces the previous special variables such as `py2k`, `py3k`, `py36`, `py37`, and works just as well for NumPy, Ruby, R, or any other variant that might be of interest in the future.
+This functionality generalizes and replaces the previous special variables such
+as `py2k`, `py3k`, `py36`, `py37`, and works just as well for NumPy, Ruby, R, or
+any other variant that might be of interest in the future.
 
 ### Preprocessing selectors
 
-You can add selectors to any item, and the selector is evaluated in
-a preprocessing stage. If a selector evaluates to `true`, the item is
-flattened into the parent element. If a selector evaluates to `false`,
-the item is removed.
+You can add selectors to any item, and the selector is evaluated in a
+preprocessing stage. If a selector evaluates to `true`, the item is flattened
+into the parent element. If a selector evaluates to `false`, the item is
+removed.
 
 ```yaml
 source:
@@ -85,8 +103,7 @@ source:
       url: http://path/to/windows/source
 ```
 
-Because the selector is a valid Jinja expression, complicated logic
-is possible:
+Because the selector is a valid Jinja expression, complicated logic is possible:
 
 ```yaml
 source:
@@ -98,7 +115,8 @@ source:
       url: http://path/to/python3/unix/source
 ```
 
-Lists are automatically "merged" upwards, so it is possible to group multiple items under a single selector:
+Lists are automatically "merged" upwards, so it is possible to group multiple
+items under a single selector:
 
 ```yaml
 test:
@@ -145,9 +163,10 @@ source:
 
 Jinja has built-in support for some common string manipulations.
 
-In the new spec, complex Jinja is completely disallowed as we try to produce YAML that is valid at all times.
-So you should not use any `{% if ... %}` or similar Jinja constructs that produce invalid yaml.
-Furthermore, quotes need to be applied when starting a value with double-curly brackets like so:
+In the new spec, complex Jinja is completely disallowed as we try to produce
+YAML that is valid at all times. So you should not use any `{% if ... %}` or
+similar Jinja constructs that produce invalid yaml. Furthermore, quotes need to
+be applied when starting a value with double-curly brackets like so:
 
 ```yaml
 package:
@@ -155,9 +174,9 @@ package:
   name: "{{ name }}" # correct
 ```
 
-Some Jinja functions remain valid, but this is out of the scope of this spec. However, as an example, 
-the `compiler` Jinja function will still be available, with the main difference being the quoting of the
-brackets.
+Some Jinja functions remain valid, but this is out of the scope of this spec.
+However, as an example, the `compiler` Jinja function will still be available,
+with the main difference being the quoting of the brackets.
 
 ```yaml
 requirements:
@@ -165,11 +184,43 @@ requirements:
     - "{{ compiler('cxx') }}"
 ```
 
+## Shortcomings
+
+Since we deliberately limit the amount of "Jinja" that is allowed in recipes
+there will be several shortcomings.
+
+For example, using a `{% for ... %}` loop is prohibited. After searching through
+the conda-forge recipes with the Github search, we found for loops mostly used
+in tests. 
+
+In our view, for loops are a nice helper, but not necessary for many tasks: the
+same functionality could be achieved in a testing script, for example. At the
+same time we also plan to formalize a more powerful testing harness ([prototyped
+in
+boa](https://github.com/mamba-org/boa/blob/main/docs/source/recipe_spec.md#check-for-file-existence-in-the-final-package)).
+
+This could be used instead of a for loop to check the existence of shared
+libraries or header files cross-platform (instead of relying on Jinja templates
+as done
+[here](https://github.com/conda-forge/opencv-feedstock/blob/2fc7848655ca65419050336fe38fcfd87bec0649/recipe/meta.yaml#L131)
+or
+[here](https://github.com/conda-forge/boost-cpp-feedstock/blob/699cfb6ec3488da8586833b1500b69133f052b6f/recipe/meta.yaml#L53)).
+
+Other uses of `for` loops should be relatively easy to refactor, such as
+[here](https://github.com/conda-forge/libiio-feedstock/blob/1351e5846b753e4ee85624acf3a14aee4bcf321d/recipe/meta.yaml#L45-L51).
+
+However, since the new recipe format is "pure YAML" it is very easy to create
+and pre-process these files using a script, or even generating them with Python
+or any other scripting language. That means, many of the features that are
+currently done with Jinja could be done with a simple pre-processing step in the
+future.
+
 ## Examples
 
 ### xtensor
 
-Original recipe [found here](https://github.com/conda-forge/xtensor-feedstock/blob/feaa4fd8015f96038168a9d67d69eaf06a36d63f/recipe/meta.yaml).
+Original recipe [found
+here](https://github.com/conda-forge/xtensor-feedstock/blob/feaa4fd8015f96038168a9d67d69eaf06a36d63f/recipe/meta.yaml).
 
 ```yaml
 context:
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -1,0 +1,166 @@
+# A new recipe format – part 1
+
+<table>
+<tr><td> Title </td><td> A new recipe format </td>
+<tr><td> Status </td><td> Proposed</td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> May 23, 2023</td></tr>
+<tr><td> Updated </td><td> May 23, 2023</td></tr>
+<tr><td> Discussion </td><td>  </td></tr>
+<tr><td> Implementation </td><td>https://github.com/prefix-dev/rattler-build</td></tr>
+</table>
+
+## Abstract
+
+We propose a new recipe format that is heavily inspired by conda-build. The main change is a pure YAML format without arbitrary Jinja or comments with semantic meaning.
+
+## Motivation
+
+The conda-build format has grown over the years to become quite complex. Unfortunately it has never been formally "specified" and it has grown some features over time that make it hard to parse as straightforward YAML.
+
+The CEP attempts to introduce a subset of the conda build format that allows for fast parsing and building of recipes.
+
+### History
+
+A discussion was started on what a new recipe spec could or should look like. The fragments of this discussion can be found here: https://github.com/mamba-org/conda-specs/blob/master/proposed_specs/recipe.md
+The reason for a new spec are:
+
+- Make it easier to parse ("pure yaml"). conda-build uses a mix of comments and jinja to achieve a great deal of flexibility, but it's hard to parse the recipe with a computer
+- iron out some inconsistencies around multiple outputs (build vs. build/script and more)
+- remove any need for recursive parsing & solving
+
+## Major differences with conda-build
+
+- no full Jinja2 support: no conditional or `{% set ...` support, only string interpolation. Variables can be set in the toplevel "context" which is valid YAML
+- Jinja string interpolation needs to be quoted at the beginning of a string, e.g. `- "{{ version }}"` in order for it to be valid YAML
+- Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `- sel(osx): somepkg` instead of `- somepkg  # [osx]`
+
+## Selectors
+
+Selectors in the new spec take the following format:
+
+`sel(unix): selected_value`
+
+This is a valid YAML dictionary. Selector contents are simple boolean expressions and follow Python syntax. The following selectors are all valid:
+
+```
+win and arm64
+(osx or linux) and aarch64
+something == "test"
+```
+
+### The cmp function for variant selection
+
+Furthermore, we have a special "cmp" function that can be used to run a check against a selected variant version. The `cmp` function looks like the following:
+
+```
+cmp(python, "3.6")
+cmp(python, ">=3.6")
+cmp(python, ">=3.8,<3.10")
+etc
+```
+
+This can be used in a selector like so:
+
+```
+requirements:
+  build:
+    - sel(cmp(python, ">=3.6,<3.10")): dataclasses
+```
+
+This functionality generalizes and replaces the previous special variables such as `py2k`, `py3k`, `py36`, `py37`, and works just as well for NumPy, Ruby, R, or any other variant that might be of interest in the future.
+
+### Preprocessing selectors
+
+You can add selectors to any item, and the selector is evaluated in
+a preprocessing stage. If a selector evaluates to `true`, the item is
+flattened into the parent element. If a selector evaluates to `false`,
+the item is removed.
+
+```yaml
+source:
+  - sel(not win):
+      url: http://path/to/unix/source
+  - sel(win):
+      url: http://path/to/windows/source
+```
+
+Because the selector is a valid Jinja expression, complicated logic
+is possible:
+
+```yaml
+source:
+  - sel(win):
+      url: http://path/to/windows/source
+  - sel(unix and cmp(python, "2")):
+      url: http://path/to/python2/unix/source
+  - sel(unix and cmp(python, "3")):
+      url: http://path/to/python3/unix/source
+```
+
+Lists are automatically "merged" upwards, so it is possible to group multiple items under a single selector:
+
+```yaml
+test:
+  commands:
+    - sel(unix):
+      - test -d ${PREFIX}/include/xtensor
+      - test -f ${PREFIX}/include/xtensor/xarray.hpp
+      - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
+      - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
+    - sel(win):
+      - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
+      - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfig.cmake (exit 1)
+      - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
+
+# On unix this is rendered to:
+test:
+  commands:
+    - test -d ${PREFIX}/include/xtensor
+    - test -f ${PREFIX}/include/xtensor/xarray.hpp
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
+```
+
+## Templating with Jinja
+
+The spec supports simple Jinja templating in the `recipe.yaml` file.
+
+You can set up Jinja variables in the context YAML section:
+
+```yaml
+context:
+  name: "test"
+  version: "5.1.2"
+  major_version: "{{ version.split('.')[0] }}"
+```
+
+Later in your `recipe.yaml` you can use these values in string interpolation
+with Jinja. For example:
+
+```yaml
+source:
+  url: https://github.com/mamba-org/{{ name }}/v{{ version }}.tar.gz
+```
+
+Jinja has built-in support for some common string manipulations.
+
+In the new spec, complex Jinja is completely disallowed as we try to produce YAML that is valid at all times.
+So you should not use any `{% if ... %}` or similar Jinja constructs that produce invalid yaml.
+Furthermore, quotes need to be applied when starting a value with double-curly brackets like so:
+
+```yaml
+package:
+  name: {{ name }}   # WRONG: invalid yaml
+  name: "{{ name }}" # correct
+```
+
+Some Jinja functions remain valid, but this is out of the scope of this spec. However, as an example, 
+the `compiler` Jinja function will still be available, with the main difference being the quoting of the
+brackets.
+
+```yaml
+requirements:
+  build:
+    - "{{ compiler('cxx') }}"
+```
--- a/cep-20.1.md
+++ b/cep-20.1.md
@@ -38,25 +38,56 @@ The reason for a new spec are:
 - iron out some inconsistencies around multiple outputs (build vs. build/script
   and more)
 - remove any need for recursive parsing & solving
-- cater to needs for automation and dependency tree analysis via a determinstic format
+- cater to needs for automation and dependency tree analysis via a determinstic
+  format
 
 ## Major differences with conda-build
 
-- no full Jinja2 support: no conditional or `{% set ...` support, only string
+- no full Jinja2 support: no block support `{% set ...` support, only string
   interpolation. Variables can be set in the toplevel "context" which is valid
   YAML (all new features should be native to YAML specs)
-- Jinja string interpolation needs to be quoted at the beginning of a string,
-  e.g. `- "{{ version }}"` in order for it to be valid YAML
-- Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `-
-  sel(osx): somepkg` instead of `- somepkg  # [osx]`
+- Jinja variable syntax is changed to begin with `${{` so that it becomes valid
+  YAML, e.g. `- ${{ version }}`
+- Selectors use a YAML dictionary with `if / then / else` (vs. comments in
+  conda-build) and are only allowed in lists (dependencies, scripts, ...). The
+  syntax looks like:
+  ```yaml
+  - if: win
+    then: this
+    else: that  # optional
+  ```
+- for inline values, the Jinja ternary operator can be used, e.g. `number: ${{ 0
+  if linux else 100 }}`
 
 ## Selectors
 
-Selectors in the new spec take the following format:
+Selectors in the new spec are only allowed in lists and take an explicit `if /
+then / else` syntax.
 
-`sel(unix): selected_value`
+For example the following `script` section:
 
-This is a valid YAML dictionary. Selector contents are simple boolean
+```yaml
+script:
+  - if: unix
+    then: |
+      # this is the unix script
+  - if: win
+    then: |
+      @rem a script for batch
+```
+
+The same could have been expressed with an `else`:
+
+```yaml
+script:
+  - if: unix
+    then: |
+      # this is the unix script
+    else: |
+      @rem a script for batch
+```
+
+This is a valid YAML dictionary. Selector if statements are simple boolean
 expressions and follow Python syntax. The following selectors are all valid:
 
 ```
@@ -64,6 +95,22 @@ win and arm64
 (osx or linux) and aarch64
 something == "test"
 ```
+
+If the value of a selector statement is a list, it extends the "outer" list. For
+example:
+
+```yaml
+build:
+  - ${{ compiler('cxx') }}
+  - if: unix
+    then:
+      - make
+      - cmake
+      - pkg-config
+```
+
+evaluates for `unix == true` to a list with elements `[${{ compiler('cxx') }},
+make, cmake, pkg-config]`.
 
 ### The cmp function for variant selection
 
@@ -79,10 +126,13 @@ etc
 
 This can be used in a selector like so:
 
-```
+```yaml
 requirements:
   build:
-    - sel(cmp(python, ">=3.6,<3.10")): dataclasses
+    - if: cmp(python, ">=3.6,<3.10")
+      then: dataclasses
+    # The following syntax is also valid and equivalent
+    - ${{ "dataclasses" if cmp(python, ">=3.6,<3.10") }}
 ```
 
 This functionality generalizes and replaces the previous special variables such
@@ -98,21 +148,28 @@ removed.
 
 ```yaml
 source:
-  - sel(not win):
+  - if: not win
+    then:
+      # note that we omit the `-`, both is valid
       url: http://path/to/unix/source
-  - sel(win):
-      url: http://path/to/windows/source
+      sha256: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
+    else:
+      - url: http://path/to/windows/source
+        sha256: 06f961b802bc46ee168555f066d28f4f0e9afdf3f88174c1ee6f9de004fc30a0
 ```
 
 Because the selector is a valid Jinja expression, complicated logic is possible:
 
 ```yaml
 source:
-  - sel(win):
+  - if: win
+    then:
       url: http://path/to/windows/source
-  - sel(unix and cmp(python, "2")):
+  - if: (unix and cmp(python, "2"))
+    then: 
       url: http://path/to/python2/unix/source
-  - sel(unix and cmp(python, "3")):
+  - if: unix and cmp(python, "3")
+    then:
       url: http://path/to/python3/unix/source
 ```
 
@@ -122,15 +179,17 @@ items under a single selector:
 ```yaml
 test:
   commands:
-    - sel(unix):
-      - test -d ${PREFIX}/include/xtensor
-      - test -f ${PREFIX}/include/xtensor/xarray.hpp
-      - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
-      - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
-    - sel(win):
-      - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
-      - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfig.cmake (exit 1)
-      - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
+    - if: unix
+      then:
+        - test -d ${PREFIX}/include/xtensor
+        - test -f ${PREFIX}/include/xtensor/xarray.hpp
+        - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
+        - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
+    - if: win
+      then:
+        - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
+        - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfig.cmake (exit 1)
+        - if not exist %LIBRARY_PREFIX%\lib\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
 
 # On unix this is rendered to:
 test:
@@ -151,7 +210,7 @@ You can set up Jinja variables in the context YAML section:
 context:
   name: "test"
   version: "5.1.2"
-  major_version: "{{ version.split('.')[0] }}"
+  major_version: ${{ version.split('.')[0] }}
 ```
 
 Later in your `recipe.yaml` you can use these values in string interpolation
@@ -159,30 +218,31 @@ with Jinja. For example:
 
 ```yaml
 source:
-  url: https://github.com/mamba-org/{{ name }}/v{{ version }}.tar.gz
+  url: https://github.com/mamba-org/${{ name }}/v${{ version }}.tar.gz
 ```
 
 Jinja has built-in support for some common string manipulations.
 
 In the new spec, complex Jinja is completely disallowed as we try to produce
 YAML that is valid at all times. So you should not use any `{% if ... %}` or
-similar Jinja constructs that produce invalid yaml. Furthermore, quotes need to
-be applied when starting a value with double-curly brackets like so:
+similar Jinja constructs that produce invalid YAML. We also do not use the
+standard Jinja delimiters (`{{ .. }}`) because that is confused by the YAML
+parser as a dictionary. We follow Github Actions and others and use `${{ ... }}`
+instead:
 
 ```yaml
 package:
   name: {{ name }}   # WRONG: invalid yaml
-  name: "{{ name }}" # correct
+  name: ${{ name }} # correct
 ```
 
-Some Jinja functions remain valid, but this is out of the scope of this spec.
-However, as an example, the `compiler` Jinja function will still be available,
-with the main difference being the quoting of the brackets.
+Jinja functions work as usual. As an example, the `compiler` Jinja function will
+look like this:
 
 ```yaml
 requirements:
   build:
-    - "{{ compiler('cxx') }}"
+    - ${{ compiler('cxx') }}
 ```
 
 ## Shortcomings
@@ -216,6 +276,9 @@ or any other scripting language. That means, many of the features that are
 currently done with Jinja could be done with a simple pre-processing step in the
 future.
 
+Another option would be to allow "full" Jinja inside the test script text blocks
+(as long as it doesn't change the structure of the YAML).
+
 ## Examples
 
 ### xtensor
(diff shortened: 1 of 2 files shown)


marcelotrevisani:
I think that would be very useful, it will reduce the complexity to parse those yaml comments that are a pain.
But I also think that if we could move to use toml files instead of yaml mixed with jinja2 would be way better but that is another history 😬

wolfv:
How would TOML files work? Wouldn't we still need some Jinja here and there?

The good thing about the pure YAML format is that we can translate it to either the next evolution of the YAML format or a TOML based format quite easily. 

In fact, someone has _already_ added TOML support (more specifically recipe embedding [_inside_ a pyproject.toml file in `boa`](https://github.com/mamba-org/boa/pull/345) - precisely because it is possible).

marcelotrevisani:
> How would TOML files work? Wouldn't we still need some Jinja here and there?
> 
> The good thing about the pure YAML format is that we can translate it to either the next evolution of the YAML format or a TOML based format quite easily.
> 
> In fact, someone has _already_ added TOML support (more specifically recipe embedding [_inside_ a pyproject.toml file in `boa`](https://github.com/mamba-org/boa/pull/345) - precisely because it is possible).

regarding toml files I was thinking preprocess it and postprocess it if necessary
something like poetry but on steroids, 😆 
like
```toml
[requirements.host]
pytest = {version=">=5.0.0,<6.0.0", platform=["linux", "osx"], preprocess="func_to_preprocess", postprocess="func_to_postprocess"}
```

if it needs preprocessing it could call some predefined functions to do some magic that we might need

if we need variables to be used inside it can use
```toml
[env.variables]
my_var="abc"

[package]
name="${my_var}"

```
and it can inject it and expand when processing the recipe, if I am not mistaken there is a crate that does these kinda of env var expansion.

but as I mentioned, that is another whole history 😄 


What you have proposed I think would improve and simplify a lot of parsing and when we need to pragmatically create the recipes. But I wish to get ride of all the jinja, because they are also a pain sometimes 😆

marcelotrevisani (approved):

jezdez (commented):
Just some general remarks about the design, mostly about the use of chained functions for the selectors while evaluating the boolean expressions.

I'd prefer using YAML objects, given they enable us to evolve the items progressively over time and have a better user experience when having to chain many conditions. I also like the "blocky" look.

E.g. this becomes unwieldy when we'd add another function in the future via another spec update:

```
sel(cmp(python, "3.11") and foo(bar, "spam"))
```

While this would just be another line to optionally add:

```yaml
 requirements:
   build:
     - name: dataclasses
       if: python >=3.6 and python <3.10
       foo: bar == "spam"
     - cryptography
```

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+something == "test"
+```
+
+### The cmp function for variant selection
+
+Furthermore, we have a special "cmp" function that can be used to run a check
+against a selected variant version. The `cmp` function looks like the following:
+
+```
+cmp(python, "3.6")
```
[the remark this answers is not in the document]
wolfv: The problem is that operator overloading might be harder to implement vs. the `cmp` function. With the cmp function we'd know that the left hand side is the variant specifier and the right hand side is a version number. We could do `python >= Version("3.6")` basically, but that is even trickier.

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+We propose a new recipe format that is heavily inspired by conda-build. The main
+change is a pure YAML format without arbitrary Jinja or comments with semantic
+meaning.
+
+## Motivation
+
+The conda-build format has grown over the years to become quite complex.
+Unfortunately it has never been formally "specified" and it has grown some
+features over time that make it hard to parse as straightforward YAML.
+
```
travishathaway, line 24: I think one really important motivating factor that isn't mentioned in the motivation section is increasing the ease by which we can parse and modify these files (making them more machine readable friendly). I know that there is currently a desire to be able to do this, but the old format holds us back. It's very hard to parse and multiple attempts have already been made to try to do this if I'm not mistaken.

I know it's mentioned below, but it might be worth emphasizing here briefly too.

travishathaway:
One thing that also came to mind as reading the CEP was what the "migration strategy" would be for existing recipes. You mention this a little in the short comings section, but maybe it's worth it to have it's own section.

Unfortunately, migration will be a PITA because it will most likely involve manually updating recipes (will it?), but maybe there's a tool we could write that could handle translating 90% of the of existing recipes (I'm thinking like a Python 2 to 3 migration tool).

Or, maybe migration isn't even a concern 🤷‍♂️. It should probably be mentioned somewhere in the document.

wolfv:
Thanks for the comments @travishathaway. Indeed, migration is a concern. In `boa` we have a small & hacky tool that can convert simple recipes (`boa convert meta.yaml > recipe.yaml`).

It works by analyzing a line, and if it finds a comment it moves the comment right after the `-` and wraps it with a `sel(...)`. It also parses the `{% set ...` expressions and sticks them in a context dictionary. 

But obviously this works only for relatively straightforward cases.

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+- Jinja string interpolation needs to be quoted at the beginning of a string,
+  e.g. `- "{{ version }}"` in order for it to be valid YAML
+- Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. `-
+  sel(osx): somepkg` instead of `- somepkg  # [osx]`
+
+## Selectors
+
+Selectors in the new spec take the following format:
+
+`sel(unix): selected_value`
```
skupr-anaconda, line 56: How about a simplified variant:
`(unix): selected_value`
`(win and arm64)`
`((osx or linux) and aarch64)
`(something == "test")

wolfv: Looks like this is also valid YAML:

Screenshot 2023-06-07 at 14 49 00

However, it would effectively special case all dictionary keys that start with a round bracket. I think that would be OK, but less explicit than the `sel(...)` operator.

JeanChristopheMorinPerso:
I'm curious, the title says "A new recipe format – part 1". Is there a part 2 somewhere or some written notes about a possible future part 2?

wolfv:
Hi @JeanChristopheMorinPerso indeed, there is the "full" proposal (of which this is a derivative) here: https://github.com/conda-incubator/ceps/pull/53

Also, all this work is based on 2+ years of work on the [`boa`](https://github.com/mamba-org/boa) "build tool" as well as the [rattler-build build tool](https://github.com/prefix-dev/rattler-build). Both are many times faster than the current conda-build and implement this recipe format :)

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+The fragments of this discussion can be found here:
+https://github.com/mamba-org/conda-specs/blob/master/proposed_specs/recipe.md
+The reason for a new spec are:
+
+- Make it easier to parse ("pure yaml"). conda-build uses a mix of comments and
+  jinja to achieve a great deal of flexibility, but it's hard to parse the
+  recipe with a computer
+- iron out some inconsistencies around multiple outputs (build vs. build/script
+  and more)
+- remove any need for recursive parsing & solving
```
msarahan, line 40: The recursive parsing and solving was mainly to support things like inheriting specs, like run_constrained, from explicitly specified dependencies. Is there any kind of explicit spec for whether/how packages add/impose dependencies on sections outside of the one where they are specified?

What about the `pin_compatible` jinja? I'm getting the sense that it will be cut in this spec. I'm not sure whether you could impose limits on it to avoid the need to be recursive. It is very useful, so I hope it can stay somehow.

wolfv: In boa (and rattler-build) we have "deferred" evaluation for the pins, so that still works. We just require that the name of the pin is available, in order to build the proper topologically sorted build-graph :)

A dependency list in `rattler-build` contains an "enum" of type: 

https://github.com/prefix-dev/rattler-build/blob/665c981a392a6ebeba5842eee1d968b8753b2dcd/src/render/dependency_list.rs#L22-L27

Indeed, `pin_subpackage` is not yet implemented but quite trivial to achieve.

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+
+Since we deliberately limit the amount of "Jinja" that is allowed in recipes
+there will be several shortcomings.
+
+For example, using a `{% for ... %}` loop is prohibited. After searching through
+the conda-forge recipes with the Github search, we found for loops mostly used
+in tests. 
+
+In our view, for loops are a nice helper, but not necessary for many tasks: the
+same functionality could be achieved in a testing script, for example. At the
```
h-vetinari, line 197: While I agree that it's generally possible to move stuff to test scripts, this is IMO very disadvantageous:
* far away from meta.yaml (if they use the "magic" test script names, they're not even mentioned and then often not up-to-date because people forget to look at them)
* duplication between platforms (aside from the pain of writing loops in batch)

For example, it would be a real pain to rewrite the [following](https://github.com/conda-forge/grpc-cpp-feedstock/blob/main/recipe/meta.yaml)
<details>

```
        {% for each_lib in core_libs + core_cpp_libs + vendored_libs %}
        # presence of shared libs
        - test -f $PREFIX/lib/lib{{ each_lib }}.so              # [linux]
        - test -f $PREFIX/lib/lib{{ each_lib }}.dylib           # [osx]
        - if not exist %LIBRARY_BIN%\{{ each_lib }}.dll exit 1  # [win]
        - if not exist %LIBRARY_LIB%\{{ each_lib }}.lib exit 1  # [win]

        # absence of static libs (unix)
        - test ! -f $PREFIX/lib/lib{{ each_lib }}.a             # [unix]
        {% endfor %}

        # binaries
        {% for each_lang in binaries_plugin_langs %}
        - test -f $PREFIX/bin/grpc_{{ each_lang }}_plugin                    # [unix]
        - if not exist %LIBRARY_BIN%\grpc_{{ each_lang }}_plugin.exe exit 1  # [win]
        {% endfor %}

        # pkg-config (no metadata for vendored libs)
        {% for each_lib in core_libs %}
        - pkg-config --print-errors --exact-version "{{ core_version }}" {{ each_lib }}
        {% endfor %}
        {% for each_lib in core_cpp_libs %}
        - pkg-config --print-errors --exact-version "{{ version }}" {{ each_lib }}
        {% endfor %}
```

</details>

to the point that those tests would just not be written (or rather: maintained) as well as now (I [moved](https://github.com/conda-forge/grpc-cpp-feedstock/commit/80701ab5145ca8e5a3e84a6f7b31a2197b8c39a0) them out of the heavily outdated test scripts for exactly that reason).

Now, it's possible that this is still a worthwhile trade-off compared to what "pure YAML" gets us, but it will reduce the quality of our testing (at least, as someone who's added a lot of test coverage to feedstocks I got involved with, I believe it will strongly impact my output that front).

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+same time we also plan to formalize a more powerful testing harness ([prototyped
+in
+boa](https://github.com/mamba-org/boa/blob/main/docs/source/recipe_spec.md#check-for-file-existence-in-the-final-package)).
```
h-vetinari, line 200: I think this looks very nice, but it's critically missing a way to check for the _absence_ of files (which is crucial for many recipes, especially multi-output ones).

It would actually be a nice improvement over what conda currently does to check for absence of files _in the output under testing_, rather than just looking at the path in the prefix. That's because a given path might be satisfied -- intentionally -- by another output; for example, it's currently not possible to check[^1] that in a lib + bin + bindings setup, the binary output (that's depending on the lib) does not accidentally package any library artefacts.

[^1]: without an extra output

h-vetinari (commented):
I'll admit that I'm probably missing a bunch of context of _why_ moving to "pure YAML" is such a self-evidently good idea (though I'm a big fan of the mamba/boa effort!).

To me, jinja is a very standard templating engine, and it's not obvious from the OP (nor #53, nor the links I found), what problems we're solving.

So, not knowing what improvements we're "buying" with this, I mainly see the costs, which to me are definitely non-zero.

cep-20.1.md
wolfv: So there are two options:

- We can allow Jinja (plus the values from `context`) inside the script text blocks. I wouldn't mind that too much since it doesn't affect the recipe execution at all.
- You can rewrite your tests to take advantage of the new test helpers in boa. In that case it would look something like this:

```yaml
      exists:
        lib:
          - gpr
          - grpc
          - grpc_unsecure
          - grpc++
          - grpc++_unsecure
          - address_sorting
          - upb
        glob:
          # absence of static libs (unix)
          - ~ **/*.a
        bin:
          - grpc_cpp_plugin
          - grpc_csharp_plugin
          - grpc_node_plugin
          - grpc_objective_c_plugin
          - grpc_php_plugin
          - grpc_python_plugin
          - grpc_ruby_plugin
        pkg_config: 
          - grpc
          - grpc++
          - grpc_unsecure
          - grpc++_unsecure
      commands:
        # final CMake test to compile examples
        - ./test_grpc.sh   # [unix]
        - ./test_grpc.bat  # [win]
```

Note that this is implemented in a prototype-y fashion in boa and not yet implemented in rattler, and we'd spec this out in another CEP. Of course, there is duplication of the `core` libs. We could think of ways of being able to refer to lists that are defined elsewhere to make that less repetitive but IMO that can come later.

cep-20.1.md
wolfv: I just came up with a syntax to test the absence (negation) of the statement by using the `~` prefix. This shoudl be fairly straightforward to add. And as you suggested, these functions would operate on the `paths.json` _of the package_ (and not the environment).

But on the other hand this is a bit off topic since this is not the right CEP for that :)

cep-20.1.md
wolfv: We could also hypothetically allow lists to be defined in the context, and then use some other syntax to extend other lists. For example:

```yaml
context:
  core_libraries: ["grpc", "a", "b"...]

tests:
  exists:
    lib:
      - @ core_libraries
```

wolfv:
The current state of affairs is that we have 2 mini-languages + YAML (comments & Jinja) for the recipes. That makes it really hard to write a parser for it to statically extract information from the recipe or transform it in any way.

The way Jinja support is implemented in conda-build is also quite tricky: it does recursive evaluations of the recipe to find all the necessary values for the used Jinja variables. This contributes greatly to `conda-build` being extraordinary slow for larger recipes (especially with multiple outputs).

I think the way Jinja is used in conda-build currently is not very deliberate. It's usage is completely unrestricted, but most recipes only use a tiny subset (string interpolation & one or two "set" statements). However, the possibilities of Jinja blur the line between: is this YAML file a script or is this YAML file a static description of the package? Right now it's both, somehow, and not good at either.

If we have decide on a somewhat more limited YAML format as input, it will be very straight-forward to generate the recipes using Python or any other language (we are actually doing this for ROS recipes in the RoboStack project where we've been using boa for 2+ years now). So that can be an "escape-hatch" for some recipes that are more annoying to write without full Jinja support. At the same time, the old recipe format will be supported indefinitely so it's also no problem to stay with what is there.

Another nice property of the new recipe format will be the effortless evolution of it. Since it's static YAML we can come up with `version 2` and make a migrator that rewrites the recipe to the new spec - something that's super hard to do with the current format.

So to summarize the motivations:

- we get rid of a hybrid script/static thing
- we remove many complications in parsing & evaluating recipe files
- we only loose functionality for very few recipes for which we can figure out better ways
- generating and migrating recipes will be super straightforward

cep-20.1.md
wolfv: btw. conda doesn't do anything -- what people do is write ugly bash scripts that need special knowledge to handle Windows vs. Unix ... which is why testing is generally pretty bad on conda-forge. Another thing that we want to fix :) (but off-topic).

wolfv:
Another motivating example: we have a version autotick bot in the `boa-forge` collection of recipes and it requires _very_ little code (~600 lines of Python). The part about reading & writing the recipe is trivial: https://github.com/mamba-org/boa-forge/blob/de9a92f755f21f736e380b39bba925c755b1e35f/determine_versions.py#L607-L648

It produces PRs such as: https://github.com/mamba-org/boa-forge/pull/131/files

cep-20.1.md
jaimergp: What about the [gitignore syntax](https://git-scm.com/docs/gitignore#_pattern_format) (e.g. `!` for negation)? Devs are already familiar with that.

wolfv: https://stackoverflow.com/a/9703472

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+  commands:
+    - test -d ${PREFIX}/include/xtensor
+    - test -f ${PREFIX}/include/xtensor/xarray.hpp
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
+```
+
+## Templating with Jinja
+
+The spec supports simple Jinja templating in the `recipe.yaml` file.
```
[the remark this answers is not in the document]
wolfv: Simple means:

- Nothing that could go between `{% ...`, means no for loops, no `set`, no if / else etc.
- Everything that can go between `{{ ... }}` is still supported, including the `pipe` operator, so that `{{ version | lowercase }}` and things work. They can be assigned to helper variables inside the context, e.g. 

```yaml
context:
  version: "0.1.2.POST1234"
  lower_version: "{{ version | lower }}"  # evaluates to 0.1.2.post1234
```

- Inside the selectors, the Jinja expression language is used which only knows a few primitive types (str, int, float, ...).

This works fantastically with MiniJinja (which is what we use in `rattler-build`). However, MiniJinja would support _much_ more than what we are currently supporting.

baszalmstra:
Overall I think this is a great step. Having a more easily machine-readable recipe would open the doors for more tools to generate and parse recipes. 

I do also prefer the aforementioned `if` syntax over the `sel(..)` syntax because: 

* No need for the weird key formatting. `sel(..)` as a key looks strange in yaml land. 
* It's the same approach used by Github CI.
* I think it's a bit easier to read.
* We still use minijinja (or whatever) to evaluate the selectors which is still a second language. If we use the `if` syntax we could later also extend that to make it more machine-readable as well! 
    ```yaml
    if:
        or:
            - platform: win32 # not a minijinja expression
            - platform: osx
    ```

This is what it could look like (snippet is actually created by @wolfv , I only changed it slightly). Note that the value for `if` is also just a minijinja expression just without the `{{ }}` (this is also similar to Github CI).

```yaml
context:
  name: xtensor
  version: 0.24.6
  sha256: f87259b51aabafdd1183947747edfff4cff75d55375334f2e81cee6dc68ef655

package:
  name: "{{ name|lower }}"
  version: "{{ version }}"

source:
  fn: "{{ name }}-{{ version }}.tar.gz"
  url: https://github.com/xtensor-stack/xtensor/archive/{{ version }}.tar.gz
  sha256: "{{ sha256 }}"

build:
  number: 0
  skip:
    - if: osx or win # note that the value is a minijinja expression
      value: true
    - if: win
      # Defaults to: (because we can easily derive this from the context) 
      # value: true

requirements:
  build:
    - "{{ compiler('cxx') }}"
    - cmake
    - if: unix
      value: make
  host:
    - xtl >=0.7,<0.8
  run:
    - xtl >=0.7,<0.8
  run_constrained:
    - xsimd >=8.0.3,<10

test:
  commands:
    - if: unix
      value:
        - test -d ${PREFIX}/include/xtensor
        - test -f ${PREFIX}/include/xtensor/xarray.hpp
        - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfig.cmake
        - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfigVersion.cmake
    - if: win
      value:
        - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
        - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfig.cmake (exit 1)
        - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
```

wolfv:
I honestly don't care so much for either approach. I think the `sel(...)` syntax is more terse which might be nice.

There is one feature that is hard to express with the `if` syntax which is a dictionary with multiple keys, e.g.:

```yaml
build:
  number: 0
  script:
    sel(unix): |
        cmake ${CMAKE_ARGS} -DBUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=$PREFIX $SRC_DIR -DCMAKE_INSTALL_LIBDIR=lib
        make install
    sel(win): |
        cmake -G "NMake Makefiles" -D BUILD_TESTS=OFF -D CMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% %SRC_DIR%
        nmake
        nmake install

```

However, there was also a bit of internal debate about this because we would need to define the behavior if more than one `sel(...)` evaluates to `true`. In that case, in the current implementations, only the first one wins.

baszalmstra:
I don't see why that would be hard. Instead of a dictionary, just use a list:

```yaml
build:
  number: 0
  script:
    - if: unix
       value: |
         cmake ${CMAKE_ARGS} -DBUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=$PREFIX $SRC_DIR -DCMAKE_INSTALL_LIBDIR=lib
         make install
    - if: win
       value: |
         cmake -G "NMake Makefiles" -D BUILD_TESTS=OFF -D CMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% %SRC_DIR%
         nmake
         nmake install
```

However, I think we could do even better if the selector evaluation part is part of the model. The current implementation in Rust does two passes over the yaml. First, it evaluates the selectors which generates a new yaml file and then it parses it into a specific model format. While flattening the structure it can become unclear whether a list, a string, or an object was expected. 

I think it would be better to make the selectors part of an explicit model, evaluate that model with the selectors, lowering it into another (evaluated) model. This will most likely also give the user much better error messages if two selectors accidentally result in two valid values. It also allows us to generate a JSON schema that can be used in IDE's to get autocompletion support.

cep-20.1.md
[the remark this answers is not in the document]
AntoinePrv: +1 on `if` as it is consistent with CI providers as well

wolfv:
The biggest fear of the `if: ...` based syntax is that we now think that it's more "correct" or something like that, but then it might present a hurdle to adoption because in effect it's longer / more annoying to type :)

cep-20.1.md
h-vetinari: I'm happy if there's a way to test absence of files, syntax is secondary to me. That said:

> YAML is a complex spec

I'm not deeply enough in this whole discussion, but I'd be thrilled to move to TOML as well. One thing that scares me about the `sel` / `if` discussion is that YAML breaks horrifically if you get the indentation levels wrong within a list.

wolfv:
We had some more discussions today and came up with some changes:

- instead of forcing the user to quote `"{{ jinja }}"` variables (which always coerces them to be strings in YAML) we would rather go the Github route and use `${{ jinja }}` (which plays well with YAML), or use another bracket, like `((`. However we're leaning towards the Github syntax.
- this also enables us to use the ternary syntax in places where the YAML if would be complicated to use. Moreover, the ternary if is infinitely nestable. E.g. this is possible:
```yaml
build:
  number: ${{ 0 if win else 100 }}
```
and has the correct integer type etc.

We also looked at the challenge of the following example: suppose win needs an MD5 hash, and unix needs a SHA256 for a given source file (a bit contrieved, admittedly). With the `if` or `sel` syntax, the cleanest would be to repeat the block like:

```
source:
  - if: win
    url: https://github.com/killercup/cargo-edit/archive/refs/tags/v{{ version }}.tar.gz
    sha256: 46670295e2323fc2f826750cdcfb2692fbdbea87122fe530a07c50c8dba1d3d7
  - if: unix
    url: https://github.com/killercup/cargo-edit/archive/refs/tags/v{{ version }}.tar.gz
    md5: 0800fc577294c34e0b28ad2839435945   
```

Although in this case it would be possible to solve with a Jinja if as well.

For lists, the syntax would look like:

```yaml
    - if: win
      then:
      - cmake
      - ninja
      else:  # optional
      - ...
```

The missing `then` makes the syntax nicer for dictionaries, but also make it less homogenous.

cep-20.1.md
jakirkham: Sometimes the package name it isn't known. For example [cross-compilation usage]( https://github.com/conda-forge/tensorflow-feedstock/blob/2fdde9bd0977c9356ad192d684c8ca2751bf94ab/recipe/meta.yaml#L147 ) or [compilers themselves]( https://github.com/conda-forge/ctng-compilers-feedstock/blob/923b25fa711bd71cc7c43299e032bd87579dbce9/recipe/meta.yaml#L25-L37 )

cep-20.1.md
jakirkham: Not to bikeshed, but what about something like `is(unix)`?

cep-20.1.md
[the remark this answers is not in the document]
jakirkham: One caveat with `cmp` specifically is some languages use that to mean `cmp(x, 3)` to mean:

* `cmp(x, 3) > 0     # x > 3`
* `cmp(x, 3) == 0    # x == 3`
* `cmp(x, 3) < 0     # x < 3`

<hr>

Not to bikeshed, but maybe the critiques in the comments above (and the caveat around `cmp`) can be resolved with a different name? Some options:

* `within(...)`
* `contains(...)`
* `includes(...)`

Same syntax, but maybe this already implies the conditional case evaluating to `true`?

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+requirements:
+  build:
+    - sel(cmp(python, ">=3.6,<3.10")): dataclasses
+```
+
+This functionality generalizes and replaces the previous special variables such
+as `py2k`, `py3k`, `py36`, `py37`, and works just as well for NumPy, Ruby, R, or
+any other variant that might be of interest in the future.
```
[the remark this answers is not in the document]
jakirkham: One follow up question, can we assign these somehow?

Maybe users could roll their own `py36` or similar. This might help with readability when there are a lot of these

```yaml
context:
  py36 = cmp(python, ">=3.6,<3.10")
```

Another thought could users have multiple keys under a selector. For example

```yaml
  build:
    - sel(cmp(python, ">=3.6,<3.10")):
      - dataclasses
      - importlib_metadata
```

Maybe both together?


```yaml
context:
  py36 = cmp(python, ">=3.6,<3.10")

...

  build:
    - py36:
      - dataclasses
      - importlib_metadata
```

cep-20.1.md
```diff
@@ -0,0 +1,287 @@
...
+  - sel(win):
+      url: http://path/to/windows/source
+  - sel(unix and cmp(python, "2")):
+      url: http://path/to/python2/unix/source
+  - sel(unix and cmp(python, "3")):
+      url: http://path/to/python3/unix/source
+```
+
+Lists are automatically "merged" upwards, so it is possible to group multiple
+items under a single selector:
```
jakirkham, line 120: 🥳

cep-20.1.md
```diff
@@ -0,0 +1,287 @@
...
+  commands:
+    - test -d ${PREFIX}/include/xtensor
+    - test -f ${PREFIX}/include/xtensor/xarray.hpp
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfig.cmake
+    - test -f ${PREFIX}/lib/cmake/xtensor/xtensorConfigVersion.cmake
+```
+
+## Templating with Jinja
+
+The spec supports simple Jinja templating in the `recipe.yaml` file.
```
jakirkham, line 146: Honestly since we are doing a bit of a redesign here, think it is worth asking whether we still want Jinja to play a role here. In other words, if we were not already using Jinja today, how would we have solved these use cases? How do others approach YAML templating? Could we think of pure YAML solutions?

cep-20.1.md
jakirkham: Since many folks are already coming from Python, think significant whitespace is already an accepted thing

Not to say moving to a simpler markup language might be beneficial for us, but that seems like a separate discussion altogether. Right now Wolf is trying to just move use from 2 meta languages and 1 markup language to just 1 markup language, which would be a huge improvement over the status quo. Maybe a future discussion could be moving from 1 markup language to another (once we are fully in 1 :)

jakirkham:
Thanks Wolf! 🙏

Really appreciate you taking the time to write this up with nice examples. Think there are a lot of good improvements here. The biggest simply being in one markup language

Asked a few questions and made a few minor suggestions above. Will probably need to mull on this more and bring a few more examples into the discussion so we can think about what it would mean to move recipes into a format like the one proposed here

Another question that comes up for me frequently, which may be off-topic for this, but I'd like to at least put it out there. The multi-output format that we have today was a bolt on to an existing single output format. Though I wonder if we could do better by moving to some kind of model where the multi-output format is a first-class citizen. In any event this may be CEP in its own right, but I think this would be a valuable discussion to have and would be very beneficial for recipe maintainers. Something to think about 🙂

wolfv:
Hi @jakirkham thanks for your reply! Indeed, multi-outputs are one of the things I desperately want to fix and it's going to be the topic of another CEP. In that CEP, the idea will be to make the outputs completely independent (no weird inheritance of dependencies etc.) and have no-package outputs (that other outputs can depend on as a build cache). 

This one here focuses on the YAML format for future recipes.

We did go back to the drawing board a little and came up with the following:

Allow the conditional `if` syntax only in lists, such as dependency, script or or run exports. In all other places, users can use the "ternary" operator of MiniJinja. This will simplify the implementation quite a lot.

This is what an example recipe with proposed syntax would look like:

```yaml
context:
  name: xtensor
  version: 0.24.6
  sha256: f87259b51aabafdd1183947747edfff4cff75d55375334f2e81cee6dc68ef655

package:
  name: ${{ name|lower }}
  version: ${{ version }}

source:
  fn: ${{ name }}-${{ version }}.tar.gz
  url: https://github.com/xtensor-stack/xtensor/archive/${{ version }}.tar.gz
  sha256: ${{ sha256 }}

build:
  number: 0
  skip:
    # special case if -- raw "statements"
    # note that the value is a minijinja expression
    - osx or win 
    - win
  # alternative syntax
  # script: ${{ "build_cmd.sh" if unix else "build_cmd.bat" }}
  script:
    - if: unix
      then: |
        cmake ${CMAKE_ARGS} -DBUILD_TESTS=OFF -DCMAKE_INSTALL_PREFIX=$PREFIX $SRC_DIR -DCMAKE_INSTALL_LIBDIR=lib
        make install
    - if: win
      then: |
        cmake -G "NMake Makefiles" -D BUILD_TESTS=OFF -D CMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% %SRC_DIR%
        nmake
        nmake install

requirements:
  build:
    - ${{ compiler('cxx') }}
    - cmake
    - if: unix
      then: make
  host:
    - xtl >=0.7,<0.8
  run:
    - xtl >=0.7,<0.8
  run_constrained:
    - xsimd >=8.0.3,<10

test:
  commands:
    - if: unix
      then:
        - test -d ${PREFIX}/include/xtensor
        - test -f ${PREFIX}/include/xtensor/xarray.hpp
        - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfig.cmake
        - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfigVersion.cmake
    - if: win
      then:
        - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1)
        - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfig.cmake (exit 1)
        - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfigVersion.cmake (exit 1)
```

vyasr:
Hi all, I'm a [RAPIDS](rapids.ai/) developer. I see a few comments regarding the potential benefits of using "pure YAML", can add my two cents here and give a specific example as to why it could be of use to us (also a :+1: to @wolfv's example of a generated meta.yaml from RoboStack). We've been something similar but [focusing primarily on dependency management for now](https://github.com/rapidsai/dependency-file-generator/) in order to single source our dependency lists between multiple Python packages and C++ libraries contained in a monorepo, each of which must product some combination of environment.yaml/meta.yaml/pyproject.toml files in addition to some other more bespoke use cases. We have support for many of the use cases we care about, but we haven't supported conda recipes due to the complexities of parsing a file that is not pure YAML in a context where we may not have all the information needed to evaluate the full Jinja context (i.e. if there are variables available during a conda build that we don't know outside that environment). The best approach we've considered so far is to simply generate many small YAML files containing dependency lists corresponding to each section and then loading them with a Jinja for loop, but that's a pretty rough solution. Having a pure YAML file would make producing the dependency lists dynamically much simpler as stated in the CEP:

> However, since the new recipe format is "pure YAML" it is very easy to create and pre-process these files using a script, or even generating them with Python or any other scripting language. That means, many of the features that are currently done with Jinja could be done with a simple pre-processing step in the future. 

Note that pulling dependencies from pyproject.toml into meta.yaml is generally insufficient because the two sets of dependencies generally will not match exactly due to various factors (pyproject's build-requires typically splits into conda's build + host, non-Python library dependencies are typically not available on pip, differing package names or metapackage structures between conda and PyPI, etc). You may reduce some apparent redundancy, but in practice you still have to think about both files, and you have now coupled them in a nontrivial way since logic in one file now depends on the contents of the other.

wolfv:
Hmm, @isuruf this was the original "discussion document": https://hackmd.io/axI1tQdwQB2pTJKt5XdY5w 

Maybe we had some more discussion in one of the conda-forge meetings or on the conda-build issue tracker?

I found some discussion here: https://github.com/conda-forge/conda-forge.github.io/issues/1612

And [here](https://www.google.com/search?q=conda-forge+boa+recipe+spec+site%3Aconda-forge.org&sxsrf=APwXEdfKnYUyFzJNK3sdq5zi93rXNxJE5w%3A1686605637861&ei=RY-HZJWONK6Nxc8P6oOZqA0&ved=0ahUKEwjVxK7P177_AhWuRvEDHepBBtUQ4dUDCA8&uact=5&oq=conda-forge+boa+recipe+spec+site%3Aconda-forge.org&gs_lcp=Cgxnd3Mtd2l6LXNlcnAQAzoKCAAQRxDWBBCwAzoFCCEQoAE6BwghEKABEApKBAhBGABQzQFYvyZglidoAXAAeACAAYABiAHTDJIBBDE5LjOYAQCgAQHAAQHIAQQ&sclient=gws-wiz-serp) are all the boa mentions in conda-forge meetings.

cep-20.1.md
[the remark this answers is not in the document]
wolfv: I haven't thought about assigning a boolean variable in the context but it should indeed be possible. 

Regarding the list-merging, that is indeed mentioned in the CEP and what we want to implement.

wolfv:
I've update the CEP with the new syntax as discussed. We'll be implementing the `ListWithConditionals` in `rattler-build` today and give it a try :)

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+- remove any need for recursive parsing & solving
+- cater to needs for automation and dependency tree analysis via a determinstic
+  format
+
+## Major differences with conda-build
+
+- no full Jinja2 support: no block support `{% set ...` support, only string
+  interpolation. Variables can be set in the toplevel "context" which is valid
+  YAML (all new features should be native to YAML specs)
+- Jinja variable syntax is changed to begin with `${{` so that it becomes valid
```
jaimergp, line 49: Awesome and in line with what GHA uses too :D

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+  YAML, e.g. `- ${{ version }}`
+- Selectors use a YAML dictionary with `if / then / else` (vs. comments in
+  conda-build) and are only allowed in lists (dependencies, scripts, ...). The
+  syntax looks like:
+  ```yaml
+  - if: win
+    then: this
+    else: that  # optional
+  ```
```
jaimergp, line 58: Are `elif` blocks allowed? I am not even sure it's a good idea given that we can just use one more `if` block, but since we do have `else`... I wonder if it was discussed :D I guess it's get trickier syntax-wise, though.

wolfv: Not allowed because we could only have a single one (since there cannot be duplicate keys).

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+You can add selectors to any item, and the selector is evaluated in a
+preprocessing stage. If a selector evaluates to `true`, the item is flattened
+into the parent element. If a selector evaluates to `false`, the item is
+removed.
+
+```yaml
+source:
+  - if: not win
+    then:
+      # note that we omit the `-`, both is valid
```
jaimergp, line 153: I can see this becoming a bit confusing down the line. Are there cases were we do need a nested list that should not be flattened?

Maybe we can have one more key in the `if` dictionary? e.g. `flatten: true`?

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+  fn: ${{ name }}-${{ version }}.tar.gz"
+  url: https://github.com/xtensor-stack/xtensor/archive/${{ version }}.tar.gz
+  sha256: ${{ sha256 }}
+
+build:
+  number: 0
+  # note: in the new recipe format, `skip` is a list of conditional expressions
+  #       but for the "YAML format" discussion we pretend that we still use the 
+  #       `skip: bool` syntax
+  skip: ${{ true if (win and vc14) }}
```
jaimergp, line 309: Is this a valid expression? I had understood skip would take a list of selectors, so:

```suggestion
  skip:
    - win and vc14
```

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+  host:
+    - xtl >=0.7,<0.8
+  run:
+    - xtl >=0.7,<0.8
+  run_constrained:
+    - xsimd >=8.0.3,<10
+
+test:
+  commands:
+    if: unix
```
jaimergp, line 326:
```suggestion
  - if: unix
```

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+
+about:
+  home: https://github.com/xtensor-stack/xtensor
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE
+  summary: The C++ tensor algebra library
+  description: Multi dimensional arrays with broadcasting and lazy computing
+  doc_url: https://xtensor.readthedocs.io
+  dev_url: https://github.com/xtensor-stack/xtensor
```
jaimergp, line 346: Will minijinja be allowed anywhere, e.g. in the `about` section? We may want to think about whether some stuff should ALWAYS be static so there's no evaluation needed.

cep-20.1.md
wolfv: yes, but that's for another CEP! :)

cep-20.1.md
wolfv: No there is no case with nested lists allowed

cep-20.1.md
wolfv: We would continue to ship the rendered YAML as part of the package. I think jinja should be allowed everywhere, yeah.

cep-20.1.md
```diff
@@ -0,0 +1,352 @@
...
+  YAML, e.g. `- ${{ version }}`
+- Selectors use a YAML dictionary with `if / then / else` (vs. comments in
+  conda-build) and are only allowed in lists (dependencies, scripts, ...). The
+  syntax looks like:
+  ```yaml
+  - if: win
+    then: this
+    else: that  # optional
+  ```
+- for inline values, the Jinja ternary operator can be used, e.g. `number: ${{ 0
```
jaimergp, line 59: How crazy can these get? For example, are these ok?

- `${{ 0 if (linux or osx) else 1 }}`
- `${{ 0 if (linux or osx) else (100 if cuda else 50) }}`

What about defining new blocks in a Jinja expression?

```yaml
requirements:
  build:
    - ${{ compiler('cxx') }}
    - ${{ {"if": "unix", "then": "make"} if python_version >= (3,10) else None }}
```

cep-20.1.md
jaimergp: I was thinking of bots preprocessing some recipes for metadata collection without requiring the previously built artifacts, but I guess that can be restricted via linters in each specific deployment (e.g. conda-forge), not part of the standard itself.

cep-20.1.md
wolfv: Yes, nesting if statements works. You can make multiline strings and have crazy expressions if you want, e.g.

```yaml
- |
  ${{ "bla" if python_version >= (3,10) else 
      (100 if other else 
      (200 if that else "123123"
    }}
```

Minijinja has a nice online playground, btw: https://mitsuhiko.github.io/minijinja-playground/

We can enforce that only `bool, int, string` are allowed evaluation types of minijinja expressions (lists, dicts and floats would be forbidden).

cep-20.1.md
wolfv: It would also be easy to use `context + about` to evaluate some jinja, leave everything regarding variants as undefined, and render only the about section.

sterlinm:
This sounds very promising!

One thing I'm not entirely clear on is on is whether the Jinja2 functions [for loading data from other files described here](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#loading-data-from-other-files) will continue to work. I'm guessing yes?

wolfv:
@sterlinm I think in most implementations they would not be supported initially. Especially `load_setup_py` will definitely not be support in the Rust version (`rattler-build`). 

However, with the new format it is easier than ever to write a small script to create recipes from a script.

Also note that the proposal doesn't allow for something like `{% set contents = load_pyproject_toml(...) %}` because that syntax doesn't parse as YAML.

We could support some form of this inside the context, but I am not sure on the specifics yet.

cep-20.1.md
mikemhenry: I agree, what does jinja accomplish that we can't do with the other new features?

sterlinm:
@wolfv Thanks, that all makes sense!

For scenarios where you want to build a conda recipe and pypi project from the same list of packages, I agree that being able to generate both from simple scripts sounds better than having lots of magic functions in the recipe at the expense of the other goals you've laid out.

mbargull (changes requested):
Thanks for the proposal(s)!

Overall I like the approach, including structuring/splitting the discussion and proposals into multiple focused parts (I have yet to look into the other parts, though).

### 1. Base (serialization) format

As for the base serialization format, I think YAML is a good choice since it can convey the content in a comparatively succinct and readable manner and many people are familiar with it (be it via the current `conda-build` recipe format or, e.g., usage through automation tools/services like ones mentioned but others above).

That being said, YAML is -- despite looking simple on a first glance -- complex in comparison to other serialization formats.
With things like
- easy modifications via automation (as done, e.g., via conda-forge migrators),
- seamless (back-and-forth) conversion to other formats (in-memory or other serialization like mentioned TOML etc.),
- avoidance of ambiguities, etc.

in mind, I strongly suggest that we discourage or better yet forbid usage of some YAML-specific features like
1. Anchors and tags, see https://yaml.org/spec/1.2.2/#69-node-properties
   While it might seem attractive to some to use anchors to avoid repetitions etc., it makes the document more complex (non-tree-like structure) and prevents one-to-one mappings to other formats.
2. Multiple documents per stream, see https://yaml.org/spec/1.2.2/#chapter-9-document-stream-productions
   One could imagine to have a multi-document file that includes the `conda_build_config.yaml` in there, but again, this prevents direct conversion to other formats that don't have this concept.
3. Use any YAML schema apart from the "Failsafe Schema" (i.e., what `ruamel.yaml.BaseLoader` uses), see https://yaml.org/spec/1.2.2/#101-failsafe-schema
   This means only have string scalars and no implicit parsing of those to numbers/Booleans/etc. (e.g., unwanted conversion of version "numbers" to actual numbers).

### 2. Expression language

I'm happy to see the Jinja+line-selector text-templating from the current recipe format to be superseded by an expression language tucked into the outer structure.
Also happy there seems to be some consensus to use static-key-dictionary route for the conditional entries -- and with the proposed hoisting of list entries, I would assume most concerns w.r.t. the slightly more verbose syntax are also tackled.

I'd like us to have some clarifications/more exact language to avoid some ambiguities:
1. The text currently states that the conditional phrases "follow Python syntax".
   Is it intended to use the same language as in the `${{ }}` expressions?
   If yes, it would make sense to define an "expression language" (e.g., Jinja, as currently proposed) and then refer to that for conditionals and string interpolation uses.
   (N.B.: With the current "follow Python syntax" phrasing, this reads more like how `conda-build` processes line selectors, i.e., evaluate Python expressions.)
2. "Jinja variable syntax is changed to begin with `${{` so that it becomes valid YAML":
   <details><summary>There can still be some ambiguities for corner cases, e.g., around <code>:</code> followed by a space:</summary>
   
   ```python
   >>> import jinja2, ruamel.yaml
   >>> render_jinja = lambda s, **kwargs: jinja2.Template(s, variable_start_string="${{", variable_end_string="}}", undefined=jinja2.StrictUndefined).render(**kwargs)
   >>> load_yaml = lambda s: ruamel.yaml.YAML(typ="base").load(s)
   >>> x = lambda s, **kwargs: print(s, load_yaml(s), render_jinja(s, **kwargs), load_yaml(render_jinja(s, **kwargs)), "", sep="\n")
   >>> 
   >>> x('${{ [1,2]|join(":#") }}')
   ${{ [1,2]|join(":#") }}
   ${{ [1,2]|join(":#") }}
   1:#2
   1:#2
   
   >>> x('${{ [1,2]|join(": #") }}')
   ${{ [1,2]|join(": #") }}
   {'${{ [1,2]|join("': ''}
   1: #2
   {'1': ''}
   
   >>> x('${{ version.split(".")[:-1] | join(".") }}', version="1.2.3")
   ${{ version.split(".")[:-1] | join(".") }}
   ${{ version.split(".")[:-1] | join(".") }}
   1.2
   1.2
   
   >>> x('${{ version.split(".")[: -1] | join(".") }}', version="1.2.3")
   ${{ version.split(".")[: -1] | join(".") }}
   {'${{ version.split(".")[': '-1] | join(".") }}'}
   1.2
   1.2
   

   ```
   </details>
   So, to avoid that ambiguity, the proposal should explicitly state the order of evaluation of YAML/Jinja processing.
   I'm heavily in favor of always evaluating the expressions after the YAML is loaded, i.e., for the examples above require the user to add outer quotation marks or use "block scalar" style to avoid the inline `:[space]`.
3. The proposal mentions "Jinja templating" and as such gives the impression that it could still be used to template all of the recipe's text, i.e., dictionary keys could be composed with Jinja expressions and even whole sections could be added to the text before it gets loaded as YAML.
   I'd assume that only the actual scalars ("string values") are meant to have these expressions? We should clarify this in the proposal.
4. "Selectors in the new spec are only allowed in lists":
   A stated reason for this restriction is missing from the text.
   I'd assume the train of thought could have been that `{if: cond, then: [{key: value}]}` constructs wouldn't work since we'd have duplicate `if`/`then` keys for multiple `key`s?
   However, I don't believe the restriction is necessary because we'd actually have `{key: {if: cond, then: [value]}}` constructs.
   There are some use cases for this we have in some recipes, too, like:
   ```yaml
   source:
     url: https://some.url/${{ target_platform }}/package-${{ version }}.archive
     sha256:
       if: target_platform == "linux-64"
       then:
         - 123...
       else:
         - if: target_platform == "osx-64"
           then:
             - 456...
   ```
   Yes, this example could be written differently to circumvent the list-only restriction. Point being just that there shouldn't be the need for restricting where we could use the conditionals.
   (Then again: I'm more than happy to have more rather than less restrictions initially! We can always relax the format when needed, of course.)
5. To avoid common sources for errors in recipes, we might want to impose some constraints in the conditionals/expressions:
   1. For conditional expressions it would make sense to put a hard requirement on needing Boolean types.
      E.g., to avoid `skip: condition` with `condition = "no"` => `bool(condition) == bool("no") == True` => `skip: true` or similar.
   2. For all expressions, ensure valid types for comparisons.
      E.g., with examples for potential expressions inspired from this proposal's discussion:
      - `python > 3.10` would yield a typing error in a Python/Jinja context if (as is currently the case with `conda-build`) `python` is of type `str` and `3.10` written as a `float`.
      - `python == 3.10` would be worse since this silently evaluates to a falsy value for any string `python`.
   
   Point 1 should not be too hard to implement.
   Point 2 depends on what kind of agreement is reached for the version comparison logic.

### 3. Miscellaneous

1. Variant version comparison:
   There does not seem consensus about the naming/style/(semantics?) regarding version comparison/matching of variant keys.
   As others, I'm also not sold on the currently proposed `cmp` function.
   For the naming, I believe we should find something more descriptive -- problem here is the usual one: find a more descriptive/focused identifier which is also not too clumsy to get in the way... I don't have a concrete suggestion here :/
   For the problem about string/number comparisons we'd probably want the version constraints/specification to always be written as strings to avoid mismatched types. E.g., have a function for which we can call `version_match(python, "<3.10")` (exactly like the proposed `cmp`) or even `version_match("python >=3.10,<3.11.0a0")` with the latter retrieving the value for `python` on its own; sample implementation (handling both cases) for clarification:
   ```python
   >>> config = {"python": "3.10"}
   >>> import conda.models.match_spec
   >>> import conda.models.version
   >>> def version_match(version_or_match_spec: str, version_spec: str|None = None):
   ...     if version_spec:
   ...         vs = conda.models.version.VersionSpec(version_spec)
   ...         version = version_or_match_spec
   ...     else:
   ...         ms = conda.models.match_spec.MatchSpec(version_or_match_spec)
   ...         vs = ms.version
   ...         version = config[ms.name]
   ...     return vs.match(version)
   ... 
   >>> version_match(python, "<3.10")
   False
   >>> version_match("python >=3.10,<3.11.0a0")
   True
   ```
   The `python < Version("3.10")` variant is also something I always liked because of good readability, but having the comparison operator outside of the version specification string limits us to Python/Jinja/? compatible operators (e.g., `~=`, `=` would not be covered).
   Hence, being able to give a single string à la `python >=3.10,<3.11.0a0` seems preferable to me.
   Ergonomics (function name and/or special syntax for `if` (`if_match`?)) -- I'm not sure about.
   
   All in all: I think there is some more discussion needed and we could defer that decision to a subsequent proposal?
   I.e., for now, just change the wording to say something like "there is a version matching function(/functionality)" but not nail it down to the currently proposed `cmp`.
   
2. The example for the `context` section includes an important detail that is not described further:
   ```yaml
   context:
     name: "test"
     version: "5.1.2"
     major_version: ${{ version.split('.')[0] }}
   ```
   Here, `major_version` has a dependency on `version`. Since this a YAML mapping, the order of the entries is only given in the serialized form but not significant otherwise. Meaning, the values in `context` form a directed (dependency) graph and as such we need to impose an acyclicality condition (at least for values used in the recipe). Do we want to mention this?

3. The name `recipe.yaml` has been used but it hasn't been explicitly specified.
   I'm +1 on `recipe` vs `meta`.
   (Plus, I'd also favor to explicitly disallow other file extensions, i.e., not allow `.yml` if `.yaml` is used or vice versa; it's just slightly error-prone if one has to look for `.*\.ya?ml` in automation scripts, etc.)

4. Big +1 to move further aspects to subsequent discussions/proposals!
   (Like the format/inclusion of `conda_build_config.yaml`, new/different semantics for parts of the recipe (like the mentioned `tests` section), available functions/functionality in the (Jinja) expressions, etc. pp.)


----
### TL;DR

I'm supportive of the proposal and the above points are mostly to address ambiguity
- which should lessen pitfalls recipe writers could encounter otherwise, and
- automation/conversion could trip over.

wolfv:
Hi Marcel, big sorry for taking so long to get back on the points! 

- Yes, maybe it's better to say "Jinja syntax"
- The plan is to evaluate YAML first, and then Jinja. So keys can't contain Jinja. I should probably add that to the proposal if it wasn't clear.
- There is indeed an evaluation order in the context (it's parsed as an OrderedDict and evaluated top-to-bottom). I thought I had specified that somewhere, but I didn't mention anything about it having to be acyclical. I also didn't really think about it that way as I thought that it would be obvious that it's evaluated top-to-bottom (nothing fancy).
- Selectors in the new spec are only allowed in lists - the idea is that we would have a special "ConditionalList" class in the implementation that looks for the if / else and takes care of them. In the initial version even nested `if`'s would probably be forbidden. The goal was that we could make it more type-safe this way.
- Regarding types in expressions ... I've looked around in MiniJinja and it seems fairly hard to make it type-safe or "typed". Maybe we could do it. I am not sure I want it as a requirement at this point as it might hold up the implementation for too long. Maybe in the future we should actually think about a proper language (like Nickel or so).
- How about we move from `cmp` to a `cfg` function? That might be similar to how Rust / Cargo handles additional target dependent configuration like `cfg(python = ">3.10")` could be the syntax that we use. In Rust you can use additional operators like `cfg(all(python="3.10", numpy=">1.5"))`. Although I cannot promise that we can stick that into Jinja (minijinja)... The other way around (`all(cfg(python="3.10"), cfg(numpy=">1.5"))`) might be easier to implement.

I'll continue later!

cep-20.1.md
wolfv: @jakirkham in these cases you can infer those variables from the build-time config though (or variant config) so no recursion needed :)

cep-20.1.md
```diff
@@ -0,0 +1,286 @@
...
+
+```
+win and arm64
+(osx or linux) and aarch64
+something == "test"
+```
+
+### The cmp function for variant selection
+
+Furthermore, we have a special "cmp" function that can be used to run a check
```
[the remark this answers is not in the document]
wolfv: So we have "access" to jinja2 (e.g. to lowercase a string, extract a character from a string, join things etc.). This is about extending it with +1 function specific to conda matchspec comparison and variant-conditional builds (e.g. add this dependency only for python 3.8. I think it warrants a special function. I might however remove it from the proposal for now as it seems to be a point of contention ... and then we introduce a new proposal for this functioanlity.

cep-20.1.md
wolfv: Well other new features would also have to be implemented and I am not sure we would necessary end up with something simpler.
I think there is a future where the entire build tool looks a bit different (and the recipe as well) and wolfi-os / melange have a nice approach: https://github.com/wolfi-dev/os/blob/main/libidn2.yaml

But I am not sure how to proceed here - we could either go down that road _right now_ or build something that is more reasonable for people to port existing recipes to and then iterate in a next step on the recipe format (when we have the necessary infrastructure nicely implemented in rattler-build).

wolfv:
OK, continuation:

The point you raise about YAML and YAML being complex: totally agree. I don't think anybody uses (and I also don't understand) anchors and multi-documents so I am happy to explicitly forbid them. However, I do think there are places where integers are required (build number). Floating point numbers might indeed not appear in a recipe. But I do think that we can (should) enforce versions to be quoted as strings. With Rust, it's quite easy to enforce with `serde_yaml`, and our pydantic types also do it (https://github.com/prefix-dev/recipe-format it became even better with pydantic 2!).

As you alluded to, I am open to defer the final decision on the `cmp` function (or `cfg` or something else entirely). So I am considering to remove it from the spec for now in order to move past the initial spec. 

One last point that could be worthwhile to discuss is wether we want to take more inspiration from wolfi-os / melange, where packages look a bit more like Github Actions: https://github.com/wolfi-dev/os/blob/main/libidn2.yaml I really like it, but at the same time the hurdle to port existing recipes will be higher and the implementation is a bit different ... but in the future we could have a pipeline.yaml that does it that way :)

dholth:
I'm excited about the idea of a more parseable recipe format.

My first thought is that I'm not sure conda-build's special jinja2 functions are well-documented?

Too bad the existing format isn't versioned.

I'm more interested in parsing `info/recipe/meta.yaml` inside a `.conda`. The original yaml+jinja2 recipe becomes `info/recipe/meta.yaml.template`.

wolfv:
Based on feedback received, I have just removed mentions to the `cmp` function. We should have a seperate discussion about it.

I am ready to call for a vote on this CEP! :)

wolfv:
Not sure what to do but I just updated the dates and "changed" the status to proposed which means that the vote is open!

I mentioned it in the internal Gitter, but our implementation of the new format is pretty complete and works nicely in [`rattler-build`](https://github.com/prefix-dev/rattler-build). We are also in the process of updating its documentation to reflect the format changes we discussed here and that are now implemented.

cep-20.1.md
```diff
@@ -0,0 +1,325 @@
...
+  name: xtensor
+  version: 0.24.6
+  sha256: f87259b51aabafdd1183947747edfff4cff75d55375334f2e81cee6dc68ef655
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  fn: ${{ name }}-${{ version }}.tar.gz"
```
jaimergp, line 273:
```suggestion
  fn: ${{ name }}-${{ version }}.tar.gz
```

jaimergp (approved):
I'm excited to see this move forward. There might be more iterations on the content proposed in this CEP but I am inclined to consider it a good starting point. As long as we keep versioned schemas, you have my vote!

baszalmstra (approved):

jezdez:
@conda-incubator/steering

This PR falls under the **[Enhancement Proposal Approval](https://github.com/conda-incubator/governance/tree/03f21ae79dc1525eae0091ee3cf1e8f26856c6d6#enhancement-proposal-approval)** policy of the [conda governance policy](https://github.com/conda-incubator/governance), please vote and/or comment on this PR.

This PR **needs 60% of the Steering Council** to vote yea to pass.

To vote please leave **Approve (yea)** or **Request Changes (nay)** pull request reviews.

If you would like changes to the current language please leave a comment ([in the PR](https://github.com/conda/ceps/pull/54)) or push to this branch.

**This vote will end on 2023-10-27.**

baszalmstra:
yes

jaimergp:
Yes.

jezdez:
yes

ocefpaf:
yes

xhochy:
yes

wolfv:
yes

beckermr:
my vote doesn't count, but I am 100% supportive of this and very excited to see it moving ahead!

cep-20.1.md
chenghlee: +1000

kkraus14:
Yes

cep-20.1.md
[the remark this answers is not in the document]
dholth: Is there a way to preserve the `meta.yaml` filename and cause conda-build or rattler-build to detect the format from its contents?

goanpeca:
Yes

wolfv:
Wow, I am counting 11 yes votes (out of 15 council members) which appears to me as if the proposal is accepted!

jezdez:
@wolfv We should wait until the end of the voting period to tally all votes, FWIW

mbargull:
Yes.

.. But ideally with the main points from https://github.com/conda-incubator/ceps/pull/54#pullrequestreview-1515301216 addressed. I'll try to compose a changeset and issue a pull request against this one (if not done by tomorrow, it will have to wait for a subsequent CEP).

wolfv:
I guess we can _now_ celebrate that this has been accepted?! :) :tada:
I'll update and merge the CEP tomorrow with the appropriate number and status.

marcelotrevisani:
> I guess we can _now_ celebrate that this has been accepted?! :) 🎉 I'll update and merge the CEP tomorrow with the appropriate number and status.

Can I start to modify souschef and grayskull already? 😄

wolfv:
### Voting Results

This was a standard, non-timed-out vote.

Among [Steering Council members](https://github.com/conda-incubator/governance/blob/03f21ae79dc1525eae0091ee3cf1e8f26856c6d6/steering.csv) there are 12 "yes", no "no", and no abstentions.
This vote has reached quorum (at least 60% of 15).

It has passed since it recorded 12 "yes" votes and 0 "no" (80% yes out of 15 eligible voters).

dholth:
Does this diverge from the rattler book `- sel(osx): somepkg instead of - somepkg # [osx]`

baszalmstra:
> Does this diverge from the rattler book `- sel(osx): somepkg instead of - somepkg # [osx]` , `Jinja string interpolation needs to be quoted at the beginning of a string, e.g. - "{{ version }}"`

@dholth Yes it does, that is not correct. That should have been updated in the book, can you point me to where it says that?

dholth:
https://prefix-dev.github.io/rattler-build/recipe_file.html

jakirkham:
Oops meant to say "Yes!"

Thanks for working on this Wolf and everyone for reviewing! 🙏

wolfv:
@dholth we also publish the rust book here: https://prefix-dev.github.io/rattler-build/ where it should be updated. At the same time, we haven't made a release of rattler-build with the new parser.

dholth:
@wolfv That link has the same text on the recipe format page.

> Jinja string interpolation needs to be quoted at the beginning of a string, e.g. - "{{ version }}" in order for it to be valid YAML
> Selectors use a YAML dictionary style (vs. comments in conda-build). E.g. - sel(osx): somepkg instead of - somepkg # [osx]

And then these a little further down do not contain quotes?

> 
> package:
>   name: ${{ name }}
>   version: ${{ version }}

wolfv:
A PR can be made on: https://github.com/prefix-dev/rattler-build/tree/main/docs 
Otherwise we'll fix it asap :)

PR #55, opened 2023-06-08
Optional Dependencies for conda v2

travishathaway:
This is a second attempt to draft a CEP for optional dependencies for use by conda compatible package managers (e.g. `conda` and `mamba`).

Please see the CEP itself for more information.

### DIFF
--- a/cep-draft.md
+++ b/cep-draft.md
@@ -1,0 +1,125 @@
+<table>
+    <tr><td> Title </td><td> Optional Dependencies</td>
+    <tr><td> Status </td><td> Draft </td></tr>
+    <tr><td> Author(s) </td><td> Travis Hathaway &lt;travis.j.hathaway@gmail.com&gt;</td></tr>
+    <tr><td> Created </td><td> June 7, 2023</td></tr>
+    <tr><td> Updated </td><td> June 7, 2023</td></tr>
+    <tr><td> Discussion </td><td> NA </td></tr>
+    <tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+[python-extras]: https://peps.python.org/pep-0508/#extras
+[cargo-features]: https://doc.rust-lang.org/cargo/reference/features.html
+[original-cep]: https://github.com/conda-incubator/ceps/pull/9
+[conda-issue-7502]: https://github.com/conda/conda/issues/7502
+[conda-issue-11053]: https://github.com/conda/conda/issues/11053
+[mamba-issue-245]: https://github.com/mamba-org/boa/issues/245
+
+## Abstract
+
+This CEP proposes a new method for including optional dependencies when installing conda packages.
+Optional dependencies are largely inspired by [Python's "extras"][python-extras] and
+[Cargo's "features"][cargo-features]. When implemented, package builders will have a  way to specify
+dependencies which are not required but if installed would enable different features that are
+not included by default.
+
+## Motivation
+
+Optional dependencies as a feature has been requested for conda compatible package managers for quite
+some time (e.g. [conda-7502][conda-issue-7502], [conda-11053][conda-issue-11053], and [mamba-245][mamba-issue-245]).
+Adding this feature to the conda ecosystem would also give package managers greater flexibility when
+instructing users how to install and use their software. For example, a package could support multiple
+variants by using this methodology. If a package supported either `seaborn` or `matplotlib`,
+users could specify either `seaborn` or `matplotlib` as an optional dependency.
+
+
+## Specification
+
+This section will first show how this works from the end user point of view and then talks about
+how it will work for package builders. Updates to the `repodata.json` format will also be discussed.
+
+### User point of view
+
+When a user goes to install a package, they will specify optional dependencies with the following syntax:
+
+```bash
+# For a single optional dependency
+conda install package_name[optional_a]
+
+# For multiple optional dependencies
+conda install package_name[optional_a,optional_b]
+```
+
+Optional dependencies can also be specified by using the keyword argument syntax:
+
+```bash
+# For a single optional dependency
+conda install package_name[extras=optional_a]
+
+# For multiple optional dependencies
+conda install package_name[extras=optional_a,optional_b]
+```
+
+Later, when running commands like `conda list` the packages will show up as "virtual metapackages", which
+look like the following:
+
+```bash
+conda list
+
+# Name                    Version                   Build  Channel
+package_name              0.1.0                h0c17e10_0  conda-forge
+package_name@optional_a   0.1.0              pyhd8ed1ab_0  conda-forge
+package_name@optional_b   0.1.0              pyhd8ed1ab_0  conda-forge
+```
+
+When a user wants to later remove an optional dependency, they do so by referencing
+these "virtual metapackages":
+
+```bash
+conda remove package_name@optional_a
+```
+
+
+### Packager point of view
+
+**TBD**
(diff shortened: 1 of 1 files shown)


cep-draft.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+Optional dependencies can also be specified by using the keyword argument syntax:
+
+```bash
+# For a single optional dependency
+conda install package_name[extras=optional_a]
+
+# For multiple optional dependencies
+conda install package_name[extras=optional_a,optional_b]
```
h-vetinari, line 60: How does the parsing here work? It looks like it would break the `*args,**kwargs` pattern proposed in https://github.com/conda/conda/issues/11053 (which you're referencing already).

I'm also wondering if we really _need_  two ways to do this? `package_name[optional_a,optional_b]` looks perfectly alright to me, and matches with what pip is doing.

While it's out of scope for this CEP, I'd really like to keep the door open to eventually extend the syntax to be able to use (exposition only, all kwargs need a proper design)
```
conda install package_name[optional_a,optional_b,cuda=True,license=no_gpl,cpu=avx512,...]
```
It seems to me that if we add the `extras=` way to specify optional dependencies, we don't gain much and risk making it much harder to eventually solve that problem.

h-vetinari (commented):
This is amazing to see, thanks so much for picking this up!

Just one question/comment about the syntax.

I haven't yet thought through the implications of the virtual packages in detail, but it looks like a very elegant approach!

wolfv:
Can packages in the repodata also depend on other packages + optional dependencies?

cep-draft.md
nicornk: +1 for following the pip syntax.

cep-draft.md
```diff
@@ -0,0 +1,125 @@
...
+these "virtual metapackages":
+
+```bash
+conda remove package_name@optional_a
+```
+
+
+### Packager point of view
+
+**TBD**
```
jaimergp, line 85: Points to take into account here:

- How will this be specified in the conda-build recipes
- How will this be exposed in repodata.json
- How will this be reflected in the `info/` metadata folder inside the package

cep-draft.md
```diff
@@ -0,0 +1,125 @@
...
+package_name              0.1.0                h0c17e10_0  conda-forge
+package_name@optional_a   0.1.0              pyhd8ed1ab_0  conda-forge
+package_name@optional_b   0.1.0              pyhd8ed1ab_0  conda-forge
+```
+
+When a user wants to later remove an optional dependency, they do so by referencing
+these "virtual metapackages":
+
+```bash
+conda remove package_name@optional_a
```
jaimergp, line 79: I don't know how to feel about the new syntax for an arguably odd case. Why do we need it to begin with? If one wants to remove the extras, they could use the same spec à la `package_name[optional_a]`? And if you just want to remove the extras, they can simply spell them out.

jonas-w: Agreed, but I think there are good use cases for something like the “meta-packages” to exist:
1. Install package_name[optional_deps]
2. package_name gets an update and the extra “optional_deps” adds another dependency
3. You upgrade the _(automatically, with the other packages)_ package_name@optional_deps package and also get the new dependency that was added with the upgrade.

Otherwise, if there are no “virtual meta-packages” (or anything that keeps track of the extras you installed) the user wouldn't get the new dependency.

Also, another thing that I would love to have, but which would differ from the way pip does it (at least to my knowledge), is to not stop at optional dependencies, but go further with optional modules/files.
Let's say my project has the following structure:
```
src/package/...
src/package/core/...
src/package/feature_a/...
src/package/feature_b/...
```

And in the meta.yaml I could configure it like this (just an example):
```yaml
# if files get specifically included in an extra,
# they would be excluded from the "core" package
extras:
    feature_a:
        include_files: ["feature_a/"]
        dependencies: ["optional1",..]
    feature_b:
        include_files: ["feature_b/"]
        dependencies: ["optional2",...]
# maybe also something like feature_c
    feature_c:
        include_extras: ["feature_a","feature_b"]
# optionally also include further files/dependencies etc.
```

Then `conda install package`, would only include /core.
And `conda install package[feature_a]` would include /core and /feature_a and feature_a's dependencies.
And `conda install package[feature_c]` would include everything.

mbargull:
This CEP was mentioned by @chenghlee in a discussion at https://matrix.to/#/!ugIkjExjZULQyVDDNz:gitter.im/$vMqrBnqcqeOfqNq27A4epLJ_xDuIjXiu-59XN_KeBbk?via=gitter.im&via=matrix.org

I didn't yet read the text or discussion much, but just wanted to copy my points from the chat:
<details>

> There are multiple steps; main points (list not necessarily exhaustive):
> 
> 1. Define the actual semantics on what "optional"/"extra" would mean (ask different people and you'd get different answers...).
> I.e., figure out how this fits into the general dependency resolution and what defaults should be (and look out for typical corner cases, e.g., differences between env `create`/`install`/`update`.)
> 2. Define the data storage/communication layer, i.e., "repodata CEP update" and stuff.
> 3. Define the user interface for this; e.g., (amongst other things) how would a user say "install foo with bar".
>     - This cannot be `install foo[bar]` since we have `[...]` already reserved for "match spec".
>     - Would have to be `foo[extra=bar]`/`foo[extras=[bar]]` or the like if this would use he match spec syntax; could be something entirely different, too, though.
</details>

xhochy:
I don't understand why we need this feature. We already use in several cases multi-output recipes for optional dependencies where a `package[feature]` on the `pip` side is translated to a `package-feature` package. For the sake of simplicity, I would prefer to avoid to define an additional way of doing it and keep the current way of doing it.

h-vetinari:
> We already use in several cases multi-output recipes for optional dependencies where a `package[feature]` on the `pip` side is translated to a `package-feature` package.

Yes, but it's a crutch, and a persistent paper cut that people run into and contributes substantially to friction and user-frustration for requirement-incompatibilities between PyPI and conda. Even a bunch of conda-forge [recipes](https://github.com/search?q=org%3Aconda-forge%20%2F-%20%5B%5Cw%5C-%5D%2B%5C%5B%2F%20path%3Arecipe%2Fmeta.yaml&type=code) get this wrong (and don't even get a warning that this doesn't do what's expected!). It should IMO absolutely be fixed just based on those two points alone.

If we zoom out a little bit further, this "string-based work arounds instead of sufficiently expressive syntax/metadata" is a constant source of pain (the other huge pain point on this being the completely [overloaded](https://github.com/conda/conda/issues/11053) build strings), which I mention here because IMO both should be solved with the same syntax.

```
conda install package[feature,build_number=1,cuda=True]
```
would be orders of magnitude more user-friendly than the following jumble of symbols
```
conda install package-feature=*=cuda*_1
```
and that's even before getting into https://github.com/conda/conda/issues/11442, https://github.com/conda/conda-build/issues/4498, https://github.com/conda/conda-libmamba-solver/issues/446, etc.

xhochy:
My main concern is that I want to avoid is adding an attribute to repodata.json as this will take longer to propagate to all implementations. 

It would be nice if the problem could be solved by mapping the square bracket syntax to plain packages, i.e. make this a client-side convention. 

One example could be to map `jaxlib[build_number=1,cuda,docs]` to `jaxlib=*=*_1 jaxlib-cuda jaxlib-docs`.

wolfv:
Should we close this @travishathaway as we are starting to vote on a competing proposal?

PR #57, opened 2023-06-15
A new test for file existence inside a package (for the new recipe format)

wolfv:


cep-20.3.md
```diff
@@ -0,0 +1,83 @@
...
+
+## The specification
+
+We propose a new "test" section (following the split of testing sections in this [CEP](https://github.com/conda-incubator/ceps/pull/56)) to test for package file existence.
+The test are executed on the raw package contents (without installing them). That should make them very fast to execute, as well as less brittle vs. cmd.exe / bash scripts.
+
+```yaml
+package_contents:
+  # checks for the existence of files inside $PREFIX or %PREFIX%
+  files:
```
jaimergp, line 53:
```suggestion
  prefix:
```

Given that the rest of the categories use the "base location" as key name, this might make more sense.

cep-20.3.md
```diff
@@ -0,0 +1,83 @@
...
+    - etc/libmamba/test.txt
+    - etc/libmamba
+    - etc/libmamba/*.mamba.txt # can also use globs, then _at least_ one file needs to match
+
+  # checks for the existence of `mamba/api/__init__.py` inside of the Python site-packages directory
+  # NOTE: the user can/should also use Python import checks
+  site_packages:
+    - mamba.api
```
jaimergp, line 61: I like that we can have direct access to `site_packages`, but I am not sure about the Python import path syntax. How can I make sure I have a data file in my Python package, for example? I'd say we stick to normal paths here, so this would be written like this:

```suggestion
    - mamba/api/__init__.py
```

After all, we have the `imports:` tests anyway.

cep-20.3.md
```diff
@@ -0,0 +1,83 @@
...
+
+  # looks in $PREFIX/bin/mamba for unix and %PREFIX%\Library\bin\mamba.exe on Windows
+  # NOTE: the user can/should also check the `commands` and execute something like `mamba --help` to make
+  #       sure things work fine
+  bin:
+    - mamba
+
+  # searches for `$PREFIX/lib/libmamba.so` or `$PREFIX/lib/libmamba.dylib` on Linux or macOS,
+  # on Windows for %PREFIX%\Library\lib\mamba.dll & %PREFIX%\Library\bin\mamba.bin
```
jaimergp, line 70:
```suggestion
  # on Windows for %PREFIX%\Library\lib\mamba.dll & %PREFIX%\Library\bin\mamba.lib
```

.lib?

cep-20.3.md
```diff
@@ -0,0 +1,83 @@
...
+  lib:
```
jaimergp, line 71: What about versioned libraries? The numbered suffixes are placed in different spots of the filename in linux and macos.

cep-20.3.md
wolfv: Yeah, it's a good point.

wolfv:
Are there other special locations, e.g. for R packages?

dhirschfeld:
> I am also wondering how we can test that some files were _not_ added. It is common to have `tests/` top-level packages accidentally packaged, and we could have something like:
> 
> ```yaml
> package_contents:
>   files:
>      - ~tests  # or "!tests"
> ```

Maybe:
```yaml
package_contents:
  includes:
    - mamba/api/
  excludes:
     - mamba/tests/
```

wolfv:
@dhirschfeld that is also nice, but then it becomes harder to negate the existence of a given shared library, include file, etc. I lean towards `~` as prefix for negation.

JeanChristopheMorinPerso (commented):
I think it makes sense. Though, I'm wondering if it would be a god occasion to introduce a strict mode where the recipe has to define strictly which files should be included in the package similar to how it's done in the RPM land (https://src.fedoraproject.org/rpms/yaml-cpp/blob/rawhide/f/yaml-cpp.spec#_72). `package_contents` really makes me think of that.

Though, I'm not sure if it's really related to your PR or if we should have this discussion somewhere else.

cep-20.3.md
```diff
@@ -0,0 +1,83 @@
...
+    - test -f ${PREFIX}/share/cmake/xtensor/xtensorConfigVersion.cmake # [unix]
+    - if not exist %LIBRARY_PREFIX%\include\xtensor\xarray.hpp (exit 1) # [win]
+    - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfig.cmake (exit 1) # [win]
+    - if not exist %LIBRARY_PREFIX%\share\cmake\xtensor\xtensorConfigVersion.cmake (exit 1) # [win]
+```
+
+## The specification
+
+We propose a new "test" section (following the split of testing sections in this [CEP](https://github.com/conda-incubator/ceps/pull/56)) to test for package file existence.
+The test are executed on the raw package contents (without installing them). That should make them very fast to execute, as well as less brittle vs. cmd.exe / bash scripts.
```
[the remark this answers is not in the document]
wolfv: It means that the tests only require the `paths.json` file. The package does not need to be extracted or linked into an environment.

wolfv:
@JeanChristopheMorinPerso the RPM spec that you linked also has this:

```
%{_includedir}/yaml-cpp/
```

So it seems that it similarly has a "glob" syntax to take an entire directory. But thanks for the pointer, I didn't know about that!

wolfv:
Some things that came up yesterday in the mamba meeting:

- differentiating between shared and static libraries is currently not possible – should we add another field (`static_libs`) to find the existence (or negate the existence!) of static libraries in the package?
- it came up that the comment about Windows dlls should be placed in `bin` and the comment may be wrong
- for windows dlls we should do a strong test (e.g. make sure that both `dll` and `lib` file exist as a pair

@JeanChristopheMorinPerso now that I think about it it could be a great idea to have a `strict_mode: true / false` that means that all files in the package have to be matched by any of the rules.

wolfv:
Maybe `strict_mode` should be `true` by default ... :)

h-vetinari:
`strict_mode` sounds cool!

> for windows dlls we should do a strong test (e.g. make sure that both `dll` and `lib` file exist as a pair

In an ideal world, this would also check that the `.lib` file is actually an import library, and not a static library with the same name. 😇

wolfv:
@h-vetinari I just chatted with ChatGPT and looks like we can do that with Goblin and some inspection of the files.

JeanChristopheMorinPerso:
> @JeanChristopheMorinPerso the RPM spec that you linked also has this:
> 
> ```
> %{_includedir}/yaml-cpp/
> ```
> 
> So it seems that it similarly has a "glob" syntax to take an entire directory. But thanks for the pointer, I didn't know about that!

The support a glob-style match, but I don't know if `%{_includedir}/yaml-cpp/` necessarily does a recursive include or not.

jezdez:
Closing in favor of #84.

PR #63, opened 2023-11-23
Add purls (Package URLs) to `PackageRecord`

baszalmstra:
This CEP describes a change to the `PackageRecord` format and the corresponding `repodata.json` file to include `purls` ([Package URLs]((https://github.com/package-url/purl-spec/)) of repackaged packages to identify packages across multiple ecosystems.

[rendered](https://github.com/baszalmstra/ceps/blob/cep-purls/cep-purls.md)


wolfv:
Awesome CEP! :)

cep-purls.md
```diff
@@ -0,0 +1,100 @@
...
+<tr><td> Author(s) </td><td> Bas Zalmstra &lt;bas@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> Nov 23, 2023</td></tr>
+<tr><td> Updated </td><td> Nov 23, 2023</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CEP describes a change to the `PackageRecord` format and the corresponding `repodata.json` file to include `purls` (Package Urls) of repackaged packages to identify packages across multiple ecosystems.
```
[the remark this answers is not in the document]
baszalmstra: Unfortunately, I believe that atm there is no actual "authorative" source. 

There is this relatively old definition of a `RepoDataRecord`: https://github.com/conda/schemas/blob/main/repodata-record-1.schema.json

There is this new effort to document the schemas better (https://github.com/conda/schemas/pull/26) where it's also called `RepoDataRecord`: https://github.com/conda/schemas/blob/b143c82a71833570fbe9be2313368b33c0e84726/conda_models/package_record.py#L23

And we have the definition in rattler: https://docs.rs/rattler_conda_types/latest/rattler_conda_types/struct.PackageRecord.html

In rattler (and I believe in conda as well), there is this distinction:

* `PackageRecord`: contains all the fields for a single entry in the `repodata.json`
* `RepoDataRecord`: inherits all fields from `PackageRecord` and adds fields to identify the origin of the data (channel, url, etc.)
* `PrefixRecord`: inhertis all fields from `RepoDataRecord` and additionally stores information about how the package was installed.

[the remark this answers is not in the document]
jaimergp: Yea, I think the most "official" source for this is https://github.com/conda/conda/blob/e783377439ed1c413c6bffb9b785ae1d79c2392a/conda/models/records.py#L247. That module also offers some sort of definition in the top-level docstring.

cep-purls.md
```diff
@@ -0,0 +1,100 @@
...
+    "name": "pinject",
+    "version": "0.14.1",
+    "purls": ["pkg:pypi/pinject@0.14.1"],
+    ...
+  }
+  ...
+}
+```
+
+PURL is already supported by dependency-related tooling like SPDX (see [External Repository Identifiers in the SPDX 2.3 spec](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f35-purl)), the [Open Source Vulnerability format](https://ossf.github.io/osv-schema/#affectedpackage-field), and the [Sonatype OSS Index](https://ossindex.sonatype.org/doc/coordinates); not having to wait years before support in such tooling arrives is valuable.
```
jaimergp, line 33: I would also mention [PEP-725](https://peps.python.org/pep-0725/) (WIP).

cep-purls.md
```diff
@@ -0,0 +1,100 @@
...
+Using the PURL standard allows us to link vulnerabilities from other ecosystems to conda package.
+
+## Rationale
+
+Adding the information to the `repodata.json` file has some advantages:
+
+* We can keep this information close to the conda package description.
+* We can incrementally add `purls` through repodata patches.
+
+The downside is that the (already large) repodata.json file will grow.
```
jaimergp, line 53: What if we add a separate-yet-adjacent `purls.json` like we did with `run_exports.json` in CEP-12?

jaimergp (changes requested):
I like the idea and I will be supportive. Havin this metadata readily available would allow us to be listed in repology.org, for example! It would also play nicely with the (draft) PEP-725 for external metadata in PyPI.

However, I think this CEP right now is talking about _serving_ metadata before we have discussed how to source it, define it and store it.

Whatever ends up in the repodata.json comes, in part, from the `info/index.json` metadata _inside_ the conda artifact. Then this is augmented with things like sha256 and final size by conda-index (because they cannot be known when the package is being archived).

So before we speak about repodata, we should discuss where in the inner artifact metadata we will store the PURL info. To answer that, we must answer where in the conda-build recipe we will include that information :D 

IOW, I'd like to know your thoughts about:

- Where in the current `meta.yaml` we should define the PURLs. `about` seems to be the most obvious one, which means this will probably end up in `info/about.json`.
- Whether to serve the PURLs separately in a `purls.json` or not. I honestly don't think putting it in `repodata.json` is a good idea. I get that it makes sense if you want to have a canonical link between PyPI in conda-forge so Pixi can solve things nicely. It might also be served in `channeldata.json` (since most of the time PURLs are tied to the _source_ not the platform-dependent, target artifact).

jakirkham:
Would this also help us address Repology's needs for supporting Conda packages ( https://github.com/repology/repology-updater/issues/518 )?

ytausch:
> Where in the current meta.yaml we should define the PURLs. about seems to be the most obvious one, which means this will probably end up in info/about.json.

I agree that `about` makes the most sense. However, this adds the redundancy of defining the upstream package twice in the recipe. A more sophisticated solution would be adding a new `purl` source type for the source section, which gets resolved to a PyPI tarball URL by conda build. The purls for a package could then be automatically inferred from the sources it has been built from. In all cases, a manual option to define the purls likely has to remain for some specific use cases.

While this would facilitate simplicity, avoid redundancy, and avoid errors in the recipe, I see the following downsides with that solution:
- different package outputs or variants may not actually use all sources that are available, requiring manually overriding the purls or another clever solution for that
- how to verify the hash of a source tarball is more evident if the source URL is stated explicitly in the recipe
- to avoid complexity, we could only support a subset of [purl types](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst). PyPI is by far the most important IMO. It could confuse people if only a subset of purls are valid sources.
- introducing a new source type is simply more work than introducing a new `about` field - especially in related tooling such as cf-scripts or conda-smithy
- backward compatibility?

> Whether to serve the PURLs separately in a purls.json or not. I honestly don't think putting it in repodata.json is a good idea. I get that it makes sense if you want to have a canonical link between PyPI in conda-forge so Pixi can solve things nicely. It might also be served in channeldata.json (since most of the time PURLs are tied to the source not the platform-dependent, target artifact).

I do not have a strong opinion here since I am not too involved with the tools that would need to process that data.

bollwyvl:
I think a broader question is whether `package-url`can be adopted more directly _by_ the ecosystem. Relying on "the URL where you get your package" or "what it's called on disk" aren't as effective as an agreed-upon grammar for identifying packages, especially in the "is this CVE relevant to me" scenario.

> Brief aside, and likely worth including in the text
>> A purl is a URL composed of seven components:
>> ```
>>  scheme:type/namespace/name@version?qualifiers#subpath
>> ```

To put this in the context of the above, a given `.conda` package might claim a PURL as a proxy in one or more other _types_, but by existing, it should claim one in the `conda` _type_.  Indeed, a subset is already [part of the spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#conda). For example, an [old version of `django`](https://anaconda.org/conda-forge/django/files/modal/info/57cb27b8a05b447ad1d69f59):

It might make sense to advocate for some changes to the `conda` part of the spec (and [test data](https://github.com/package-url/purl-spec/blob/c02b00/test-suite-data.json#L351)), namely:
- update the default `channel` from `r.a.com -> c.a.org`
  - then the `namespace` part of the url would encode the conda channel (e.g. `pkg:conda/conda-forge/django`)
  - i don't know what the "new `null`" would be, but pretty sure it can't be `defaults`
- use `label` for e.g. `main` (not `channel`, as in the example)

While i don't think much can be done about "where you got the source tarball" (because GitHub sources, etc), I _don't_ think a recipe author should have to calculate all these things... but certainly could given the available data today:

```yaml
# meta.yaml
{% set version = "1.10.1" %}
package: 
  name: django
  version: {{ version }}
# ...
about:
  # ...
  purls:
    - pkg:pypi/django@{{ version }}
    # this should be fully automated, either at build time (weird?) or trivially-derivable
    - pkg:conda/{{ channel_targets.split(" ")[0] }}/django@1.10.1?subdir={{ target_platform }}&label={{ channel_targets.split(" ")[1] }}&build=py{{ py }}_{{ build_number }}
```

So the above full `purls` might expand to

```yaml
purls:
- pkg:pypi/django@1.10.1
- pkg:conda/conda-forge/django@1.10.1?subdir=win-32&label=main&build=py35_0
```

bollwyvl:
Thinking about this more in the context of "accidental cross-ecosystem namesquatting" on [zulip](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.E2.9C.94.20Name.20conflict.20with.20inactive.20package): as `pkg:` isn't presently a valid package name, the `MatchSpec` grammar _could_ be expanded to include `purl`s as an alternative package identifier

```yaml
dependencies:
- pkg:pypi/django >=1.10.1,<1.11
```

treating everything after the whitespace as "this part is about conda" would still allow for all our variant business, but presumably could _eventually_ be expanded to allow per-ecosystem fields... luckily, pypi _only_ has semi-irrelevant stuff like `file_name`, but other ecosystems could be more complex.

ytausch:
> include purls as an alternative package identifier

I don't follow entirely. What would your example refer to? The PyPI package or the corresponding conda-forge package?

bollwyvl:
>> include purls as an alternative package identifier
>
> I don't follow entirely. What would your example refer to? The PyPI package or the corresponding conda-forge package?

Right, the user wants the corresponding  `conda` package, from the highest-priority channel, but resolved in the  parallel namespace.

```toml
# e.g. in pixi.toml
[dependencies]
# |  a new package identifier
# V
"pkg:pypi/django" = ">=1.10.1,<1.11"
#                    ^      
#                    |  the conda constraints, in the MatchSpec grammar
"pkg:golang/github.com/rhysd/actionlint" = ">=1.7.7"
```

Where this would be _most_ excellent, for the PyPI case, is if the [spec](https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#pypi)
could also capture `[extras]`, which are seeing increasing usage in Python inter-package
dependencies (even though `pip` is bad at them). 

There is no consensus in `conda` packaging on how to capture such "optional dependencies," 
and some packages just ship _all_ the optional dependencies (this hurts on Big 
Cloud Vendor API extras), or have multiple outputs, but again without any naming 
conventions (e.g. some folk just  `-{extra-name}`, I generally push for 
`-with-{extra-name}`).

An extreme case might be `fastapi`:

```yaml
# e.g. in rattler-build recipe.yaml
recipe:
  version: ${{ version }}
outputs:
  # with fully-specified purls
  - package:
      name: fastapi
      purl: pkg:pypi/fastapi@${{ version }}
    dependencies:
      run:
        - pkg:pypi/starlette >=0.40.0,<0.42.0
        - pkg:pypi/pydantic >=1.7.4,!=1.8,!=1.8.1,!=2.0.0,!=2.0.1,!=2.1.0,<3.0.0
        - pkg:pypi/typing-extensions >=4.8.0
  # or maybe it makes sense to CURIE them, using a `pip:`-like syntax
  - package:
      name: fastapi-standard
      purl:
        pkg:pypi:
          - fastapi[standard]@${{ version }}
    dependencies:
      run:
        - ${{ pin_subpackage("fastapi", exact=True) }}
        - pkg:pypi:
          - fastapi-cli[standard] >=0.0.5
          - httpx >=0.23.0
          - jinja2 >=2.11.2
          - python-multipart >=0.0.7
          - itsdangerous >=1.1.0
          - pyyaml >=5.3.1
          - ujson >=4.0.1,!=4.0.2,!=4.1.0,!=4.2.0,!=4.3.0,!=5.0.0,!=5.1.0
          - orjson >=3.2.1
          - email-validator >=2.0.0
          - uvicorn[standard] >=0.12.0
          - pydantic-settings >=2.0.0
          - pydantic-extra-types >=2.0.0
```

The latter form would all but remove any package-naming impedance, making tools
like `grayskull` have to do much less work to maintain package resolution stuff.

cep-purls.md
baszalmstra: I think we should just include them with sharded repodata but for backwards compatibility it seems logic to go the same route as run exports as well (also allowing patching them).

cep-purls.md
```diff
@@ -0,0 +1,100 @@
...
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+This CEP describes a change to the `PackageRecord` format and the corresponding `repodata.json` file to include `purls` (Package Urls) of repackaged packages to identify packages across multiple ecosystems.
+
+## Specification
+
+We propose to add the optional `purls: [string]` field to `PackageRecord`.
```
pavelzw, line 17: As @jaimergp mentioned, we should also mention in this CEP where in the built conda package this should live and where it is specifiable in the recipe.
IMO, `index.json` is a good fit for it as IMO, it should also go into repodata (at least into sharded).

Adding it to `index.json` will lead to `conda-index` automatically adding it to `repodata.json`. This is IMO a bad default in `conda-index`; it should instead just keep a whitelist of things to put into `repodata.json` instead of putting everything from `index.json` in there.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+Tools that generate packages like rattler-build and conda-build should also be able to inject a PURL at build time via CLI flags.
+For this, v1-recipes can use the jinja syntax from the recipes.
+
+```bash
+rattler-build build -r recipe/ \
+    --append-purl-pattern \
+    'pkg:conda/conda-forge/${{ PACKAGE_NAME }}@${{ PACKAGE_VERSION }}?build=${{ BUILD_NUMBER }}'
+```
+
+Variables that are available in the build process (like `PACKAGE_NAME`, `PACKAGE_VERSION` and `BUILD_NUMBER`) can be specified here.
```
pavelzw, line 125: Not sure about this one yet. We need some way of adding the PURL to conda-forge packages.

An alternative would be to enforce with conda-smithy that something like this is always in the recipe:

```yml
about:
  purls:
    - pkg:pypi/pinject@0.14.1
    - pkg:conda/conda-forge/${{ name }}@${{ version }}?build=${{ build }}
```

this would be a bit more explicit. But we would not be easily able to add things like `&platform=osx-arm64`

Passing it through the infrastructure via CLI flags makes sure it's correct and would immediately apply to each new build though.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+  }
+}
+```
+
+### PURL of a conda packages
+
+Conda packages itself should have a PURL as well.
+This makes it possible for CVE Numbering Autorities (CNA) to publish vulnerabilities for conda packages.
+A package `pinject-0.14.1-pyh29332c3_1.conda` published on `conda-forge` should have the PURL `pkg:conda/conda-forge/pinject@0.14.1?build=pyh29332c3_1&platform=noarch`.
+Packages can also specify a custom `repository_url`, for example `pkg:conda/custom-channel/my-package@0.1.0?build=pyh29332c3_1&platform=noarch&repository_url=prefix.dev`.
```
JeanChristopheMorinPerso, line 114: If I'm not mistaken, `repository_url` should be an actual URL, not a domain name: https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst#known-qualifiers-keyvalue-pairs

>repository_url is an extra URL for an alternative, non-default package repository or registry. When a package does not come from the default public package repository for its type a purl may be qualified with this extra URL. The default repository or registry of a type is documented in the "Known purl types" section.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+    ...
+  }
+}
+```
+
+### PURL of a conda packages
+
+Conda packages itself should have a PURL as well.
+This makes it possible for CVE Numbering Autorities (CNA) to publish vulnerabilities for conda packages.
+A package `pinject-0.14.1-pyh29332c3_1.conda` published on `conda-forge` should have the PURL `pkg:conda/conda-forge/pinject@0.14.1?build=pyh29332c3_1&platform=noarch`.
```
JeanChristopheMorinPerso, line 113: Should we worry that this diverges from https://github.com/package-url/purl-spec/blob/main/PURL-TYPES.rst#conda? This seems to be quite problematic IMO. In the spec, channel is a specifier (`pkg:conda/pinject@0.14.1?channel=conda-forge`).

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+    'pkg:conda/conda-forge/${{ PACKAGE_NAME }}@${{ PACKAGE_VERSION }}?build=${{ BUILD_NUMBER }}'
+```
+
+Variables that are available in the build process (like `PACKAGE_NAME`, `PACKAGE_VERSION` and `BUILD_NUMBER`) can be specified here.
+
+## Motivation
+
+Conda packages can mostly repackage packages from other ecosystems.
+Conda-forge and other channels famously repackages a lot of PyPI packages.
+However, without actually downloading the conda package and inspecting its contents there is no reliable way to know whether a certain conda package is a repackaged package.
```
JeanChristopheMorinPerso, line 131:
```suggestion
However, without actually downloading the conda package and inspecting its contents there is no reliable way to know whether a certain conda package is a repackaged package and which package it repackages.
```
?

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+  ...
+}
+```
+
+PURL is already supported by dependency-related tooling like SPDX (see [External Repository Identifiers in the SPDX 2.3 spec](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f35-purl)), the [Open Source Vulnerability format](https://ossf.github.io/osv-schema/#affectedpackage-field), and the [Sonatype OSS Index](https://ossindex.sonatype.org/doc/coordinates); not having to wait years before support in such tooling arrives is valuable.
+[PEP 725 (WIP)](https://peps.python.org/pep-0725) also proposes how to specify non-PyPi dependencies using PURLs.
+
+### PURLs in recipes
+
+PURLs can be specified in the recipe under `.about.purls` (implementation for [v1-recipes](cep-00014.md): [prefix-dev/rattler-build #1664](https://github.com/prefix-dev/rattler-build/pull/1664)):
```
jaimergp, line 38: Let's use RFC language here:

```suggestion
PURLs MAY be specified in the recipe under `.about.purls` (implementation for [v1-recipes](cep-00014.md): [prefix-dev/rattler-build #1664](https://github.com/prefix-dev/rattler-build/pull/1664)):
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+```yaml
+about:
+  # ...
+  # PURLs that this package belongs to
+  purls: [string (PURL enforced)]
```
jaimergp, line 44:
```suggestion
  purls: [string (PURL enforced)]
```

What do you mean by "PURL enforced"? Should we add some language like:

> The strings in the `purls` field MUST be valid PURL identifiers.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+```json
+{
+  ...,
+  "purls": ["pkg:pypi/pinject@0.14.1"]
+}
+```
+
+### PURLs in repodata
+
+In `repodata.json`, PURLs should be an optional field.
```
jaimergp, line 58:
```suggestion
In `repodata.json`, PURLs SHOULD be an optional field.
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+Tools that generate `repodata.json` (like `rattler-index` or `conda-index`) can decide whether to include it or not in `repodata.json`.
```
jaimergp, line 59:
```suggestion
Tools that generate `repodata.json` (like `rattler-index` or `conda-index`) MAY decide whether to include it or not in `repodata.json` and derivatives.
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+Instead of writing the PURLs in `repodata.json`, these tools can decide to add PURLs to a separate `purls.json` like in [CEP 12](cep-0012.md):
```
jaimergp, line 61:
```suggestion
Instead of writing the PURLs in `repodata.json`, these tools MAY add PURLs to a separate `purls.json` file, like in [CEP 12](cep-0012.md), with the following schema:
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+  "packages.conda": {
+    "package-version-build.conda": {
+      ...,
+      "purls": ["pkg:pypi/pinject@0.14.1"]
+    }
+  }
+}
+```
+
+If the `purls` field is not present in the `repodata.json` record, it means no `purls` information is stored with the record. Then, a fallback mechanism (i.e., read `purls.json`) should be used to acquire the PURL information.
```
jaimergp, line 86:
```suggestion
If the `purls` field is not present in the `repodata.json` record, it means no `purls` information is stored with the record. Then, a fallback mechanism (i.e., read `purls.json`) SHOULD be used to acquire the PURL information.
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+In sharded repodata ([CEP 16](cep-0016.md)), PURLs can be specified similar to [CEP 21](cep-0021.md).
```
jaimergp, line 88: Can you elaborate here with an example?

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+### Patching
+
+With the purls part of the repodata we propose to also allow repodata patching this field. The original purls can still be extracted from the packages in `index.json`. We do not see a reason how the PURL information is different from other patchable information stored in the repodata (like the dependencies).
```
jaimergp, line 92:
```suggestion
With the purls part of the repodata we propose to also allow repodata patching this field. The original purls can still be extracted from the packages in `index.json`. We do not see a reason how the PURL information is different from other patchable information stored in the repodata (like the dependencies).
```

I'm not sure we want to dynamically alter a field that might end up being used to generate SBOMs and other types of audit-related data. I need to think about this.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+    "package-version-build.tar.bz2": {
+      ...,
+      "purls": ["pkg:pypi/pinject@0.14.1"]
+    },
+    ...
+  }
+}
+```
+
+### PURL of a conda packages
```
jaimergp, line 109:
```suggestion
### PURL of a conda package
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+Conda packages itself should have a PURL as well.
+This makes it possible for CVE Numbering Autorities (CNA) to publish vulnerabilities for conda packages.
```
jaimergp, line 112:
```suggestion
Conda packages itself MAY have a PURL as well. This makes it possible for CVE Numbering Autorities (CNA) to publish vulnerabilities for conda packages. This section defines the conventions followed to generate such an identifier.
```

cep-purls.md
jaimergp: Yes, for now we need to stick to the PURL spec (even if I'm not a big fan of how they are currently defining the `conda` type). Changes like the ones proposed here need to happen there first, if needed.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+A package `pinject-0.14.1-pyh29332c3_1.conda` published on `conda-forge` should have the PURL `pkg:conda/conda-forge/pinject@0.14.1?build=pyh29332c3_1&platform=noarch`.
+Packages can also specify a custom `repository_url`, for example `pkg:conda/custom-channel/my-package@0.1.0?build=pyh29332c3_1&platform=noarch&repository_url=prefix.dev`.
+
+Tools that generate packages like rattler-build and conda-build should also be able to inject a PURL at build time via CLI flags.
```
jaimergp, line 116:
```suggestion
Tools that generate packages like rattler-build and conda-build SHOULD provide an option to inject a PURL at build time via CLI flags.
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+For this, v1-recipes can use the jinja syntax from the recipes.
```
jaimergp, line 117:
```suggestion
For this, v1-recipes MAY use the Jinja syntax from the recipes.
```

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+
+```bash
+rattler-build build -r recipe/ \
+    --append-purl-pattern \
+    'pkg:conda/conda-forge/${{ PACKAGE_NAME }}@${{ PACKAGE_VERSION }}?build=${{ BUILD_NUMBER }}'
```
jaimergp, line 122: I don't see why we need to inject the PURL of the package being built into itself. All this information can be derived from the filename if we want to publish it somewhere else, but I don't see how publishing an easily buildable string adds value.

Also, the `build` qualifier is the build _string_ not the number.

cep-purls.md
jaimergp, line 109: I think this section deserves its own CEP.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+## Motivation
+
+Conda packages can mostly repackage packages from other ecosystems.
+Conda-forge and other channels famously repackages a lot of PyPI packages.
+However, without actually downloading the conda package and inspecting its contents there is no reliable way to know whether a certain conda package is a repackaged package and which package it repackages.
+
+Tools like pixi or conda-lock try to combine conda and PyPI packages through heuristics. This doesn't work deterministically as package names between the two indices may differ.
+
+Its hard to use open-source vulnerability databases because they often do not contain conda packages.
+Using the PURL standard allows us to link vulnerabilities from other ecosystems to conda package.
```
jaimergp, line 136: I would move this above Specification so it's clear _why_ we are doing this. It would also need some more details about the intended usage and benefits, since there are many!

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+In `repodata.json`, PURLs should be an optional field.
+Tools that generate `repodata.json` (like `rattler-index` or `conda-index`) can decide whether to include it or not in `repodata.json`.
+
+Instead of writing the PURLs in `repodata.json`, these tools can decide to add PURLs to a separate `purls.json` like in [CEP 12](cep-0012.md):
+
+```json
+{
+  "info": {
+    "platform": "string",
+    "arch": "string",
```
jaimergp, line 67: I think these are superfluous here.

cep-purls.md
```diff
@@ -0,0 +1,192 @@
...
+    ...
+  }
+  ...
+}
+```
+
+PURL is already supported by dependency-related tooling like SPDX (see [External Repository Identifiers in the SPDX 2.3 spec](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f35-purl)), the [Open Source Vulnerability format](https://ossf.github.io/osv-schema/#affectedpackage-field), and the [Sonatype OSS Index](https://ossindex.sonatype.org/doc/coordinates); not having to wait years before support in such tooling arrives is valuable.
+[PEP 725 (WIP)](https://peps.python.org/pep-0725) also proposes how to specify non-PyPi dependencies using PURLs.
+
+### PURLs in recipes
```
jaimergp, line 36: I'd like to read recommendations on _which_ PURLs should be included, e.g.:

- Sources being built, seems obvious
- Vendored dependencies?
- Statically compiled libraries?

jaimergp (changes requested):
Happy to see more activity here! Some comments and clarifications.

I also wonder about the following. I think we are conflating two items:

1. Allowing a recipe to specify which packages are included in a given conda artifact
2. Standardizing the conventions to assign a PURL to each conda artifact

It appears that both would end up in the `purls` field, but that doesn't sound right to me. (1) is about enumerating components in package, and (2) is about identifying a conda artifact. Mixing these two doesn't seem like a good idea.

I propose we split (2) from this CEP, and create its own mini-CEP, which in a way it should simply recognize the PURL spec and how to use it in the conda ecosystem. After all it just maps `pkg:conda/numpy?channel=conda-forge` to `conda-forge::numpy`.

---

Also please see https://github.com/conda/ceps/pull/114 for some related discussions, I think part of the text there has some merit and I'd be happy to merge that PR here to centralize the conversation.

cep-purls.md
jaimergp: I asked the PURL team today about how to submit changes to their spec, and right now they are focused on some large rearrangements of the specification (a centralized schema for type definitions and tests), so it's bad timing to submit type-specific (e.g `conda`) changes. But later this year this work will be done so type-specific feedback will be welcome by then. I'll stay in touch with them to monitor the situation and report once we can submit some actionable items. In the meantime, we can brainstorm as a community what we'd like to propose to the standard.

cep-purls.md
jaimergp: xref https://github.com/conda/ceps/pull/159

cep-purls.md
bollwyvl: Continuing a riff from [zulip](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/PURL.20and.20CVE.20mapper/with/600049336) as to where to store PURLs in an artifact's `info/`:

> if not SBOM, maybe `/purls/{:pedigree}.txt` where pedigree is a [CDX value](https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_pedigree)? (e.g. `ancestor`, usually exactly one, known at authoring time with `${{ version }}`, or `variant` which can probably only be known at build time in the general case).

tdejager: @bollwyvl I think this should be the way to go indeed, because for CVE's the `variants` field could be very interesting. Especially in cases like https://github.com/conda-forge/libpng-feedstock/blob/00d42af47191ac863343a6a8142a596ca069c192/recipe/recipe.yaml#L12 where we cannot map cleanly to a "source PURL" so to speak. 

In that sense we could use a `variant` like debian or something to piggyback on the vulnerabilities. However, the wording for CDX is:

> Variants describe relations where the relationship between the components is not known. For example, if Component A contains nearly identical code to Component B. They are both related, but it is unclear if one is derived from the other, or if they share a common ancestor.

I guess that in my debian libpng example, the ancestor is known to be the same. So not sure if we are using the field "technically" correct in that sense. :)

beckermr (changes requested):
A few comments:

1. We should not have so many optional things in a spec. For example, either all tools generate a `purls.json` artifact with a specific format, or they do not. I don't see any good reason to make this optional. Similarly, why would we let tools ignore the PURL field in the repo index.json files from packages, while others do not. **Each time we make things optional, we are adding branches, additional test cases, additional edge cases, and eventually additional bugs to downstream tools that implement the spec.** 
2. PURLS have to be patchable, if only for the reason that if someone has a bug, it should be fixed, not enshrined forever. Specifically for conda-forge, we may want to use repodata patching to centrally set the PURLS and so patching is a hard requirement.

cep-purls.md
bollwyvl: > some large rearrangements

Yep, this is handled authoritatively in:

https://github.com/package-url/purl-spec/blob/main/types/conda-definition.json

I don't like this:

```yaml
"namespace_definition": {
    "note": "there is no namspace",
    "requirement": "prohibited"
  },
```

... cuz, yeah, all `$CONDA_EXE` _are_ namespace-aware, and e.g. `bioconda`/`conda-forge` packages don't map 1:1. That being said, this section could be collapsed down to a reference to that spec file.

PR #70, opened 2024-04-12
CEP for the OCI storage of conda packages & repodata

wolfv:
Fill in the technical details of the OCI registry storage for reference.

### DIFF
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -1,7 +1,7 @@
 <table>
 <tr><td> Title </td><td> OCI registries as conda channels </td>
 <tr><td> Status </td><td> Proposed </td></tr>
-<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;<br> Hind Montassif &lt;hind.montassif@quantstack.net&gt;</td></tr>
 <tr><td> Created </td><td> April 12, 2024</td></tr>
 <tr><td> Updated </td><td> August 7, 2024</td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/70 </td></tr>
@@ -48,7 +48,6 @@ Custom mediaTypes defined for the conda channels use case are as follows:
 | repodata         | `repodata.json.zst` file  | application/vnd.conda.repodata.v1+json+zst  |
 | repodata         | `repodata.json.gz` file   | application/vnd.conda.repodata.v1+json+gzip |
 | repodata         | `repodata.json.bz2` file  | application/vnd.conda.repodata.v1+json+bz2  |
-| repodata         | `repodata.json.jlap` file | application/vnd.conda.jlap.v1               |
 
 If needed, more mediaTypes could be specified (i.e `application/vnd.conda.info.v1.tar+zst`).
 
@@ -70,12 +69,6 @@ Other encodings are also accepted:
 
 - `application/vnd.conda.repodata.v1+json+gzip`
 - `application/vnd.conda.repodata.v1+json+bz2`
-
-For `jlap`, the following mediaType is used:
-
-- `application/vnd.conda.jlap.v1`
-
-The `jlap` file should also be stored under the `<channel>/<subdir>/repodata.json` path as an additional layer.
 
 ### Conda package artifacts on an OCI registry
 
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -27,7 +27,8 @@ For example, a package like `xtensor-0.10.4-h431234.conda` would map to a OCI re
 
 A conda package, in an OCI registry, should ship up to 3 layers:
 
-- The package itself, as a tarball. (mandatory)
+- The package data itself, as a tarball. (mandatory)
+  - This can be either a `.tar.bz2` (v1) or a `.conda` (v2) file, or both as separate layers.
 - The package `info` folder as a gzipped "tar.gz" file.
 - The package `info/index.json` file as a plain JSON file.
 
@@ -48,7 +49,7 @@ It is used by conda to find packages in a channel.
 
 On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
 The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
-All versions of the repodata should also be tagged  with a UTC timestamp of the following format: `YYYY.MM.DD.HH.MM.SS`, e.g. `2024.04.12.07.06.32`.
+All versions of the repodata should also be tagged with a UTC timestamp of the following format: `YYYY.MM.DD.HH.MM.SS`, e.g. `2024.04.12.07.06.32`.
 
 The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zstd` as an additional layer in `<channel>/<subdir>/repodata.json`. ([ref](https://github.com/opencontainers/image-spec/blob/main/layer.md#gzip-media-types))
 
@@ -67,15 +68,24 @@ The `jlap` file should also be stored under the `<channel>/<subdir>/repodata.jso
 
 A given conda-package is identified by a URL like `<subdir>/<package-name>-<version>-<build>.<ext>` where `<subdir>` is the platform and architecture, `<package-name>` is the name of the package, `<version>` is the version of the package, `<build>` is the build string of the package, and `<ext>` is the extension of the package file.
 
+#### Mapping the package name
+
+> ![NOTE]
+> **Package names in the conda world**
+> The following regex is given by `conda/schemas` for a valid package name: `^[a-z0-9_](?!_)[._-]?([a-z0-9]+(\.|-|_|$))*$`
+> That means, a package can start with an alphanumeric character or a _single_ underscore (not multiple), and can contain dots, dashes, and underscores. It also has to end with a alphanumeric character (cannot end with a dot, dash, or underscore).
+
 To store this package on an OCI registry, we need to map it to a name and tag. The name is `<channel>/<subdir>/<package-name>`. The tag is `<version>-<build>`. There are some special rules for OCI registry names and tags for which we need some mapping. The regex for valid names is as follows:
 
 `[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` ([ref](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests))
 
 The regex expresses that names can only start with an alphanumeric letter.
 
-In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we replace a leading underscore with the string `zzz`.
+In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we prepend packages with a leading underscore with the string `zzz`. The name would thus be changed to `zzz_libgcc_mutex`.
 
-The tag is the version and build string of the packages, using a `-` as a separator. However, a OCI tag can only contain the following regex:
+#### Mapping the tag
+
+The tag is the version and build string of the packages, using a `-` as a separator. However, a OCI tag has to conform to the following regex:
 
 `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
 
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -50,7 +50,7 @@ All versions of the repodata should also be tagged  with a timestamp of the foll
 
 The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst` as an additional layer in `<channel>/<subdir>/repodata.json`.
 
-Other encoding are also accepted:
+Other encodings are also accepted:
 
 - `application/vnd.conda.repodata.v1+json+gzip`
 - `application/vnd.conda.repodata.v1+json+bz2`
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -119,6 +119,12 @@ Some characters that are used in the conda-forge repository as part of the build
 - `!` is replaced by `__e__`
 - `=` is replaced by `__eq__`
 
+> [!IMPORTANT]
+> 
+> The mapping of the package names and tags, as previously outlined, is solely intended for internal storage within an OCI registry.
+> 
+> From the user's perspective, or in contexts outside of the OCI registry—beyond the middleware that interacts with it—, the package names and tags will remain unchanged, preserving the original strings.
+
 #### Authentication
 
 Pulling a public image from a Container registry can be done anonymously ([ref](https://docs.github.com/en/packages/learn-github-packages/about-permissions-for-github-packages#visibility-and-access-permissions-for-packages)).
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -48,7 +48,7 @@ On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
 The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
 All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
 
-The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst`.
+The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst` as an additional layer in `<channel>/<subdir>/repodata.json`.
 
 Other encoding are also accepted:
 
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -139,15 +139,9 @@ Note that in the case of pulling repodata, the name `repodata.json` is always us
 
 ##### mamba
 
-In order to fetch packages from an OCI registry, we need to set a mirror (can be more than one) for the channel to be used (e.g `conda-forge`).
-This can be done in the rc file as follows:
+In order to fetch packages from an OCI registry, the corresponding URL should be used as a channel (i.e `oci://ghcr.io/channel-mirrors/conda-forge`).
 
-```
-mirrored_channels:
-  conda-forge: ["oci://ghcr.io/channel-mirrors/conda-forge"]
-```
-
-When a user requests installing a package (with the configuration set above, and using `conda-forge` channel), a set of requests to fetch `repodata.json` are first performed as follows:
+When a user requests installing a package, a set of requests to fetch `repodata.json` are first performed as follows:
 
 - A token is requested to anonymously pull `repodata.json` using the following URL:\
 `https://ghcr.io/token?scope=repository:channel-mirrors/conda-forge/<subdir>/repodata.json:pull`
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -1,0 +1,62 @@
+# OCI registries as conda channels
+
+We want to use OCI registries as a storage for conda packages. This CEP specifies how we lay out conda packages on an OCI registry.
+
+## Specification
+
+An OCI artifact consists of a manifest and a set of blobs. The manifest is a JSON document that describes the contents of the artifact. The blobs are the actual data that the manifest refers to. The manifest is stored in the registry as a blob, and the blobs are stored in the registry as blobs.
+
+The manifest consists of some metadata and a number of "layers". Each layer is a reference to a blob.
+
+Layers can have arbitrary names and mediaTypes.
+
+An OCI manifest is referenced by a name and a tag.
+
+### Conda package artifacts on an OCI registry
+
+The manifest for a conda package on an OCI registry should look like follows.
+
+It should have a name and a tag. The name is `<channel>/<subdir>/<package-name>`.
+The tag is the version and build string of the packages, using a `-` as a separator.
+
+For example, a package like `xtensor-0.10.4-h431234.conda` would map to a OCI registry `conda-forge/linux-64/xtensor:0.10.4-h431234`.
+
+### Layers
+
+A conda package, in an OCI registry, should ship up to 3 layers:
+
+- The package itself, as a tarball. (mandatory)
+- The package `info` folder as a gzipped "tar.gz" file.
+- The package `info/index.json` file as a plain JSON file.
+
+The mediaType for the different layers is as follows:
+
+- for a .tar.bz2 package, the mediaType is `application/vnd.conda.package.v1`
+- for a .conda package, the mediaType is `application/vnd.conda.package.v2`
+- for the `info` folder as gzip the mediaType is `application/vnd.conda.info.v1.tar+gzip`
+- for the `index.json` file the mediaType is `application/vnd.conda.info.index.v1+json`
+
+Using the `mediaType` field in the manifest, we can find the layer + SHA256 hash to pull the corresponding blob.
+Each `mediaType` should only be present in one layer.
+
+## Repodata on OCI registries
+
+The `repodata.json` file is a JSON file that contains metadata about the packages in a channel.
+It is used by conda to find packages in a channel.
+
+On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
+The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
+All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
+
+The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst`.
+
+Other encoding are also accepted:
+
+- `application/vnd.conda.repodata.v1+json+gzip`
+- `application/vnd.conda.repodata.v1+json+bz2`
+
+For `jlap`, the following mediaType is used:
+
+- `application/vnd.conda.jlap.v1`
+
+The `jlap` file should also be stored under the `<channel>/<subdir>/repodata.json` path as an additional layer.
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -1,4 +1,14 @@
-# OCI registries as conda channels
+<table>
+<tr><td> Title </td><td> OCI registries as conda channels </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> April 12, 2024</td></tr>
+<tr><td> Updated </td><td> August 7, 2024</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/70 </td></tr>
+<tr><td> Implementation </td><td>  </td></tr>
+</table>
+
+# Abstract
 
 We want to use OCI registries as a storage for conda packages. This CEP specifies how we lay out conda packages on an OCI registry.
 
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -12,6 +12,8 @@ Layers can have arbitrary names and mediaTypes.
 
 An OCI manifest is referenced by a name and a tag.
 
+For further details, please refer to the official [OCI Distribution spec](https://github.com/opencontainers/distribution-spec/blob/v1.0/spec.md#definitions).
+
 ### Conda package artifacts on an OCI registry
 
 The manifest for a conda package on an OCI registry should look like follows.
@@ -48,7 +50,7 @@ On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
 The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
 All versions of the repodata should also be tagged  with a UTC timestamp of the following format: `YYYY.MM.DD.HH.MM.SS`, e.g. `2024.04.12.07.06.32`.
 
-The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst` as an additional layer in `<channel>/<subdir>/repodata.json`.
+The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zstd` as an additional layer in `<channel>/<subdir>/repodata.json`. ([ref](https://github.com/opencontainers/image-spec/blob/main/layer.md#gzip-media-types))
 
 Other encodings are also accepted:
 
@@ -60,3 +62,25 @@ For `jlap`, the following mediaType is used:
 - `application/vnd.conda.jlap.v1`
 
 The `jlap` file should also be stored under the `<channel>/<subdir>/repodata.json` path as an additional layer.
+
+### Mapping a conda-package to the OCI registry
+
+A given conda-package is identified by a URL like `<subdir>/<package-name>-<version>-<build>.<ext>` where `<subdir>` is the platform and architecture, `<package-name>` is the name of the package, `<version>` is the version of the package, `<build>` is the build string of the package, and `<ext>` is the extension of the package file.
+
+To store this package on an OCI registry, we need to map it to a name and tag. The name is `<channel>/<subdir>/<package-name>`. The tag is `<version>-<build>`. There are some special rules for OCI registry names and tags for which we need some mapping. The regex for valid names is as follows:
+
+`[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` ([ref](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests))
+
+The regex expresses that names can only start with an alphanumeric letter.
+
+In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we replace a leading underscore with the string `zzz`.
+
+The tag is the version and build string of the packages, using a `-` as a separator. However, a OCI tag can only contain the following regex:
+
+`[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
+
+Some characters that are used in the conda-forge repository as part of the build string are not allowed in the OCI registry. For this reason, we use the following mapping:
+
+- `+` is replaced by `__p__`
+- `!` is replaced by `__e__`
+- `=` is replaced by `__eq__`
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -46,7 +46,7 @@ It is used by conda to find packages in a channel.
 
 On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
 The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
-All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
+All versions of the repodata should also be tagged  with a UTC timestamp of the following format: `YYYY.MM.DD.HH.MM.SS`, e.g. `2024.04.12.07.06.32`.
 
 The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst` as an additional layer in `<channel>/<subdir>/repodata.json`.
 
--- a/cep-oci.md
+++ b/cep-oci.md
@@ -70,9 +70,11 @@ A given conda-package is identified by a URL like `<subdir>/<package-name>-<vers
 
 #### Mapping the package name
 
-> ![NOTE]
+> [!NOTE]
 > **Package names in the conda world**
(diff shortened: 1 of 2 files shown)


wolfv:
- we should link refernce impls.
- we should say that conda / mamba / rattler use `oci://` as the scheme.

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+## Repodata on OCI registries
+
+The `repodata.json` file is a JSON file that contains metadata about the packages in a channel.
+It is used by conda to find packages in a channel.
+
+On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
+The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
+All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
+
+The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst`.
```
beckermr, line 51: I assume if the repodata is compressed, the path should be `<channel>/<subdir>/repodata.json.zstd` or w/e?

wolfv: No, it is another layer under the same name.

wolfv: The name is _always_ repodata.json

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+Each `mediaType` should only be present in one layer.
+
+## Repodata on OCI registries
+
+The `repodata.json` file is a JSON file that contains metadata about the packages in a channel.
+It is used by conda to find packages in a channel.
+
+On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
+The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
+All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
```
jaimergp, line 49:
```suggestion
All versions of the repodata should also be tagged  with a UTC timestamp of the following format: `YYYY.MM.DD.HH.MM.SS`, e.g. `2024.04.12.07.06.32`.
```

I assume it's UTC/GMT. Can we add seconds to it too? That's the resolution of the [`Last-Modified` timestamp in the http world](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Last-Modified).

cep-oci.md
beckermr: Ahhhhhh. This may be clear to you all but I either missed the text or didn't catch that.

jaimergp: Looks like different compression schemas are distributed under different layers instead of new paths.

jaimergp, line 51:
```suggestion
The mediaType for the raw `repodata.json` (with `repodata_version: 1`) file is `application/vnd.conda.repodata.v1+json`. Subsequent versions like `2` would be stored under `application/vnd.conda.repodata.v2+json`, and so on. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst`.
```

Assuming this is true :)

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+It should have a name and a tag. The name is `<channel>/<subdir>/<package-name>`.
+The tag is the version and build string of the packages, using a `-` as a separator.
+
+For example, a package like `xtensor-0.10.4-h431234.conda` would map to a OCI registry `conda-forge/linux-64/xtensor:0.10.4-h431234`.
+
+### Layers
+
+A conda package, in an OCI registry, should ship up to 3 layers:
+
+- The package itself, as a tarball. (mandatory)
```
jaimergp, line 28: You mean the contents of the package, or the compressed artifact (tarball or not)? In `.conda` the outer layer is actually a ZIP file, and the inner ones are zstd tarballs. Would be helpful to clarify.

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+The `repodata.json` file is a JSON file that contains metadata about the packages in a channel.
+It is used by conda to find packages in a channel.
+
+On an OCI registry it should be stored under `<channel>/<subdir>/repodata.json`.
+The repodata file should have one entry that has the `latest` tag. This entry should point to the latest version of the repodata.
+All versions of the repodata should also be tagged  with a timestamp of the following format: `YYYY.MM.DD.HH.MM`, e.g. `2024.04.12.07.06`.
+
+The mediaType for the raw `repodata.json` file is `application/vnd.conda.repodata.v1+json`. However, for large repositories it's advised to store the `zstd` encoded repodata file with the mediaType `application/vnd.conda.repodata.v1+json+zst`.
+
+Other encoding are also accepted:
```
jaimergp, line 53:
```suggestion
Other encodings are also accepted:
```

Not sure if encodings is the best term here though. I would understand it clearer as "compression schemes" or similar.

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+### Conda package artifacts on an OCI registry
+
+The manifest for a conda package on an OCI registry should look like follows.
+
+It should have a name and a tag. The name is `<channel>/<subdir>/<package-name>`.
+The tag is the version and build string of the packages, using a `-` as a separator.
+
+For example, a package like `xtensor-0.10.4-h431234.conda` would map to a OCI registry `conda-forge/linux-64/xtensor:0.10.4-h431234`.
+
+### Layers
```
jaimergp, line 24: A table with the different layer `mediaTypes` and the expected contents would make this section super easy to read.

cep-oci.md
wolfv: Right, should leave the tarball out of the sentence. It's the package data itself.

jaimergp:
> we should say that conda / mamba / rattler use oci:// as the scheme.

What does a `oci://` URL look like? Is it a direct translation of the anaconda.org artifact that gets manipulated? How is the server referred to? Basically, which one is correct:

- `oci://ghcr.io/channel-mirrors/conda-forge/linux-64/xtensor:0.10.4-h431234`
- `oci://ghcr.io/channel-mirrors/conda-forge/linux-64/xtensor-0.10.4-h431234`
- `oci://ghcr.io/channel-mirrors/conda-forge/linux-64/xtensor-0.10.4-h431234.tar.bz2`

Is it possible to refer to a layer directly by URL? In case of repodata:

- `oci://ghcr.io/channel-mirrors/conda-forge/linux-64/repodata.json.zstd`
- `oci://ghcr.io/channel-mirrors/conda-forge/linux-64/repodata.json:zstd`

I think this needs to be standardized too.

cep-oci.md
```diff
@@ -0,0 +1,62 @@
+# OCI registries as conda channels
+
+We want to use OCI registries as a storage for conda packages. This CEP specifies how we lay out conda packages on an OCI registry.
+
+## Specification
```
jaimergp, line 5: The distribution-spec contains useful definitions for these terms:

https://github.com/opencontainers/distribution-spec/blob/v1.0/spec.md#definitions

cep-oci.md
jaimergp: Also, some artifacts in `defaults` are available both as .tar.bz2 and .conda. Should we restrict one package layer for each label? I don't think we need to.

cep-oci.md
jaimergp: Are there any guarantees of availability for older timestamps or will they eventually get deleted? What about the repodata variants like current_repodata.json and repodata_from_packages.json? Same about channeldata.json.

cep-oci.md
```diff
@@ -0,0 +1,62 @@
...
+
+- The package itself, as a tarball. (mandatory)
+- The package `info` folder as a gzipped "tar.gz" file.
+- The package `info/index.json` file as a plain JSON file.
+
+The mediaType for the different layers is as follows:
+
+- for a .tar.bz2 package, the mediaType is `application/vnd.conda.package.v1`
+- for a .conda package, the mediaType is `application/vnd.conda.package.v2`
+- for the `info` folder as gzip the mediaType is `application/vnd.conda.info.v1.tar+gzip`
```
jaimergp, line 36: The info folder can get heavy with licenses, test files and what not. Will we allow `.tar+zstd`, for example?

cep-oci.md
wolfv: Good point(s). We could add them under their respective file names.

wolfv: Regarding the timestamp availability - that's outside the scope of this CEP IMO. The latest tag is the only requirement for working clients.

cep-oci.md
wolfv: Everything would work fine if there are two layers, they should just not have the same mediaType.

cep-oci.md
wolfv: Feel free to format and push to this PR! :) This is a collaborative document

cep-oci.md
beckermr: I 100% agree. There are a lot of potential issues around serving old repodata related to patching, removals, etc. Best to defer and keep this spec as small as possible.

wolfv:
Yeah, the nice thing is that the URLs map directly to the OCI registry. There just needs to be some post-processing for the tags. 

E.g. instead of

```
https://conda.anaconda.org/conda-forge/linux-64/repodata.json
```

We ask for 

```
oci://ghcr.io/channel-mirrors/conda-forge/linux-64/repodata.json:latest
```

And for packages

```
https://conda.anaconda.org/conda-forge/linux-64/numpy-1.23.1-h123123.tar.bz2
```

We ask for 

```
oci://ghcr.io/channel-mirrors/conda-forge/linux-64/numpy:1.23.1-h123123
```

wolfv:
One thing that also needs to go into this document is how tags are formatted. Some versions cannot be translated to tags directly because of the rules of OCI registries.

The functions are here:

https://github.com/mamba-org/rattler/blob/a2073aa6b92196c50208d39fc6b6c67469bf7810/crates/rattler_networking/src/oci_middleware.rs#L77-L92

jaimergp:
Food for thought: implement subdirs with [platform metadata per layer](https://github.com/opencontainers/image-spec/blob/main/image-index.md#platform-variants).

jaimergp:
> Some versions cannot be translated to tags directly because of the rules of OCI registries.

For reference, rules are here: https://github.com/opencontainers/distribution-spec/blob/v1.0/spec.md#pulling-manifests

Hind-M:
>     * we should say that conda / mamba / rattler use `oci://` as the scheme.
Hmm, AFAIK we are rather using `https` as scheme in mamba: `https://ghcr.io/...`
Last time I checked, I think curl was complaining about unknown `oci` protocol...

wolfv:
Yeah, but regular HTTPs requests don't work. That's why we need a middleware or some other layer that converts `oci://...` requests to `https://` before sending them to cURL.

cep-oci.md
wolfv: I linked it :)

cep-oci.md
```diff
@@ -0,0 +1,98 @@
...
+- The package data itself, as a tarball. (mandatory)
+  - This can be either a `.tar.bz2` (v1) or a `.conda` (v2) file, or both as separate layers.
+- The package `info` folder as a gzipped "tar.gz" file.
+- The package `info/index.json` file as a plain JSON file.
+
+The mediaType for the different layers is as follows:
+
+- for a .tar.bz2 package, the mediaType is `application/vnd.conda.package.v1`
+- for a .conda package, the mediaType is `application/vnd.conda.package.v2`
+- for the `info` folder as gzip the mediaType is `application/vnd.conda.info.v1.tar+gzip`
```
wolfv, line 39: I am wondering if this should have been `+tar+gzip` and not `.tar+gzip` :(

wolfv: Ah, no, looks like that's what OCI people do: https://github.com/opencontainers/image-spec/blob/main/layer.md#image-layer-filesystem-changeset :)

cep-oci.md
dholth: It would be nice to be able to use the unmodified info-pip-23.3.1-py312haa95532_0.tar.zst out of the .conda

wolfv: Indeed. I used the tar.gz approach because of how easy it is to open and explore it with pure Python. But the unmodified one makes sense as well. WE could specify that `application/vnd.conda.info.v1.tar+zst` will also be accepted?

dholth: Makes sense.
.tar.gz just seemed odd although gzip is very ordinary it hasn't been used in conda packaging.

Hind-M:
Added a [commit](https://github.com/Hind-M/ceps/commit/d339ca0e17a94c86a489180da0768ac31023c909) but couldn't push it here.

wolfv:
Looks good, @Hind-M. Maybe you can make a PR against my branch (https://github.com/wolfv/ceps/tree/oci-cep). I can then merge it there, and the PR will be updated.

Hind-M:
Here you are https://github.com/wolfv/ceps/pull/3. Thanks!

Hind-M:
The rendered version for convenience: https://github.com/wolfv/ceps/blob/oci-cep/cep-oci.md

cep-oci.md
```diff
@@ -96,3 +115,39 @@ Some characters that are used in the conda-forge repository as part of the build
...
+
+In order to fetch packages from an OCI registry, we need to set a mirror (can be more than one) for the channel to be used (e.g `conda-forge`).
+This can be done in the rc file as follows:
+
+```
+mirrored_channels:
+  conda-forge: ["oci://ghcr.io/channel-mirrors/conda-forge"]
+```
+
+When a user requests installing a package (with the configuration set above, and using `conda-forge` channel), a set of requests to fetch `repodata.json` are first performed as follows:
```
beckermr, line 150: Why are we marking these as mirrored instead making them first-class things via using the normal channels config setting with an `oci://` prefix to mark OCI channels?

beckermr (changes requested):
One question.

cep-oci.md
wolfv: I believe this is orthogonal - you should be able to use the `oci://` URL as your only channel as well.

beckermr: Then why do we need the `mirrored_channels` option at all?

wolfv: Maybe not in this CEP? Specifying how we expect channel mirroring works could have merits. Specifically, we should define where the `repodata.json` file comes from when we have multiple mirrors.

Ideally we would also define how we combine download counts from multiple mirrors (but all that would definitely belong into a different CEP).

beckermr: Yeah we should handle mirroring in another CEP I would think.

Hind-M: Yes it's orthogonal. I changed it to not mention mirrors anymore.

Hind-M:
@conda/steering-council

This vote falls under the "[Enhancement Proposal Approval](https://github.com/conda/governance/blob/702278b8f19be31423a9b9c7d6955de4d2a8c2df/README.md#enhancement-proposal-approval)" policy of the [conda governance policy](https://github.com/conda-incubator/governance),

It needs **60%** of the Steering Council to vote yes to pass.

**This vote will end on 2024-09-02, End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

To **vote**, please use the **form** below:

@xhochy (Uwe Korn)

- [ ] yes
- [ ] no
- [ ] abstain

@CJ-Wright (Christopher J. 'CJ' Wright)

- [ ] yes
- [ ] no
- [ ] abstain

- [ ] yes
- [ ] no
- [ ] abstain

- [ ] yes
- [ ] no
- [ ] abstain

@chenghlee (Cheng H. Lee)

- [ ] yes
- [ ] no
- [ ] abstain

@ocefpaf (Filipe Fernandes)

- [ ] yes
- [ ] no
- [ ] abstain

- [ ] yes
- [ ] no
- [ ] abstain

@msarahan (Michael Sarahan)

- [ ] yes
- [ ] no
- [ ] abstain

@mbargull (Marcel Bargull)

- [ ] yes
- [ ] no
- [ ] abstain

@jakirkham (John Kirkham)

- [ ] yes
- [ ] no
- [ ] abstain

@jezdez (Jannis Leidel)

- [ ] yes
- [ ] no
- [ ] abstain

@wolfv (Wolf Vollprecht)

- [ ] yes
- [ ] no
- [ ] abstain

- [ ] yes
- [ ] no
- [ ] abstain

@kkraus14 (Keith Kraus)

- [ ] yes
- [ ] no
- [ ] abstain

@baszalmstra (Bas Zalmstra)

- [ ] yes
- [ ] no
- [ ] abstain

cep-oci.md
```diff
@@ -0,0 +1,157 @@
+<table>
+<tr><td> Title </td><td> OCI registries as conda channels </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
```
jezdez, line 4: @Hind-M Could you please add yourself to the author list?

cep-oci.md
```diff
@@ -0,0 +1,157 @@
...
+Other encodings are also accepted:
+
+- `application/vnd.conda.repodata.v1+json+gzip`
+- `application/vnd.conda.repodata.v1+json+bz2`
+
+For `jlap`, the following mediaType is used:
+
+- `application/vnd.conda.jlap.v1`
+
+The `jlap` file should also be stored under the `<channel>/<subdir>/repodata.json` path as an additional layer.
```
jezdez, line 78: Since the JLAP proposal has not been accepted, and sharded repodata approved instead, it may not be needed to keep this in here.

cep-oci.md
```diff
@@ -0,0 +1,157 @@
...
+### Mapping a conda-package to the OCI registry
+
+A given conda-package is identified by a URL like `<subdir>/<package-name>-<version>-<build>.<ext>` where `<subdir>` is the platform and architecture, `<package-name>` is the name of the package, `<version>` is the version of the package, `<build>` is the build string of the package, and `<ext>` is the extension of the package file.
+
+#### Mapping the package name
+
+> [!NOTE]
+> **Package names in the conda world**
+> 
+> The following regex is given by [`conda/schemas`](https://github.com/conda/schemas/blob/473708ac97283708d6664cbd89b8049ad1623489/common-1.schema.json#L58-L82) for a valid package name: `^[a-z0-9_](?!_)[._-]?([a-z0-9]+(\.|-|_|$))*$`
```
jezdez, line 105: It would be good to not rely on conda/schema alone, and verify with the conda code base about the regex, in case the schema isn't up-to-date.

cep-oci.md
```diff
@@ -0,0 +1,157 @@
...
+> 
+> That means, a package can start with an alphanumeric character or a _single_ underscore (not multiple), and can contain dots, dashes, and underscores. It also has to end with a alphanumeric character (cannot end with a dot, dash, or underscore).
+
+To store this package on an OCI registry, we need to map it to a name and tag. The name is `<channel>/<subdir>/<package-name>`. The tag is `<version>-<build>`. There are some special rules for OCI registry names and tags for which we need some mapping. The regex for valid names is as follows:
+
+`[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` ([ref](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests))
+
+The regex expresses that names can only start with an alphanumeric letter.
+
+In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we prepend packages with a leading underscore with the string `zzz`. The name would thus be changed to `zzz_libgcc_mutex`.
```
jezdez, line 115:
```suggestion
In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we prepend packages with a leading underscore with the string `internal`. The name would thus be changed to `internal_libgcc_mutex`.
```

`internal` seems more fitting than `zzz` 💤

cep-oci.md
```diff
@@ -0,0 +1,157 @@
...
+
+The tag is the version and build string of the packages, using a `-` as a separator. However, a OCI tag has to conform to the following regex:
+
+`[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`
+
+Some characters that are used in the conda-forge repository as part of the build string are not allowed in the OCI registry. For this reason, we use the following mapping:
+
+- `+` is replaced by `__p__`
+- `!` is replaced by `__e__`
+- `=` is replaced by `__eq__`
```
jezdez, line 127: That list seems to be a little random, it looks like Python dunder methods, but not quite valid, how about this instead?

```suggestion
- `+` is replaced by `__add__`
- `!` is replaced by `__not__`
- `=` is replaced by `__eq__`
```

cep-oci.md
jaimergp: I'm afraid we might be late here. [~74 packages are mirrored like this already](https://github.com/orgs/channel-mirrors/packages?tab=packages&q=zzz). We would need to delete and re-mirror? Or can can we just copy images and labels?

cep-oci.md
jaimergp: `p` for `plus` and `e` for `exclamation`, I guess? These characters might not mean "adding" or "negating" in a version/build string, so I'm not sure whether this is a good way to solve the "randomness" here. These names are only for the OCI names, anyway, right? They are renamed once downloaded.

cep-oci.md
```diff
@@ -0,0 +1,157 @@
...
+- The package data itself, as a tarball. (mandatory)
+  - This can be either a `.tar.bz2` (v1) or a `.conda` (v2) file, or both as separate layers.
+- The package `info` folder as a gzipped "tar.gz" file.
+- The package `info/index.json` file as a plain JSON file.
+
+### Mapping a conda-package to the OCI registry
+
+A given conda-package is identified by a URL like `<subdir>/<package-name>-<version>-<build>.<ext>` where `<subdir>` is the platform and architecture, `<package-name>` is the name of the package, `<version>` is the version of the package, `<build>` is the build string of the package, and `<ext>` is the extension of the package file.
+
+#### Mapping the package name
```
jaimergp, line 100: Are these "mappings" reverted to their original names once downloaded/extracted? I would assume so and I think that should be part of the CEP.

cep-oci.md
jaimergp: Also, how can we safely remove `internal_` from the names? It might be something being used already. What about `internal___` (triple underscore)?

cep-oci.md
Hind-M: > `p` for `plus` and `e` for `exclamation`, I guess? 

Yes!
These mappings (prepending the packages names starting with `_` with `zzz` and replacing `+` etc for tags) are really only intended to make them valid in order to be stored in an OCI registry (correct me if I'm wrong @wolfv). It doesn't impact anything elsewhere as the names and tags are still visible (by users when downloaded for example) as their original strings.
So I don't think having `internal` instead of `zzz` (same for the tags suggestions) would really matter (apart from making sure that it wouldn't conflict with potential similar existing strings).
While something more explicit would be preferable,  given that a bunch of packages have already been mirrored that way as @jaimergp  mentioned, I would say to keep it as is?

beckermr: Is there a standard encoding scheme we can rely on here instead of inventing our own?

cep-oci.md
Hind-M: I added a [note](https://github.com/conda/ceps/pull/70/commits/6e6cd7d8f8c6dd337ce3d2f94f83103061a72387) for that.

cep-oci.md
beckermr: Also, if we insist of a special encoding, we can add underscores themselves as `__under__`, which would render our scheme unambiguous, since an existing package with `__under__` in it would be mapped to five `__under__` in a row.

cep-oci.md
beckermr: See my comment below on making this not ambiguous by merging this part of the spec with other encodings.

beckermr:
I am emeritus so I cannot vote here, but IMHO the issues around encoding/decoding above need to be resolved before this spec can be passed.

cep-oci.md
jezdez: > I'm afraid we might be late here. [~74 packages are mirrored like this already](https://github.com/orgs/channel-mirrors/packages?tab=packages&q=zzz). We would need to delete and re-mirror? Or can can we just copy images and labels?
The mirrors are proofs of concept from the conda community's perspective (I think from conda-forge as well), so I don't think these matter. Re-mirroring 74 packages seems a small price to pay for a better spec.

beckermr: 100% agreed @jezdez readding 74 packages is simple enough.

jakirkham:
It seems like there is a fair amount of discussion still happening here for an active vote. Should we cancel the vote and reschedule after that discussion has reached a conclusion?

cep-oci.md
wolfv: @beckermr - underscores are allowed, but not at the beginning of the package name. Not sure what you mean. 

We could encode the strings with something like SHA256 or MD5 but then they are not easily recoverable and more cryptic. We selected this "encoding" to make it work, while still possible to map back (even as a human) to the source string. I am not particularly attached to this encoding, so very open to alternative suggestions.

- We could remove the offending characters (but then it might not be unique per package name anymore, so might not work
- We could append a hash, and remove the offending characters. That should work fine, but look weird, too

cep-oci.md
```diff
@@ -0,0 +1,156 @@
...
+> 
+> That means, a package can start with an alphanumeric character or a _single_ underscore (not multiple), and can contain dots, dashes, and underscores. It also has to end with a alphanumeric character (cannot end with a dot, dash, or underscore).
+
+To store this package on an OCI registry, we need to map it to a name and tag. The name is `<channel>/<subdir>/<package-name>`. The tag is `<version>-<build>`. There are some special rules for OCI registry names and tags for which we need some mapping. The regex for valid names is as follows:
+
+`[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` ([ref](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests))
+
+The regex expresses that names can only start with an alphanumeric letter.
+
+In `conda`, names can start with an underscore and it is used by conda-forge (e.g. `_libgcc_mutex`). For this reason, we prepend packages with a leading underscore with the string `zzz`. The name would thus be changed to `zzz_libgcc_mutex`.
```
baszalmstra, line 108: What happens if there is also a package called zzz_libgcc_mutex? 

This seems like a possible attack vector.

baszalmstra: This is similar to symbol name mangling I guess? Perhaps we should prefix all packages with a common prefix and a special one for underscores?

cep-oci.md
baszalmstra: > Is there a standard encoding scheme we can rely on here instead of inventing our own?

The text itself mentiones that package names can start with an underscore. _libgcc_mutex is the example given.

Perhaps we can learn from symbol name mangling instead of coming up with a unique scheme?

jezdez: Another interesting thing from the [spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests):

> Implementers note: Many clients impose a limit of 255 characters on the length of the concatenation of the registry hostname (and optional port), /, and <name> value. If the registry name is registry.example.org:5000, those clients would be limited to a <name> of 229 characters (255 minus 25 for the registry hostname and port and minus 1 for a / separator). For compatibility with those clients, registries should avoid values of <name> that would cause this limit to be exceeded.

Do we see this as a risk for hosting channels with long names? Does this character limit only apply for fetching the manifest or also artifacts in real world OCI implementations?

beckermr: @wolfv I made a mistake above, but the point stands. Suppose we have two packages:

 - zzzblah
 - _blah

The current scheme replacing only leading underscores with `zzz` would render the packages as 

 - zzzblah
 - zzzblah

creating a conflict.

However, if we encode underscores with a pattern that uses underscores and always encode all of them, say `under__`, then for a potential conflict, we always mangle the encoding string too. For example, with

 - under__blah
 - _blah

we get

 - underunder__under__blah
 - under__blah

The mistake I made above was using a pattern with a leading underscore for an underscore. That won't work.

cep-oci.md
beckermr: See my comments above on this.

Hind-M:
> It seems like there is a fair amount of discussion still happening here for an active vote. Should we cancel the vote and reschedule after that discussion has reached a conclusion?

I think we can wait and see if we reach a conclusion before the voting deadline, and possibly extend it if necessary. If not, we would definitely reschedule yes.
I'm not familiar with the usual processes (of voting and such), so if this isn't the correct approach, please let me know and I'll adjust accordingly.

cep-oci.md
jaimergp: Then we need to forbid using `under_` as a part of package name in conda because if we have a package named `under__something` (notice the extra underscore), we won't know if the encoded result `under_under_` is actually `_under_something`, `under__something` or `__something` :D

jaimergp: The other question: do we need to map back? Or can we just check the internal metadata at that point? I think we just need to go back from `original` to `mangled`, but not back, right? Once we have fetched `mangled` we can check if the metadata matches `original`. So I think we are fine as long as original maps to mangled in a unique way, and that might mean "exploding" all the original underscores into `under_`.

jaimergp:
> This is similar to symbol name mangling I guess? Perhaps we should prefix all packages with a common prefix and a special one for underscores?

I like this proposal by @baszalmstra, but we will need to remirror the whole thing.

cep-oci.md
beckermr: > Then we need to forbid using under_ as a part of package name in conda because if we have a package named under__something (notice the extra underscore), we won't know if the encoded result under_under_ is actually _under_something, under__something or __something :D

No. @jaimergp  `under__something` gets translated to `underunder__under__somthing`. **Every underscore gets translated.**

beckermr: I have not. The issue with any scheme is that it cannot start with an `_` because the OCI package name cannot start with an `_`.

beckermr:
I think the name length limit is the bigger issue here though. I guess we'll need a fixed length hash of the name.

wolfv:
Let me clarify some things:

Name attacks are not really possible (for now) since we also mirror the repodata from conda-forge / Anaconda.org. From the repodata, we just use the SHA256 hash to directly reference the right blob. The names & tags for the individual packages are mainly "cosmetic".

Even if we would run our own indexing, we would refer back to the stored `index.json` file (and the name in there) and not derive back the original name from the OCI image.
So I am personally not super concerned by this.

But I can sympathize with a mapping that would disallow any such overlaps.

I also think that re-mirroring might not be a big issue, since we just need to move the names / tags in the OCI registry to the right places (SHA hashes and packages will stay the same).

Hind-M:
So just to wrap up here:
- Since we are constrained by the limit of characters, we could say that we should avoid mapping every underscore.
- The names should be easily identifiable by users (useful when searching for packages in the registry), so using a hash instead of the name would be cryptic.
- Appending the name hash or a part of it sounds weird (and somehow redundant?) - even if we do that for every package to be consistent (not only the ones starting with an underscore), but we still can.
- Prepending a few characters like the first strategy (_zzz_) seems to be a good compromise.
Maybe use something else instead like @jezdez suggested? (_pkg_, _oci_, _int_ (for internal)...)
I have a preference for _oci_, since it's suggesting that this is an _oci_ specific thing.
We can do this for all packages or just for the package names starting with `_`.

Should we think of handling exceeding the max limit of characters as well? (I'm not sure if this is something already taken care of).

beckermr:
We CANNOT only prepend to packages that start with an underscore. This would declare part of the package namespace off limits for all conda users. That action requires at minimum a separate CEP where we codify package names formally. We have to preprend to everything. 

And yes we need to handle the max characters limit properly in this CFEP.

dholth:
Where is the specification for the length limit?
Do users have business browsing through the raw OCI anyway?

beckermr:
Tags have a limit of 128 characters: https://docs.docker.com/reference/cli/docker/image/tag/
names have limits. See the implementers note in the OCI spec for `pull` for other practical limits: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull.

It is hard to know if people want to browse an OCI registry. For sure github provides searches based on image names. See for example the existing OCI mirror (https://github.com/orgs/channel-mirrors/packages) based on a beta spec.

cep-oci.md
baszalmstra: Reading all of the above, I think we want:

1. Each package should have a unique representation when converted to a format that is accepted in OCI (I will refer to it as mangled).
2. Mangled package names should be kept as short as possible because there is a limit on the number of characters (128).
3. The mangled names should not stray too far away from the unmangled human-readable text to make it easy on users when navigating an OCI registry (otherwise base64 encoding would also suffice). 

So how about the following:

* We prepend all package names with a `c`. Packages that start with an underscore instead start with an uppercase `C`.     
* We escape any character that is not allowed with an underscore including the underscore itself.

    `+` -> `_P`
    `!` -> `_N`
    `=` -> `_E`
    `_` -> `__`
    
    I used uppercase letters here because they standout between the usually lowercase text.
    
E.g: 

    `conda` -> `cconda`
    `_libgcc_mutex` -> `Clibgcc_mutex`
     
Tags:

* `1!0.0.0_dev0+placeholder` -> `1_N0.0.0__dev0_Pplaceholder`

Thoughts?

wolfv: @beckermr - when writing out on the file system we would use the real name anyways (not the mapped name).

Although I would have to double-check if uppercase is allowed in OCI registries, and wether the registry is case-sensitive.

baszalmstra: >This will not work on case-insensitive file systems, which includes OSX in some cases.

I added case-insensitivity as a requirement and modified the example.

beckermr: @baszalmstra It appears that lower-case only is required by the OCI spec for container names. The names must match this regex `[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` [here](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull) IIUIC.

beckermr: Tags can have upper case though. The regex is `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}` from the same document.

baszalmstra: Are conda package names case sensitive?

baszalmstra: Based on your findings do you want to change anything with regard to the name mangling suggested above?

beckermr: Conda package names are always lower case, so presumably case insensitive.

beckermr: @baszalmstra we might as well have everything lowercase since that is what the OCI registry will look like.

Hind-M:
FYI, and considering the still ongoing discussions, the vote has been postponed to a later date that is yet to be determined.

cep-oci.md
dholth: All of these schemas are probably wrong in some way.

cep-oci.md
dholth: Case insensitive filesystems are usually case preserving though.

thomasjpfan:
> It is hard to know if people want to browse an OCI registry. For sure github provides searches based on image names.

As a user, as long as https://conda-forge.org/packages/ (or https://bioconda.github.io/conda-package_index.html) has all the correct metadata and points to the correct images in the oci registry, I do not mind if the oci image name itself is hashed.

Concretely, I would be okay with

```
conda-forge/linux-64/43asfdsafr9434234:321424fdasfsdf
```
 
which has the format `<hash_package_name>:<hash_version+build>`

If it's important to keep some of the original package name in the oci image name, another option is to strip away the invalid characters in the names/tags and add the hash next to it:

```
<stripped_package_name><hashed_package_name>:<stripped_version+build>-<hash_version+build>
```

Something like this:

```
conda-forge/linux-64/numpy-fsafsdaffs:2.2.2-py313hc518a0f_0-43asfdsafr9434234
```

The stripped package names and tags can also be trimmed to make sure they are under the character limits.

cep-oci.md
beckermr: This regex is wrong. It flags things like `p_astro` which are fine.

beckermr: This one appears to be correct:

```
r"^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$"
```

beckermr:
I wrote some code to examine all packages from conda-forge for invalid names, the length of the OCI-encoded package, etc. Here are the results 

```
current maximum oci length: 117
invalid conda dist: pyside2-2.0.0~alpha0-py27_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/linux-64/cpyside2:2.0.0~alpha0-py27__0 (label: broken-and-invalid)
invalid conda dist: pyside2-2.0.0~alpha0-py35_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/linux-64/cpyside2:2.0.0~alpha0-py35__0 (label: broken-and-invalid)
invalid conda dist: pyside2-2.0.0~alpha0-py36_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/linux-64/cpyside2:2.0.0~alpha0-py36__0 (label: broken-and-invalid)
invalid conda dist: pyside2-2.0.0~alpha0-py27_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/osx-64/cpyside2:2.0.0~alpha0-py27__0 (label: broken-and-invalid)
invalid conda dist: pyside2-2.0.0~alpha0-py35_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/osx-64/cpyside2:2.0.0~alpha0-py35__0 (label: broken-and-invalid)
invalid conda dist: pyside2-2.0.0~alpha0-py36_0.tar.bz2 (label broken-and-invalid)
invalid oci dist: conda-forge/osx-64/cpyside2:2.0.0~alpha0-py36__0 (label: broken-and-invalid)
```

The maximum length is the combined length of the oci image name and the tag. conda-forge has a few invalid artifact names that we've known about for a while.

Overall, we're pretty far from the oci limits which is good. 

However, we should put something in the spec to handle cases where the limits are exceeded. I will write a proposal for that.

beckermr:
cc @jaimergp

cep-oci.md
jaimergp: Thinking again about this, the leading underscore doesn't really affect _package names_ but _channel names_, given our image name scheme: `channel/subdir/package`. So maybe we don't have to worry too much about this limitation?

beckermr: I had the same thought but it turns out the underscore is not allowed after any other separator according the regex in the spec.

jaimergp: Ugh... Ok, then I said nothing! 😬

beckermr:
I have a working version of the spec with the updates from @baszalmstra and the hashing. Should we make a new PR, or should I edit this one directly. There is a failed vote on this PR, so I think a new one might be cleaner, but I don't want to erase history either.

jaimergp:
> There is a failed vote on this PR, so I think a new one might be cleaner, but I don't want to erase history either.

I'm partial to opening a new PR with sufficient references to this PR and the open discussions.

jaimergp:
Closing in favor of https://github.com/conda/ceps/pull/115. Thank you everyone for your comments and feedback!

PR #56, opened 2023-06-14
add cep for definition of new keys & values in build format

wolfv:

PR #71, opened 2024-04-12
CEP 39 - A new recipe format (part 3): `jinja` functions in recipes

wolfv:


h-vetinari:
In https://github.com/conda/conda-build/issues/5053 you had said:
> I do think that for the future (ie. rattler-build) we'll have an upcoming CEP to define this feature better.

Is this that CEP? In any case, I don't see `{{ stdlib(...) }}` in the diff, and I think it needs to be covered in _some_ way, given that we're now rolling this out at scale.

wolfv:
Yep, this is the one. Do you want to have it as a separate function or rolled into `compiler`? I prefer to roll it into `compiler`...

h-vetinari:
> Yep, this is the one. Do you want to have it as a separate function or rolled into `compiler`? I prefer to roll it into `compiler`...

I'd prefer a separate function, because it's less magic to have to understand ("why do I need to specify `c_stdlib_version` and how does that affect my build environment?")

wolfv:
My points:

- There are already two variables for compilers (`c_compiler` and `c_compiler_version`). Adding additional 2 variables (that are non-mandatory) is fine IMO, especially if it's well-documented.
- It complicates things for beginners. I think recipes would continue to work without adding the stdlib (correct?). So it's something they _should_ add but don't have to add. For `compiler` it's clear - the recipe will not work without a compiler.
- In most cases, it will _always_ be added when there is a compiler. 

I can imagine two behaviors:

- We control the stdlib behavior from the `variant_config`. This means, if the variant config contains entries for `c_stdlib` / `c_stdlib_version` we use them from the `compiler` function.
- We add additional arguments to the jinja function, so that `compiler('c', stdlib='glibc')`.

Note that while you are correctly pointing out a dislike about magical behavior, we are printing _a lot_ more information, clearly displayed, with rattler-build. So we can make it obvious where the `stdlib` comes from (we already point out that the `clangxx_osx-arm64` package comes from the compiler, for example).

And lastly, I just think it's a matter of writing really good documentation.

h-vetinari:
> I think recipes would continue to work without adding the stdlib (correct?).

Define "work". When we increase the `c_stdlib_version` on linux and macos later this year, the artefacts produced by a recipe without `{{ stdlib("c") }}` would continue to work on modern systems, but would be broken on systems with a too-old C standard library, because it's through the meta-packages pulled in by `stdlib` that we set the correct run-exports in the artefact metadata.

It would only affect very old systems, but I'd still consider that broken.

> So it's something they _should_ add but don't have to add. For `compiler` it's clear - the recipe will not work without a compiler.

For me it's the opposite here - a compiled recipe needs (in 99.9% of cases) a C standard library just as much as it needs a compiler. Up until now this was just an ambient assumption built-in to our infrastructure, but I'd prefer people to learn this at the same time as the necessity of `{{ compiler(...) }}` - the two go hand in hand (as implied by having the same `"c"` argument, and the same way the keys are constructed), even though the relevant versions are completely independent from each other.

> I can imagine two behaviors:
>  * We add additional arguments to the jinja function, so that `compiler('c', stdlib='glibc')`.

I really dislike this, because it would replace the nicely universal
```yaml
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
```
with a horrible mess of platform-specific selectors (or however the syntax or that ends up looking like in the new recipe format):
```yaml
    - {{ compiler('c', stdlib='glibc') }}                     # [linux]
    - {{ compiler('c', stdlib='macosx_deployment_target') }}  # [osx]
    - {{ compiler('c', stdlib='vs') }}                        # [win]
``` 

> * We control the stdlib behavior from the `variant_config`. This means, if the variant config contains entries for `c_stdlib` / `c_stdlib_version` we use them from the `compiler` function.

I agree that this is technically feasible, and with good output and documentation would be a workable solution. When I say I prefer to keep `{{ stdlib("c") }}` explicit, this is not an all-or-nothing statement, but more a 70:30 kind of preference over the implicit handling through the compiler.

One reason that pushes this from a 60:40 (based on "less magic") to a 70:30 preference is that the C stdlib has a special role, in the sense that it can become necessary even without an explicit C compiler. In other words, we'd have to cover cases where a recipe with only `{{ compiler("cxx") }}` or `{{ compiler("rust") }}` also includes `c_stdlib{,_version}`.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+cdt stands for "core dependency tree" packages. These are repackaged from Centos.
+
+The function expands to the following: 
+
+- package-name-<cdt_name>-<cdt_arch>
+
+Where `cdt_name` and `cdt_arch` are loaded from the variant config. If they are undefined, they default to:
+
+- `cos6` for `cdt_name` on `x86_64` and `x86`, otherwise `cos7`
```
JeanChristopheMorinPerso, line 48: I feel a little bit uneasy with having default values hardcoded here. Do we really need to define default values in the spec? I feel like this kind of thing should be explicitly defined in an ecosystem (the central variant config in an ecosystem) instead of relying on some defaults.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+Pin compatible will pin the dependency to the same version as "previously" resolved in the `host` or `build` environment. This is useful to ensure that the same package is used at run time as was used at build time.
+
+Example:
+
+```yaml
+requirements:
+  host:
+    - numpy
+  run:
+    - ${{ pin_compatible('numpy', exact=True) }}
```
JeanChristopheMorinPerso, line 81: Does it support `min_pin` and `max_pin`? It's unclear right now in this spec.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+outputs:
+  - package:
+      name: libfoo
+      version: "1.2.3"
+  - package:
+      name: foo
+      version: "1.2.3"
+    requirements:
+      run:
+        - ${{ pin_subpackage('libfoo', exact=True) }}
```
JeanChristopheMorinPerso, line 101: Does it support `min_pin` and `max_pin`? It's unclear right now in this spec.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+```
+
+## The `cmp` function
+
+The `cmp` function is used to compare two versions. It returns `True` if the comparison is true and `False` otherwise.
```
JeanChristopheMorinPerso, line 106: Should we add an example of how it can be used?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+## The `env` object
+
+You can use the `env` object to retrieve environment variables and forward them to your build script. There are two ways to do this:
+
+- `${{ env.get("MY_ENV_VAR") }}` will return the value of the environment variable `MY_ENV_VAR` or throw an error if it is not set.
+- `${{ env.get_default("MY_ENV_VAR", "default_value") }}` will return the value of the environment variable `MY_ENV_VAR` or `"default_value"` if it is not set.
+
+You can also check for the existence of an environment variable:
+
+- `${{ env.exists("MY_ENV_VAR") }}` will return `true` if the environment variable `MY_ENV_VAR` is set and `false` otherwise.
```
JeanChristopheMorinPerso, line 125: Nitpick. `false` and `False` are both used in the spec. It's hard to know if it returns a string or a bool.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+- To the `platform::arch` for `cdt_arch`, except for `x86` where it defaults to `i686`.
+
+## The `pin_subpackage` function
+
+### Pin definition
+
+A pin has the following arguments:
+
+- `min_pin`: The lower bound of the dependency spec. This is expressed as a `x.x....` version where the `x` are filled in from the corresponding version of the package. For example, `x.x` would be `1.2` for a package version `1.2.3`. The resulting pin spec would look like `>=1.2` for the `min_pin` argument of `x.x` and a version of `1.2.3` for the package.
+- `max_pin`: This defines the upper bound and follows the same `x.x` semantics but adds `+1` to the last segment. For example, `x.x` would be `1.(2 + 1)` for a package version `1.2.3`. The resulting pin spec would look like `<1.3` for the `max_pin` argument of `x.x` and a version of `1.2.3` for the package.
```
chenghlee, line 58: There seems a strong but unstated assumption here that version strings are integers separated by dots (`.`). If that is the case, that assumption should be explicitly stated and guidance provided as to what should happen if the assumption is violated. (Just sayin', because `1!24.4.0+91_gbfa8ccdce` is a perfectly valid PEP-440/conda package version string.)

chenghlee: Also, as written, it seems like `min_pin` and `max_pin` do not support an explicit version; e.g., recipes cannot simply have `max_pin="1.2.3"`.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+        - ${{ pin_subpackage('libfoo', exact=True) }}
+```
+
+## The `cmp` function
+
+The `cmp` function is used to compare two versions. It returns `True` if the comparison is true and `False` otherwise.
+
+## The `hash` variable
+
+`${{ hash }}` is the variant hash and is useful in the build string computation. This used to be `PKG_HASH` in the old recipe format. Since the `hash` variable depends on the variant computation, it is only available in the `build.string` field and is computed after the entire variant computation is finished.
```
chenghlee, line 110: Should we explicitly say that this is a read-only variable? E.g., is providing "hash: deadbeef123455" in the variant config file explicitly prohibited?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+## The `version_to_buildstring` function
+
+- `${{ python | version_to_buildstring }}` converts a version from the variant to a build string (it removes the `.` character and takes only the first two elements of the version).
```
chenghlee, line 114: Should we provide an example here?

chenghlee (changes requested):

cep-recipe-jinja.md
h-vetinari: > Also, as written, it seems like `min_pin` and `max_pin` do not support an explicit version; e.g., recipes cannot simply have `max_pin="1.2.3"`.

Conda's existing `pin_compatible` jinja [has](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#extra-jinja2-functions) _both_ `{min,max}_pin` as well as `{lower,upper}_bound`, in order to be able to distinguish `'x.x'` from an explicit version limit.

Perhaps it's easier to understand to overload both styles into two kwargs (one for lower, one for upper), and allow either `'x(.x)*'` or an explicit version number? It's more work to implement, but less confusing IMO (otherwise it's easy to specify conflicting things e.g. by setting both `max_pin` and `upper_bound`)

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
+# Jinja functions in recipes
+
+The new recipe format has some Jinja functionalities. We want to specify what functions exist and their expected behavior.
```
jaimergp, line 3: I feel this is lacking a short intro. At least we should link into the previous CEPs.

```suggestion
The new recipe format (introduced by CEP-XX, CEP-XY) has some Jinja functionalities. We want to specify what functions exist and their expected behavior.
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+## The compiler function
+
+The compiler function is used to stick together a compiler from {lang}_compiler and {lang}_compiler_version
```
jaimergp, line 7:
```suggestion
The compiler function is used to stick together a compiler from `{lang}_compiler` and `{lang}_compiler_version`
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+The function looks as follows:
+
+```yaml
+${{ compiler('c') }}
+```
+
+This would pull in the c_compiler and c_compiler_version from the variant config. The compiler function also adds the target_platform to render to something such as:
```
jaimergp, line 15:
```suggestion
This would pull in the `c_compiler` and `c_compiler_version` values from the variant config. The compiler function also adds the `target_platform` value to render to something such as:
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+gcc_linux-64 8.9
+clang_osx-arm64 12
+msvc_win-64 19.29
+```
+
+The function thus evaluates to `{compiler}_{target_platform} {compiler_version}`.
+
+The variant config could look something like:
+
+```
```
jaimergp, line 27:
```suggestion
```yaml
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+gcc_linux-64 8.9
+clang_osx-arm64 12
+msvc_win-64 19.29
+```
+
+The function thus evaluates to `{compiler}_{target_platform} {compiler_version}`.
+
+The variant config could look something like:
```
jaimergp, line 25: I'm not sure "could" is the right verb here. Are we suggesting? Standardizing? If it's not set in stone, why is it useful in this doc? As Cheng proposes, we should standardize the configuration file here.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+  - "8.9"
+cxx_compiler:
+  - clang
+cxx_compiler_version:
+  - "12"
+```
+
+## The `cdt` function
+
+cdt stands for "core dependency tree" packages. These are repackaged from Centos.
```
jaimergp, line 40:
```suggestion
CDT stands for "core dependency tree" packages. These are repackaged from Centos.
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+## The `pin_subpackage` function
+
+### Pin definition
+
+A pin has the following arguments:
+
+- `min_pin`: The lower bound of the dependency spec. This is expressed as a `x.x....` version where the `x` are filled in from the corresponding version of the package. For example, `x.x` would be `1.2` for a package version `1.2.3`. The resulting pin spec would look like `>=1.2` for the `min_pin` argument of `x.x` and a version of `1.2.3` for the package.
+- `max_pin`: This defines the upper bound and follows the same `x.x` semantics but adds `+1` to the last segment. For example, `x.x` would be `1.(2 + 1)` for a package version `1.2.3`. The resulting pin spec would look like `<1.3` for the `max_pin` argument of `x.x` and a version of `1.2.3` for the package.
+- `exact`: This is a boolean that specifies whether the pin should be exact. It defaults to `False`. If `exact` is `True`, the `min_pin` and `max_pin` are irrelevant. We pin to an `==` version and also include the build string exactly (e.g. `==1.2.3=h1234`).
```
jaimergp, line 59:
```suggestion
- `exact`: This is a boolean that specifies whether the pin should be exact. It defaults to `False`. If `exact` is `True`, the `min_pin` and `max_pin` are irrelevant. We pin to a `==` version and also include the build string exactly (e.g. `==1.2.3=h1234`).
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+
+#### Example
+
+If we consider a package like `numpy-1.21.3-h123456_5` we could apply some pin expressions. 
+
+- `min_pin=x.x, max_pin=x.x` would result in `>=1.21,<1.22`
+- `min_pin=x.x.x,max_pin=x` would result in `>=1.21.3,<2`
```
jaimergp, line 66:
```suggestion
- `min_pin=x.x.x, max_pin=x` would result in `>=1.21.3,<2`
```

cep-recipe-jinja.md
jaimergp: Yes, what operators are available and what is the syntax?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+      version: "1.2.3"
+  - package:
+      name: foo
+      version: "1.2.3"
+    requirements:
+      run:
+        - ${{ pin_subpackage('libfoo', exact=True) }}
+```
+
+## The `cmp` function
```
jaimergp, line 104:
```suggestion
## The `cmp` function
```

I prefer a more explicit name like `compare_versions`.

cep-recipe-jinja.md
jaimergp: Shall we call it `variant_hash`?

cep-recipe-jinja.md
jaimergp: This is the first time we introduce the `|` syntax as well. Can we do `version_to_buildstring(python)`?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+## The `version_to_buildstring` function
+
+- `${{ python | version_to_buildstring }}` converts a version from the variant to a build string (it removes the `.` character and takes only the first two elements of the version).
+
+## The `env` object
+
+You can use the `env` object to retrieve environment variables and forward them to your build script. There are two ways to do this:
+
+- `${{ env.get("MY_ENV_VAR") }}` will return the value of the environment variable `MY_ENV_VAR` or throw an error if it is not set.
+- `${{ env.get_default("MY_ENV_VAR", "default_value") }}` will return the value of the environment variable `MY_ENV_VAR` or `"default_value"` if it is not set.
```
jaimergp, line 121: Could this be overloaded to a single function? With the 2nd param being optional.

cep-recipe-jinja.md
jaimergp, line 1: I'm missing a section on error handling:

- What should happen when Jinja functions fail to process the input?
- Should they render to null or just the input?
- Should they raise the error?

Not sure if this should be a general section or a subsection in each function.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
...
+```
+c_compiler:
+  - gcc
+c_compiler_version:
+  - "8.9"
+cxx_compiler:
+  - clang
+cxx_compiler_version:
+  - "12"
+```
```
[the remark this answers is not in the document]
wolfv: yeah we need to write a spec for the variant config (or even more global "config"). But IMO that shouldn't be part of this CEP.

cep-recipe-jinja.md
wolfv: Yeah using a function would also work. But we also expose the default minijinja filters (such as `lower`, `replace`, ...), that can be used like `${{ "mystring" | upper }}`

or `${{ val | replace('foo', 'moo') }}`

cep-recipe-jinja.md
```diff
@@ -0,0 +1,125 @@
+# Jinja functions in recipes
+
+The new recipe format has some Jinja functionalities. We want to specify what functions exist and their expected behavior.
+
+## The compiler function
```
h-vetinari, line 5: Since the discussion on stdlib further upthread died down, just leaving a note that this either needs to add the `stdlib` function, or propose an alternative way of how to realize the same functionality (and how it gets populated through the variant config).

h-vetinari (changes requested):

cep-recipe-jinja.md
wolfv: Sure, we can get rid of this. I think we adopted this from `conda-build` where these values are also there by default.

cep-recipe-jinja.md
wolfv: Wow, I wasn't aware! I found a few places where it's used but I don't know how it works. For example, in the `onnx-feedstock` it says:

```
    - {{ pin_compatible('numpy', lower_bound='1.19', upper_bound='3.0') }}  # [py>38]
```

Does this still take the lower bound from the build-time resolution? Or is this equivalent to just writing `numpy >=1.19,<3.0`?

wolfv: I also don't really understand why we would want this in the function. One can just as well write something like:

```
- numpy >=1.5
- ${{ pin_compatible("numpy", max_pin="x.x", min_pin="x.x.x") }}
```

If one needs additional (hard-coded) constraints. The solver doesn't really care about more constraints.

cep-recipe-jinja.md
wolfv: I added the word `boolean true` to make it clear that this is a bool value.

cep-recipe-jinja.md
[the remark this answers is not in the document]
wolfv: looks OK to me? :)

cep-recipe-jinja.md
wolfv: @jaimergp I added some more text regarding jinja filters to the bottom of the CEP.

cep-recipe-jinja.md
h-vetinari: > I also don't really understand why we would want this in the function.

I wanted this for numpy2, but that's a rare enough case that it probably doesn't matter. In that case, we're building against 2.0, but I want the lower bound to be 1.19, which cannot be done with what you propose (because it's _looser_ than what `pin_compatible` enforces).

More generally though, in your example, I find that adding additional constraints just feels like a weaksauce work-around. If I'm already using a function to specify the constraint on numpy, why won't it give me the expressivity needed to do that?

> Does this still take the lower bound from the build-time resolution? Or is this equivalent to just writing `numpy >=1.19,<3.0`?

I think the intention of the author was clear (be compatible with numpy 2 on py>38), but it doesn't work because it just creates an _additional_ constraint, where - again - it would need to be looser. IOW, onnx's [metadata](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fonnx-1.16.0-py39h974b0fd_0.conda) contains both:
```
numpy >=1.19,<3.0
numpy >=1.26.4,<2.0a0
```

h-vetinari: > but that's a rare enough case that it probably doesn't matter

On second thought, probably it does. That onnx recipe is precisely trying to work around the same thing that I need to handle in the numpy run-export - in other words, building against numpy >=1.25 (with no extra manual overrides) gives a run-time lower bound of `>= 1.19` currently.

h-vetinari: > but it doesn't work because it just creates an _additional_ constraint, where - again - it would need to be looser.

Ah wait, maybe it actually does work correctly, and the package just falls prey to the numpy run-export, which adds the `numpy >=1.26.4,<2.0a0` on top`.

wolfv:
@h-vetinari since we don't have a CEP for stdlib it might be good to define it here. 

What are the default values for the different operating systems?

h-vetinari:
> since we don't have a CEP for stdlib it might be good to define it here.

Sounds good!

> What are the default values for the different operating systems?

I was under the impression that defaults here aren't necessarily a good thing? Perhaps even more than the compiler packages, the stdlib-metapackages don't have a "canonical" naming. Are you planning to set defaults for `{{ compiler(...) }}` as well? Looking at the diff here it doesn't seem like there are default being proposed?

In any case, here are the specs that we're using in conda-forge. They won't change substantially anymore, barring unforeseen events.
 
```
c_stdlib:
  - sysroot                    # [linux]
  - macosx_deployment_target   # [osx]
  - vs                         # [win]
m2w64_c_stdlib:                # [win]
  - m2w64-toolchain            # [win]              <-- might change to m2w64-sysroot in the future
c_stdlib_version:              # [unix]
  - 2.17                       # [linux]            <-- per July 2024
  - 10.13                      # [osx and x86_64]
  - 11.0                       # [osx and arm64]
```

wolfv:
I expanded the content a bit. 

Some things that are left to decide:

- Limit the filters from Jinja to a default subset?
- Should `env` work more like Python `env`
  - `env["FOO"]` instead of `env.get("FOO")`
  - `env.get("FOO", "default")` instead of `env.get_default("FOO", "default")`
- Or should it be more like Github Actions
  - `env.FOO` instead of `env.get("FOO")`
  - `env.FOO or "default"` for default values
- Should `${{ compiler('c') }}` evaluate directly to `gcc_linux-64` or should there be an intermediate representation (as is now)
- For `pin_subpackage` it will be impossible to evaluate it directly (due to self-referential nature of the function).
- Should we add `lower_bound` and `upper_bound` to the pin functions? And how should they work (in conjunction with `min_pin` and `max_pin`)?

schuylermartin45:
> * Limit the filters from Jinja to a default subset?
> * Should we add a `split(str)` filter (useful for version numbers, so that one can do `"${{ version | split(".") | slice(0, 2) | join(".") }}` or something along those lines. We could also create a more specialized function/filter for version string manipulation of this sort.
> * Should `env` work more like Python `env`
>   
>   * `env["FOO"]` instead of `env.get("FOO")`
>   * `env.get("FOO", "default")` instead of `env.get_default("FOO", "default")`
> * Or should it be more like Github Actions
>   
>   * `env.FOO` instead of `env.get("FOO")`
>   * `env.FOO or "default"` for default values
I can't speak as much for the needs of package builders, but I will say I think the  any allowed JINJA syntax needs to minimized. It adds a significant level of complication in the development of automated tooling (namely package editing). But if we must have them, we should limit the syntax and make the rules of using the features very clear. The more variety available, the more developers will use the tools in "clever" ways, the more breakages we will have when trying to automatically edit recipes.

cep-recipe-jinja.md
wolfv: Added some examples!

cep-recipe-jinja.md
wolfv: They both are derived from the same "pin" function. So yeah, they both support the same inputs.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+>  fortran: gfortran
+>  rust: rust
+> osx:
+>   c: clang
+>   cxx: clangxx
+>   fortran: gfortran
+>   rust: rust
+> win:
+>   c: vs2017
+>   cxx: vs2017
```
h-vetinari, line 130: It might be nicer to use unversioned `vs` here? This is a relatively recent [addition](https://github.com/conda-forge/vc-feedstock/blob/ef25bb03d65e5c7adc024c07cbb2ccbe46b562bd/recipe/meta.yaml#L223), but that way the version would be independent of the default name, like on unix.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+>   fortran: gfortran
+>   rust: rust
+> ```
+
+## The `stdlib` function
+
+The `stdlib` function works exactly as the `compiler` function.
```
h-vetinari, line 137: Not sure if that's the intention here, but perhaps a bit of clarification nevertheless.
```suggestion
The `stdlib` function works exactly as the `compiler` function, except that it has not pre-defined defaults.
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+cxx_compiler:
+  - clang
+cxx_compiler_version:
+  - "12"
+```
+
+> [!NOTE]
+> Default values for the `compiler` function
+>
+> If not further specified, the following values are used as default values:4
```
jaimergp, line 115:
```suggestion
> If not further specified, the following values are used as default values:
```
Typo?

cep-recipe-jinja.md
wolfv: Yeah this section is not complete yet. I think we should define defaults for stdlib? We need to make the experience of conda recipes better...

cep-recipe-jinja.md
wolfv: We are inheriting the standards from `conda-build` here. I am happy to change them, and using an non-versioned `vs` would probably be better (altough also potentially annoying when it moves since you have to re-install vs on the build machine).

wolfv: And lastly, not sure what Anaconda has available on the defaults channel ...

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+- `min_pin`: The lower bound of the dependency spec. This is expressed as a `x.x....` version where the `x` are filled in from the corresponding version of the package. For example, `x.x` would be `1.2` for a package version `1.2.3`. The resulting pin spec would look like `>=1.2` for the `min_pin` argument of `x.x` and a version of `1.2.3` for the package.
+- `max_pin`: This defines the upper bound and follows the same `x.x` semantics but adds `+1` to the last segment. For example, `x.x` would be `1.(2 + 1)` for a package version `1.2.3`. The resulting pin spec would look like `<1.3` for the `max_pin` argument of `x.x` and a version of `1.2.3` for the package.
+- `exact`: This is a boolean that specifies whether the pin should be exact. It defaults to `False`. If `exact` is `True`, the `min_pin` and `max_pin` are irrelevant. We pin to a `==` version and also include the build string exactly (e.g. `==1.2.3=h1234`).
+
+#### Example
+
+If we consider a package like `numpy-1.21.3-h123456_5` we could apply some pin expressions.
+
+- `min_pin=x.x, max_pin=x.x` would result in `>=1.21,<1.22`
+- `min_pin=x.x.x, max_pin=x` would result in `>=1.21.3,<2`
```
jaimergp, line 174: Should the resolution of `max_pin` add the `.dev0` component too to avoid the dev versions leaking in in the version constraint?

Current behavior in solvers is that `<2` actually includes things like `v2.0b1`.

cep-recipe-jinja.md
jaimergp: Ah so it's more like a version matcher. When I read `compare` I thought we were comparing two packages together like "is numpy's version greater than python's", which didn't make much sense to me.

But this is more like asking "Does the version of Python match this version expression?". So in that sense the name could be better expressed as:

- `version(python, "<3.8")`
- `version_match(python, "<3.8")`
- `match_version(python, "<3.8")`
- `satisfy(python, "<3.8")`

Also this assumes that `python` will be represented by a concrete version `x.y.z` and not a spec like `x.y.*` right? IOW, it's already resolved to a specific package record.

If that's the case, should we specify that this function can only return after the solve has completed?

cep-recipe-jinja.md
h-vetinari: > altough also potentially annoying when it moves since you have to re-install vs on the build machine

Generally, our vs packages in conda-forge will pick up the compilers that you have - we have a fallback to use the local version (& toolchain target) if our default activation fails. Also, I'm not saying to abolish the `2017 / 2019 / 2022` pins, just move them to `c{,xx}_compiler_version`, like for unix.

wolfv: Nice! Didn't know that.

wolfv: cc @jezdez @chenghlee what is the situation on Anaconda's channels here? Should we make `vs` the default?

cep-recipe-jinja.md
h-vetinari: > Yeah this section is not complete yet. 

Ah, that wasn't clear to me. I'm ambivalent about the usefulness of defaults, but if your goal is that people are able to write recipes outside of cf (where we have the defaults in the global pinning anyway) & Anaconda main channels _and_ not have to worry about where their compiler is coming from, then I guess that's a usecase.

cep-recipe-jinja.md
wolfv: Do you know what `conda-build` does?

I think the pre-release discussion is another one (maybe?). I would _really_ like to have some behavior like on PyPI where you can opt-in to pre-releases. But I think that's for another CEP.

Just for reference, I didn't try to invent something new with these pin expressions but rather just copy whatever `conda-build` does.

h-vetinari: I think that would be a good idea from the POV of least surprise. It happened in the numpy 2 migration that people were surprised that  `2.0.0rc2` actually satisfies `<2` (myself included).

cep-recipe-jinja.md
h-vetinari: FWIW, in the new CDTs for Alma 8, we're planning to get rid of `cdt_name`.

cep-recipe-jinja.md
wolfv: Care to write a CEP?! :) Would love to introduce that behavior.

jaimergp: > Do you know what conda-build does?

Apparently it does handle this case. See [`gau2grid` recipe](https://github.com/conda-forge/gau2grid-feedstock/blob/32987c92c3933afe3da0dd21fec166599212d422/recipe/meta.yaml#L16) + its [rendered run_exports including the `a0` bit](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fgau2grid-2.0.7-h0b41bf4_2.conda).

The code in conda-build has it [here](https://github.com/conda/conda-build/blob/9c0ceec8d96d265e33d960c46024db78f4e3efd8/conda_build/utils.py#L1514). That should better be a `0dev0`, though, since that's technically the lowest bound (dev < alpha).

cep-recipe-jinja.md
h-vetinari: @wolfv @chenghlee WDYT about the idea of having pin-or-version arguments in `pin_compatible`? As an example, let's say for numpy 2.0.1:
``` 
pin_compatible("numpy")                                         --> numpy >=2.0.1,<3    # default: lower_bound='x.x.x.x.x.x', upper_bound='x'
pin_compatible("numpy", upper_bound="x.x")                      --> numpy >=2.0.1,<2.1
pin_compatible("numpy", lower_bound="x.x", upper_bound="2.3")   --> numpy >=2.0,<2.3
pin_compatible("numpy", lower_bound="1.19")                     --> numpy >=1.19,<3     # most libs compiled against np2
pin_compatible("numpy", lower_bound="1.19", upper_bound="2.3")  --> numpy >=1.19,<2.3   # scipy's usecase
```

The API contract could be that anything matching `x(\.x)*` is a pinning expression, and the rest gets treated as an explicit version.

cep-recipe-jinja.md
JeanChristopheMorinPerso: I would avoid defaults. I think default values in tools are counter intuitive and should be defined on a per ecosystem basis and not by tools. In the sense that conda-forge should define their default and Anaconda should define their defaults, same for all the other ecosystems.

JeanChristopheMorinPerso: To clarify my thoughts: Defaults like this are (or can) help less experienced package builders, but it can also be very disturbing and annoying for more experienced package builders. It can also cause troubles for ecosystems that might want to opt-out of these defaults (for example because they want to be very explicit about what they support and how things should work).

I sympathize to the cause of making it easier to use the build tools, but there are things which are very ecosystem specific. I believe compilers is one of these things.

cep-recipe-jinja.md
JeanChristopheMorinPerso: >I think I would like to have an awesome "default" experience. If you take a recipe from conda-forge, ideally you can just run build on it without having to worry about the variant config file.

The thing is that this is very ecosystem specific. The default might work for some time if conda-forge uses the same defaults, but as soon as an ecosystem like conda-forge diverges from the CEP defaults, then the recipes might not be buildable without a variant config. Same for the defaults channel.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+>   rust: rust
+> ```
+
+## The `stdlib` function
+
+The `stdlib` function works exactly as the `compiler` function.
+
+## The `cdt` function
+
+CDT stands for "core dependency tree" packages. These are repackaged from Centos.
```
JeanChristopheMorinPerso, line 141: My comment disappeared, so re-adding it:

Does "centos" have to be mentioned here? CDT packages could technically contain binaries from any Linux distribution right? I'm also wondering if we should add a link to some documentation that describe CDT more in details?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+- `pin_compatible`
+- `pin_subpackage`
+
+Both follow the same "pinning" mechanism as described next and have the same arguments.
+
+### Pin definition
+
+A pin has the following arguments:
+
+- `min_pin`: The lower bound of the dependency spec. This is expressed as a `x.x....` version where the `x` are filled in from the corresponding version of the package. For example, `x.x` would be `1.2` for a package version `1.2.3`. The resulting pin spec would look like `>=1.2` for the `min_pin` argument of `x.x` and a version of `1.2.3` for the package.
```
JeanChristopheMorinPerso, line 165:
```suggestion
- The package name
- `min_pin`: The lower bound of the dependency spec. This is expressed as a `x.x....` version where the `x` are filled in from the corresponding version of the package. For example, `x.x` would be `1.2` for a package version `1.2.3`. The resulting pin spec would look like `>=1.2` for the `min_pin` argument of `x.x` and a version of `1.2.3` for the package.
```
?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+
+Would evaluate to a `abc123_cuda112` (assuming the hash was `abc123`).
+
+### Various remarks
+
+#### Inline conditionals with Jinja
+
+The new recipe format allows for inline conditionals with Jinja. If they are falsey, and no `else` branch exists, they will render to an empty string (which is, for example in a list or dictionary, equivalent to a YAML `null`).
+
+When we render the recipe, we filter all values that are `null` from the YAML output. This allows us to write and parse the following easily:
```
JeanChristopheMorinPerso, line 289: Instead of saying "we". should it say "the build tool must"? I've seen "we" used in other places in this CEP too. ALso, same question for "us".

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+
+```yaml
+requirements:
+  host:
+    - ${{ "numpy" if cuda == "yes" }}
+```
+
+If `cuda` is not equal to yes, the first item of the host requirements will be empty (null) and thus filtered from the final list.
+
+This also works for dictionary keys:
```
JeanChristopheMorinPerso, line 299:
```suggestion
This also works for dictionary values:
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,320 @@
...
+### Things to decide
+
+- Limit the filters from Jinja to a default subset?
+- Should `env` work more like Python `env`
+  - `env["FOO"]` instead of `env.get("FOO")`
+  - `env.get("FOO", "default")` instead of `env.get_default("FOO", "default")`
+- Or should it be more like Github Actions
+  - `env.FOO` instead of `env.get("FOO")`
+  - `env.FOO or "default"` for default values
+- Should `${{ compiler('c') }}` evaluate directly to `gcc_linux-64` or should there be an intermediate representation (as is now)
```
JeanChristopheMorinPerso, line 319: Isn't this an implementation detail of the tool? In the sense that even if you have an intermediate representation, the end result will be `gcc_linux-64 11.1.0` right?

cep-recipe-jinja.md
wolfv: The only way around this is that we make an endpoint in the repodata or so where we can easily download the `conda_build_config.yaml` for a specified channel to get these defaults per-channel. 

If we don't have any easy way it means that each user needs _at least_ two files per recipe which is already a bit annoying, no?

cep-recipe-jinja.md
wolfv: I changed the text, and refrain from specifying _how_ the values are configured in the `variant`.

cep-recipe-jinja.md
wolfv: I am wondering some more abotu the "no defaults". E.g. if I take a feedstock, and want to add support for macOS - it's pretty annoying because I don't have a .ci_config for osx-arm64 yet. It would be most convenient to either use a default. It would be more correct to point to the `conda-forge-pinning-feedstock`. 

I think ideally we publish the pinnings next to the repodata file or something like that.

cep-recipe-jinja.md
wolfv: It's the current default value in `conda-build`. idk what to do, i am basically just documenting what's already in conda-build.

cep-recipe-jinja.md
wolfv: Yes, it does. Both use the same pin specification as mentioned earlier.

cep-recipe-jinja.md
wolfv: Agreed!

cep-recipe-jinja.md
h-vetinari: @wolfv, given your changes in 0fda2133b485275ae0ef59945e22afd61473937e, I presume you're not on board with this idea (or didn't see this thread)?

If you want to keep `{min,max}_pin` separate from `{lower,upper}_bound`, I think we need to enforce that at most one `{min_pin,lower_bound}` resp. `{max_pin,upper_bound}` each may be specified, everything else should raise an error.

I'd still prefer a unified interface, because I don't see the gain to keep these arguments separate (but rather requires a lot of complexity to ensure it is unambiguous/consistent)

wolfv: I indeed didn't see it or didn't read carefully enough. I also think this is a good change. The logic is already in place, so I don't think it will be a big change on the rattler-build side.

cep-recipe-jinja.md
wolfv: it's not really prohibited in the implementation right now. But we can say that it should be...

I don't have strong feelings about the name. We can change it to `variant_hash` if that's preferred!

cep-recipe-jinja.md
wolfv: yeah, that can work. We can also make it a named argument, e.g. 

```
env.get("FOO", default="baz")
```

What do you think?

cep-recipe-jinja.md
wolfv: Thanks, I just added this!

cep-recipe-jinja.md
wolfv: Yep, it's just something on the tool side. It's gone now.

cep-recipe-jinja.md
wolfv: @jaimergp I implemented this change: https://github.com/prefix-dev/rattler-build/pull/917

Let me know what you think.

cep-recipe-jinja.md
wolfv: I added something about aggressive error throwing at the bottom of the CEP.

cep-recipe-jinja.md
h-vetinari: I proposed the API that I think is more natural and IMO more desirable long-term.

Obviously the naming choice is just bike shedding and both would work, but `max_pin=1.2.3` seems off to me, whereas `upper_bound='x.x'` is fine because it's a more generic term, where users don't have to guess whar a pin is or isn't.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+
+For example, it can be used in the following way:
+
+```yaml
+requirements:
+  - $ {{ "six" if match(python, "<3.8") }}
+  - $ {{ "six" if match(python, "3.8") }}
+  - $ {{ "six" if match(python, "==3.8") }}
+  - $ {{ "six" if match(python, "3.8.*") }}
+  - $ {{ "six" if match(python, ">=3.8,<3.10") }}
```
mbargull, line 358:
```suggestion
  - ${{ "six" if match(python, "<3.8") }}
  - ${{ "six" if match(python, "3.8") }}
  - ${{ "six" if match(python, "==3.8") }}
  - ${{ "six" if match(python, "3.8.*") }}
  - ${{ "six" if match(python, ">=3.8,<3.10") }}
```
(BTW, thanks for going with the string-based Conda version spec format here!)

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+
+- `${{ env.get("MY_ENV_VAR") }}` will return the value of the environment
+  variable `MY_ENV_VAR` or throw an error if the environment variable is not
+  set.
+- `${{ env.get("MY_ENV_VAR", default="default_value") }}` will return the value
+  of the environment variable `MY_ENV_VAR` or `"default_value"` if it is unset.
+
+You can also check for the existence of an environment variable:
+
+- `${{ env.exists("MY_ENV_VAR") }}` will return a boolean `true` if the
```
mbargull, line 388: Is the `.exists` a standard Jinja method?
(Sorry, I didn't look this up. Just want to make sure we are conscious this in case it's some custom thing vs a plain Python  `"MY_ENV_VAR" in env`.)

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+</details>
+
+
+### Extra filters for recipes
+
+#### The `version_to_buildstring` filter
+
+- `${{ python | version_to_buildstring }}` converts a version from the variant
+  to a build string (it removes the `.` character and takes only the first two
+  elements of the version).
```
mbargull, line 456: I'm not so sure about the utility of this one.
Yes, this how we currently compose parts of build strings.
But it's also a hacky thing to do, really.
This might just support dragging on such build string massaging for longer than we should.

(That is not to say that I'm completely opposed to such functionality; I just want to caution against adding potential future burden from the get go.)

mbargull (changes requested):

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+foo_compiler: "compiler_name"
+foo_compiler_version: "1.2.3"
+
+# on linux-64 this then results in
+# compiler: "compiler_name_linux-64 1.2.3"
+```
+
+> [!NOTE] Default values for the `compiler` function
+>
+> If not further specified, the following values are used as default values:
```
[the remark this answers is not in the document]
mbargull: That's probably resolved by the commit from some minutes ago (haven't yet checked those changes).

cep-recipe-jinja.md
wolfv: Thanks!

cep-recipe-jinja.md
wolfv: No, it's not standard jinja. You're right, we might be able to implemnetn the functions to make `"FOO" in env` work, but I am not sure if its worth it (also seems a bit more magical?).

We also don't use a regular "dict" or "sequence" object to be able to throw proper errors etc.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+The function expands to the following:
+
+- package-name-<cdt_name>-<cdt_arch>
+
+Where `cdt_name` and `cdt_arch` are loaded from the variant config. If they are
+undefined, they default to:
+
+- `cos6` for `cdt_name` on `x86_64` and `x86`, otherwise `cos7`
+- To the `platform::arch` for `cdt_arch`, except for `x86` where it defaults to
+  `i686`.
```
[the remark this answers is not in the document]
wolfv: I removed this stuff

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+build:
+  - ${{ stdlib('c) }}
+```
+
+Evaluates to the `c_stdlib` and `c_stdlib_version` from the variant config
+(incl. the target platform), using the following `{lang}_stdlib` and
+`{lang}_stdlib_version` keys.
+
+The function should evaluate to `{($lang)_stdlib}_{target_platform}
+($lang)_stdlib_version`.
```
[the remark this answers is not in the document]
wolfv: fixed :)

cep-recipe-jinja.md
mbargull: Ah, I also missed that `env.get("MY_ENV_VAR")` should throw an error in case `MY_ENV_VAR` isn't set -- meaning I totally missed that `env` isn't meant to be a `dict`.
TBH, I don't have a strong opinion then (I actually like a method better; no idea what people would prefer syntax- and/or naming-wise).

cep-recipe-jinja.md
mbargull: Yes, it'd be a `python.split(".")[:2] | join("")` -- so, not too complicated, but also error-prone.
My remark was more about why we do these stuff in general, though.
I think turning a `11.2.0` into `112` is just not wise but just a historic habit that we might want to weed out...

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+```yaml
+context:
+  version: "1.0.5"
+
+package:
+  version: ${{ version }}
+```
+
+Context evaluation must happen from top-to-bottom. That means a later value can
+reference an earlier one like so:
```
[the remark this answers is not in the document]
wolfv: I added this text. Rattler build will need the user to use the correct order for the recipe to work but you'll get a nice error :)

cep-recipe-jinja.md
```diff
@@ -0,0 +1,511 @@
...
+```yaml
+context:
+  version: "1.0.5"
+  name_and_version: "pkg_${{ version | replace('.', '_') }}"  # evaluates to "pkg_1_0_5"
+```
+
+### Variables from variant configuration
+
+Any variable specified in the variant configuration file can be used in a recipe
+by using it in a Jinja expression. For example, with a variant config like:
```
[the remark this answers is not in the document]
wolfv: Actually we decided to allow string, integer and boolean values.

cep-recipe-jinja.md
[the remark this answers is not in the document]
mbargull: That's totally fine; meaning, it's then just a deficiency in rattler-build for now that we can resolve later.
The spec-like document only can't say "we need sequential processing of non-sequence data type".

cep-recipe-jinja.md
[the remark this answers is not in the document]
mbargull: Oh... we may want to discuss that (and the format of the variant config file in general).
That's a change that has its pros and cons and we should be careful on how to handle it.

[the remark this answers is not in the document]
wolfv: Yeah I was planning to leave that for another document :)

cep-recipe-jinja.md
wolfv: Actually, this syntax doesn't work in the new recipes (at least not in the way implemneted in `rattler-build` since we're using `minijinja` that doesn't have all the Python builtins implemented).

jezdez:
@conda/steering-council

baszalmstra:
yes

wolfv:
Please use the following form to vote:

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [x]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@kkraus14 (Keith Kraus)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [x]  yes
- [ ]  no
- [ ]  abstain

cep-recipe-jinja.md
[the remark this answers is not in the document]
mariusvniekerk: We could make another CEP also allowing for these ordered context fields to be specified as 

context:
	- key: valueExpr

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+  - "have_cuda"
+  - "no_cuda"
+```
+
+This value can be used as follows, for example in an inline `if` expression:
+
+```yaml
+requirements:
+  host:
+    - ${{ "cudatoolkit" if cuda == "have_cuda" }}
```
mariusvniekerk, line 108: Should we add an explicit callout for the case like this where `expression` can evaluate into `""` or i guess `null` and that means this is excluded.

cep-recipe-jinja.md
mariusvniekerk: using `''` as the join seperator is almost certainly a bad idea even though we do it all the time.  `_` is a FAR better solution since it avoids the `1.12.0` == `11.2.0` problem that this has

cep-recipe-jinja.md
wolfv: We have something about that in line 500 - 510: https://github.com/conda/ceps/pull/71/files/c1028142a7f6e3e8a5d011667e0bec3b3f67ab3a#diff-0154c4169103373897853cfe5287bb7a1101289dd78c7fa30d2acba92b644938R500-R510

cep-recipe-jinja.md
JeanChristopheMorinPerso: So what should we do with default values? I continue to think that default values for compilers and compiler versions should not be part of the CEP.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+```yaml
+build:
+  number: ${{ 100 if cuda == "yes" }}
+  # or an `else` branch can be used, of course
+  number: ${{ 100 if cuda == "yes" else 0 }}
+```
+
+## Error handling
+
+Build tools should be aggressive about Jinja errors:
```
JeanChristopheMorinPerso, line 523: Is "should" defined somewhere in the CEP repo? If not, an implementer could interpret it as optional compared to "must".

mariusvniekerk (approved):

wolfv:
@marcelotrevisani @jakirkham @mbargull @CJ-Wright last chance to vote - you can also choose `abstain` which would still help with the quorum.

schuylermartin45:
NP on the title of this PR: Since this is `schema_version: 1` I've been referring to this as the V1 recipe format and the pre-CEP-13 format as V0. Do we want to keep consistent with that nomenclature?

I'm also not sure if it is worth codifying somewhere to prevent confusion. Maybe it is also worth it to add a section to the top of the CEP that specifies which `schema_version` a CEP belongs to, to prevent future confusion?

chenghlee:
For the record: voted "no" because while I like the ideas presented in this CEP, I don't think the CEP as written is ready to be adopted _as a specification_. IMO, various unanswered questions/comments need to be addressed before adoption (e.g., top-down evaluation of the context object).

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+```
+
+### Default variables in the recipe
+
+Several variables are globally available in the recipe, based on the `target_platform` and `build_platform`:
+
+- `target_platform` is a string that represents the platform for which the package is built. It is a string of the form `os-arch` (e.g. `linux-64`, `osx-64`, `win-64`, `linux-aarch64`, ...).
+- `build_platform` is a string that represents the platform on which the package is built. Same format as `target_platform`.
+- `linux`, `osx`, `win`, `emscripten`: These are boolean variables that are `true` if the target platform is a Linux, macOS, Unix, or Windows platform, respectively. Note that this is the first part of the `target_platform` string.
+- `x86_64`, `aarch64`, `armv7l`, `ppc64le`, `s390x`, `sparc64`, `riscv64`: These are boolean variables that are `true` if the target platform is the respective architecture. Note that, except for `x86_64`, these are the second part of the `target_platform` string.
```
mbargull, line 118: This does not capture the `osx-arm64`/`win-arm64` cases.
Not a huge deal, but it would be good to mention here how those are handled.

cep-recipe-jinja.md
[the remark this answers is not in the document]
mbargull: > We could make another CEP also allowing for these ordered context fields to be specified as

FYI, this had also been suggested in https://github.com/conda/ceps/pull/56#discussion_r1417068195 .

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+```
+
+Context evaluation must happen from top-to-bottom. That means a later value can
+reference an earlier one like so:
+
+```yaml
+context:
+  version: "1.0.5"
+  name_and_version: "pkg_${{ version | replace('.', '_') }}"  # evaluates to "pkg_1_0_5"
+```
```
mbargull, line 90:
```suggestion
```
It would be better to leave this under-specified here instead of specifying it in conflict with the CEP-14.
This exact issue with the specification has been brought up by me in both previous CEPs for the new format and the wording here regresses on the issue instead of finally addressing it.
I do not understand why this simple issue is not being addressed, given that solutions have already been proposed.

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+### The `stdlib` function
+
+The `stdlib` function works exactly as the `compiler` function, but uses the
+`stdlib` keys in the variant.
+
+For example:
+
+```yaml
+build:
+  - ${{ stdlib('c) }}
```
jakirkham, line 186:
```suggestion
  - ${{ stdlib('c') }}
```

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+
+#### Pin definition
+
+A pin has the following arguments:
+
+- `package_name`, positional, required: The name of the package to pin.
+- `lower_bound`, defaults to `x.x.x.x.x.x`: the lower bound, either as a version
+  or as a "pin expression", or `None`
+- `upper_bound`, defaults to `x`: the upper bound, either as a version or as a
+  "pin expression" or `None`
```
jakirkham, line 227: This is specifying bounds...

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+- If the last segment is a letter, the number should be incremented and the
+  letter set to `a`, e.g. `9d` with a `x` pin expression results in `<10a`.
+- If the last segment is a number, the number should be incremented and `.0a0`
+  should be appended to prevent any alpha versions from being selected. For
+  example: `1.2.3` with a `x.x` pin expression should result in `<1.3.0a0`.
+- The epoch is left untouched by the `max_pin` (or `min_pin`). If the epoch is
+  set, it will be included in the final version. E.g. `1!1.2.3` with a
+  `max_pin='x.x'` will result in `<1!1.3.0a0`.
+- When bumping the version with a `max_pin` the local version part is removed.
+  For example, `1.2.3+local` with a `max_pin='x.x'` will result in `<1.3.0a0`.
```
jakirkham, line 262: Whereas this is specifying pins. Should we consolidate to one terminology or should we define both? Are there cases we need both?

cep-recipe-jinja.md
```diff
@@ -0,0 +1,527 @@
...
+
+> [!NOTE]
+>
+> `conda-build` uses the `lower_bound` for the version that is used in
+> the `max_pin` pinning expression. `conda-build` also ignores the `min_pin`
+> expression when a `upper_bound` is used.
```
jakirkham, line 268: Am curious about the motivation for the differences here

FWIW do find users confuse or forget about the options (usually the bounds). So consolidating somehow would makes sense

cep-recipe-jinja.md
jakirkham: `get` just returns `None` for undefined keys in Python and Jinja

https://docs.python.org/3/library/stdtypes.html#dict.get

cep-recipe-jinja.md
jakirkham: Ideally we determine these in a provisional CEP that others use

jakirkham (approved):
Thanks Wolf! 🙏

Generally seems reasonable. Included some comments/suggestions. Otherwise would be good to add

wolfv:
The vote has been closed with the following result:

Total voters: 15 (valid: 13 = 86.67%)

Yes votes (11 / 84.62%):
- @xhochy (Uwe Korn)
- @goanpeca (Gonzalo Peña-Castellanos)
- @ocefpaf (Filipe Fernandes)
- @msarahan (Michael Sarahan)
- @jakirkham (John Kirkham)
- @jezdez (Jannis Leidel)
- @wolfv (Wolf Vollprecht)
- @kkraus14 (Keith Kraus)
- @baszalmstra (Bas Zalmstra)

No votes (1 / 7.69%)):
- @chenghlee (Cheng H. Lee)

Abstain votes (1 / 7.69%):
- @mbargull (Marcel Bargull)

Not voted (2):

Invalid votes (0):

We reached quorum, and enough YES votes for this CEP to be accepted. 🎉

cep-recipe-jinja.md
wolfv: added `arm64`.

cep-17.md
```diff
@@ -0,0 +1,527 @@
...
+- [CEP 14: A new recipe format: allowed keys and values](https://github.com/conda/ceps/blob/main/cep-14.md)
+
+Historically, `conda-build` recipes have relied on templating with `Jinja` for some
+"dynamic" functionality. For example, many recipes use the `version` of the
+package in multiple places (as package version, in the URL and the tests, for
+example). To make it easy to change recipes, Jinja has been used for some
+light-weight templating.
+
+The "old" recipe format has allowed arbitrary Jinja syntax (including set,
+if/else or for loops). The new recipe format only allows a subset of Jinja with
```
beeankha, line 27:
```suggestion
The "old" recipe format has allowed arbitrary Jinja syntax (including `set`,
`if`/`else` or `for` loops). The new recipe format only allows a subset of Jinja with
```

schuylermartin45: We should probably also change `old` to `V0` while we're at it.

cep-17.md
```diff
@@ -0,0 +1,527 @@
...
+requirements:
+  - ${{ "six" if is_unix(target_platform) }}
+  - ${{ "six" if is_win(target_platform) }}
+  - ${{ "six" if is_linux(build_platform) }}
+```
+
+## The `hash` variable
+
+`${{ hash }}` is the variant hash and is useful in the build string computation.
+This used to be `PKG_HASH` in the old recipe format. Since the `hash` variable
```
beeankha, line 393:
```suggestion
This used to be `PKG_HASH` in the `V0` recipe format. Since the `hash` variable
```

cep-17.md
```diff
@@ -0,0 +1,527 @@
...
+requirements:
+  - ${{ "six" if match(python, "<3.8") }}
+  - ${{ "six" if match(python, "3.8") }}
+  - ${{ "six" if match(python, "==3.8") }}
+  - ${{ "six" if match(python, "3.8.*") }}
+  - ${{ "six" if match(python, ">=3.8,<3.10") }}
+```
+
+In this case the value from the `python` _variant_ is used to add or remove
+optional dependencies. Note that generalizes and replaces selectors from old
```
beeankha, line 372:
```suggestion
optional dependencies. Note that generalizes and replaces selectors from `V0`
```

cep-17.md
beeankha: The title of this PR should be changed to reference `V1` recipe format vs `V2`

cep-17.md
```diff
@@ -0,0 +1,527 @@
...
+
+```yaml
+foo_compiler: "superfoo"
+foo_compiler_version: "1.2.3"
+
+# on linux-64 this then results in
+# compiler: "superfoo_linux-64 1.2.3"
+```
+
+> [!NOTE] Default values for `<lang>_compiler`
```
beeankha, line 157:
```suggestion
> [!NOTE]
>
> Default values for `<lang>_compiler`
```

JeanChristopheMorinPerso:
@wolfv @jaimergp I was searching for the CEP and couldn't find it. Should we merged this PR? It looks like the CEP was approved.

jaimergp:
I don't know how we missed this but there we go, two years after approval, this CEP is hence minted as CEP 39. Merging!

JeanChristopheMorinPerso:
Thank you!

PR #72, opened 2024-04-13
add recipe extension cep

wolfv:


wolfv:
Discussion:

Technically, it would be nicer to solve this with a flexible script. Challenge: we only want to run over the _new files in the prefix_.

Should we leak "rust" regex features? Would be easy to use them from Python as well. We can also turn off the regex feature where capture groups can be referenced easily with `$...`.

Example from [Rust docs](https://docs.rs/regex/latest/regex/struct.Regex.html#method.replace): 
```rust
use regex::Regex;

let re = Regex::new(r"(?<last>[^,\s]+),\s+(?<first>\S+)").unwrap();
let result = re.replace("Springsteen, Bruce", "$first $last");
assert_eq!(result, "Bruce Springsteen");
```

jaimergp:
Python has `\g<name>` and `\g<number>` for that. So maybe there's a way to convert things more or less easily. The `$` syntax is reasonable enough to me.

jaimergp:
> Challenge: we only want to run over the new files in the prefix.

Can that be injected in the `post_build` namespace? A `\0` separated list of paths in an env var or something? Maybe it overflows 😬. Maybe a file can be placed somewhere strategically which, upon parsing, returns the newly detected files.

cep-recipe-extensions.md
```diff
@@ -0,0 +1,54 @@
+# Extensions to the recipe spec
+
+This document describes extensions to the recipe specification as merged in cep-14.
+
+## The `build.post_process` section
+
+We became aware that the `conda_build_config.yaml` has ways to specify post-processing steps. These are regex replacements that are performed on any new files.
```
jaimergp, line 7:
```suggestion
The `conda_build_config.yaml` configuration file has ways to specify post-processing steps. These are regex replacements that are performed on any new files. 
```

Link to documentation, if available.

cep-recipe-extensions.md
```diff
@@ -0,0 +1,54 @@
...
+        - "*.txt"
+      regex: "foo"
+      replacement: "bar"
+    - files:
+        - "*.cmake"
+      regex: "/sysroot/"
+      replacement: "$${PREFIX}/"
+```
+
+## Globs, positive and negative
```
jaimergp, line 35: Does this have to do with recipe post-processing? 🤔

chenghlee: Yeah. This really feels like it should just be incorporated into the build script(s).

h-vetinari: The positive/negative globs should be part of the recipe. Using the build scripts just to shuffle files around into separate output is a lot of core for little gain, especially because it's a well-defined operation and doesn't overload the recipe syntax IMO.

C.f. also https://github.com/conda/conda-build/pull/5216

cep-recipe-extensions.md
wolfv: I don't think that there are any docs for this 🙈

cep-recipe-extensions.md
wolfv: No this doesn't really have anything to do with post-processing per se. The idea is that any list of globs can also include negative globs. 

List of globs appear in multiple places in the recipe.

wolfv:
TODO:

- [ ] I have recently been thinking about adding the compression settings to the `build` section so that compression can be configured easily per-output.

PR #74, opened 2024-04-13
CEP 40: Recipe serialization in packages

wolfv:
Rendered version: ✏️ https://github.com/wolfv/ceps/blob/cep-recipe-serialization/cep-recipe-serialization.md


cep-recipe-serialization.md
```diff
@@ -0,0 +1,1117 @@
...
+    ├── recipe.yaml
+    ├── patches
+    │   └── patch1.diff
+    ├── rendered_recipe.yaml
+    └── variant_config.yaml
+```
+
+## Specification
+
+The rendered recipe YAML file contains the following top-level sections:
```
jaimergp, line 48: Should this be versioned too? Add a `rendered_recipe_version: 1` or something?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1117 @@
...
+    environment.
+  - `build_prefix`: The build prefix directory for the build environment.
+  - `work_dir`: The working directory with extracted sources.
+  - `build_dir`: The build directory that contains `host_prefix`,
+    `build_prefix`, and `work_dir` as subdirectories.
+
+##### `channels`
+
+- Type: array of strings
+- Description: The conda channels used for resolving dependencies (e.g.,
```
jaimergp, line 183: Is this expecting the name of the channel? That might be ambiguous. What about the resolved URL? I guess without authentication details. Also, given how the resolved dependencies already contain the lockfile-like info... is this needed?

jaimergp: If we want to fully specify this, I think it's more specific to specify which repodata URLs where used and in which order and priority style (strict, flexible, etc).

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1117 @@
...
+```yaml
+url: # URL to the source code
+sha256: # sha256 hash of the downloaded file
+md5: # md5 hash of the downloaded file
+file_name: # optional file name to rename the downloaded file
+patches: # list of patches that should be applied to the source
+target_directory: # optional target directory where the source should be extracted
+```
+
+### System tools
```
jaimergp, line 399: This is implicit, but I guess that this dictionary is a `dict[str, str]` one right? Tool name string to version string.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1117 @@
...
+system_tools:
+  rattler_build: 0.14.0
+  patchelf: 0.12
+  git: 2.33.0
+```
+
+## Full Example
+
+```yaml
+recipe:
```
schuylermartin45, line 418: Just going to throwing this out there, are we worried about how these files scale? I can imagine some very complex multi-output, multi-platform recipes could produce very long `rendered_recipe.yaml` files.

wolfv: Well, each package only contains a rendered recipe for a single output, for a single platform. So it's not going to be crazy big. But lockfiles can indeed become relatively large. I think I would expect at most something like ~1 Mb. However, that should compress pretty well (except for the checksums). And the benefits are pretty huge (SBOMs, reproducibility, ...).

wolfv: It would, however, be an option to only store e.g. the URL + 1 hash. That would reduce the size quite decently. Happy to consider / discuss that.

schuylermartin45: Ah I missed that there was one file per output. I saw the build section and immediately thought those were individual outputs.

I don't have a strong opinion, I'm just pointing out this as a potential future headache. I know we have some large manifest-type files in `conda` world (looking at you, `repodata.json`) and I'm sure we all agree we don't want to repeat that pattern.

cep-recipe-serialization.md
baszalmstra: This should be the resolved URLs indeed.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+  run:
+    depends:
+      - source: run_export
+        spec: libzlib >=1.3.1,<1.4.0a0
+        from: host
+        source_package: zlib
+    constrains: []
+    run_exports: null
+finalized_sources:
+  - url: http://curl.haxx.se/download/curl-8.0.1.tar.bz2
```
wolfv, line 1105: We could add a content hash here (as proposed recently in conda-build).

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+This section records all system tools that were used during build execution and
+their versions (if available). This is useful for tracking the tools that were
+used from the host system in order to warn if reproducing the build on a
+different host system might lead to different results. It also records the
+version of `rattler_build` itself, so a future build could be reproduced with
+the same version of `rattler_build`.
+
+```yaml
+system_tools:
+  rattler_build: 0.14.0
```
JeanChristopheMorinPerso, line 403: Should it be `rattler-build` to more easily map to the origin package?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+sha256: # sha256 hash of the downloaded file
+md5: # md5 hash of the downloaded file
+file_name: # optional file name to rename the downloaded file
+patches: # list of patches that should be applied to the source
+target_directory: # optional target directory where the source should be extracted
+```
+
+### System tools
+
+This section records all system tools that were used during build execution and
```
JeanChristopheMorinPerso, line 394: How does conda-build or rattler-build know which tool was used? Should this be somewhat defined?

dholth:
Why not .json?
Do we still need info/recipe/meta.yaml ?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+##### `packaging_settings`
+
+- Type: object
+- Description: Contains settings related to the final package output.
+  - `archive_type`: The type of archive used for packaging (either `tar_bz2` or
+    `conda`).
+  - `compression_level`: The numerical compression level used for packaging
+    (e.g., `15`).
+  - `compression_threads`: The number of threads used for compression (e.g.,
+    `1`).
```
chenghlee, line 204: Is there a reason to preserve the number of compression threads? I honestly would not expect whatever compression algorithm we use to change outputs based on the number of threads used.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+    compression_threads: 1
+```
+
+### Finalized dependencies
+
+The finalized dependencies section contains a "lockfile" of all the dependencies
+exactly as they were used at build-time, for the two environments `build` and
+`host`. It also includes the final `run` and `constrains` specs that are
+computed based on the `build` and `host` dependencies (resolved all
+`run_exports`).
```
chenghlee, line 244: Should we consider making each of these sections an extension of an already existing lock format, rather than creating yet another one? Seems like it may be super convenient if we could extract, e.g., the `build:` section and feed that YAML directly into `conda-lock`.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+  run:
+    depends: [list of dependencies with dependency source information]
+    constrains: [list of constrains with dependency source information]
+    run_exports: optional run_export dictionary
+```
+
+#### Spec objects
+
+A dependency spec contains the `MatchSpec` and additional information about the
+source of the dependency. There are 6 possible `DependencyInfo` versions:
```
chenghlee, line 267: This appears to be the only mention of `DependencyInfo`. It's not clear to me what its purpose is.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+    run_exports: { package_name: run_export } # all run_exports.json that were found among the packages in this environment
+  run:
+    depends: [list of dependencies with dependency source information]
+    constrains: [list of constrains with dependency source information]
+    run_exports: optional run_export dictionary
+```
+
+#### Spec objects
+
+A dependency spec contains the `MatchSpec` and additional information about the
```
chenghlee, line 266: This reminds me: we should probably write a `MatchSpec` CEP. :smile:

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+name: libtool
+sha256: efe50471d2baea151f2a93f1f99c05553f8c88e3f0065cdc267e1aa2e8c42449
+size: 408403
+subdir: osx-arm64
+timestamp: 1672362003513 # when available
+version: 2.4.7
+fn: libtool-2.4.7-hb7217d7_0.conda
+url: https://conda.anaconda.org/conda-forge/osx-arm64/libtool-2.4.7-hb7217d7_0.conda
+channel: https://conda.anaconda.org/conda-forge/
+```
```
chenghlee, line 315: Is there a reason to repeat the contents for `RepoDataRecord`? Seems to me that the URL and SHA256 checksum of the package would be enough to recreate the various build environments.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+> [!NOTE] The `sha256` field is added if it is not already present in the source
+> description. (TODO in rattler-build) We should compute the SHA hashes of the
+> patches as well and store it in the source description.
+
+```yaml
+path: # path to the source
+sha256: # computed sha256 hash of the source (this is added if not in the source description)
+patches: # list of patches that should be applied to the source
+target_directory: # optional target directory where the source should be extracted
+file_name: # optional file name
```
chenghlee, line 366: Wouldn't the filename be part of `path`?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+pub use_gitignore: # optional bool wether to use the .gitignore file when copying a directory
+```
+
+#### Git source
+
+```yaml
+git: # git URL
+rev: # resolved commit hash at build time
+depth: # optional depth to clone the repository
+patches: # list of patches that should be applied to the source
```
chenghlee, line 376: We should clarify where these patches come from. (I suspect, relative to the top-level of the directory, but it should be explicitly stated.)

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+file_name: # optional file name
+pub use_gitignore: # optional bool wether to use the .gitignore file when copying a directory
+```
+
+#### Git source
+
+```yaml
+git: # git URL
+rev: # resolved commit hash at build time
+depth: # optional depth to clone the repository
```
chenghlee, line 375: Is this supposed to correspond to the `--depth` argument to `git clone` (which basically truncates history), or to (any) submodule clone/init operations?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+
+#### URL source
+
+```yaml
+url: # URL to the source code
+sha256: # sha256 hash of the downloaded file
+md5: # md5 hash of the downloaded file
+file_name: # optional file name to rename the downloaded file
+patches: # list of patches that should be applied to the source
+target_directory: # optional target directory where the source should be extracted
```
chenghlee, line 389: See previous comments re: `target_directory`.

cep-recipe-serialization.md
wolfv: yeah, I agree. sha hash and url should be enough. I think it might be interesting to include additional information (e.g. licenses) so that tools could report on what-went-into the package ... Of course, given that we still have access to the packages we could recreate this information from the repodata / packages, too.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+
+- Type: array of strings
+- Description: The conda channels used for resolving dependencies (e.g.,
+  `conda-forge`).
+
+##### `timestamp`
+
+- Type: string / timestamp in ISO 8601 format
+- Description: The timestamp of the build configuration (e.g.,
+  `2024-04-13T14:35:30.774863Z`).
```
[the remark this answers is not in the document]
wolfv: Yeah this wording is apparently confusing. It's the timestamp when the build was started.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+
+Some key things to note:
+
+- Compiler expressions like ${{ compiler('c') }} are rendered to their final
+  YAML dictionary form:
+
+  ```yaml
+  - ${{ compiler('c') }}
+  # becomes
+  - compiler: c
```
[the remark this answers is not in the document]
wolfv: We currently have an "intermediate" form in rattler-build.  We need that for `pin_subpackage` because that needs to have the rest of the recipe rendered first. 
However, for `compiler` we can make the function return the final value instead. It would have some other benefits (e.g. that we can use the `compiler` function in other contexts easily).

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+
+- Type: string
+- Description: The platform used for building the package (e.g., `osx-arm64`).
+
+##### `variant`
+
+- Type: object
+- Description: Dictionary with the variant configuration used for the build.
+
+##### `hash`
```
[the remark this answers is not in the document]
wolfv: Usually it's a hash "prefix", right? E.g. `py38hxxxxx_5` is `prefix: py38`, `h` (separator), `hash: xxxxx` and `_5` is derived from the build number

cep-recipe-serialization.md
wolfv: yeah it was more about being overly cautious. We can run some tests wether the number of threads has any effect.

cep-recipe-serialization.md
wolfv: Since we don't have a conda-lock spec makes it more tricky. However, building this functionality should be very easy (all the metadata is present in the proposed file).

cep-recipe-serialization.md
wolfv: That would be awesome!

cep-recipe-serialization.md
wolfv: I think there might be a few more arguments:

- repodata changes over time and this is an accurate snapshot as how it was at the time
- it makes it easier to extract a `conda-lock` file from the rendered recipe
- there might be additioanl opportunities for tooling in the future (license scanning, ...)

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+finalized_sources:
+  - path: ../
+  - git: https://github.com/ros2-gbp/rosidl_typesupport_fastrtps-release.git
+    rev: 7af149d83d7e4a2d4c9ce4ae9682bff4a8ebc28e # here we are storing the commit hash
+    target_directory: ros-humble-fastrtps-cmake-module/src/work
+  - url: https://github.com/watchexec/watchexec/archive/refs/tags/v1.25.1.tar.gz
+    sha256: 9609163c14cd49ec651562838f38b88ed2d370e354af312ddc78c2be76c08d37
+```
+
+The following fields are available for each source:
```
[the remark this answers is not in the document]
wolfv: The finalized sources are almost but not exactly. We are adding the _actual_ `rev` of the git repository (so that we can retrieve the actual commit when rebuilding).

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1141 @@
...
+    - `version`: The version of the subpackage (e.g., `8.0.1`).
+    - `build_string`: The build string for the subpackage (e.g., `h60d57d3_0`).
+
+##### `packaging_settings`
+
+- Type: object
+- Description: Contains settings related to the final package output.
+  - `archive_type`: The type of archive used for packaging (either `tar_bz2` or
+    `conda`).
+  - `compression_level`: The numerical compression level used for packaging
```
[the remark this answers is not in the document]
wolfv: No, it's for perfectly reproducible builds (https://github.com/prefix-dev/reproducible-builds)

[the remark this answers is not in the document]
wolfv: Overview here: https://prefix-dev.github.io/reproducible-builds/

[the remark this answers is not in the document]
dholth: Would it be a better idea to reproduce the data after compression? I also notice the number of threads has been removed? The zstandard version is recorded somewhere or implied by the tool version?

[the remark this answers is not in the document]
baszalmstra: The number of threads doesnt influence the final artifact so we removed it.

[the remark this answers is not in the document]
wolfv: In the ideal world we can re-create the compressed artifact bit-for-bit. Content hashes are also important. But for security reasons and so on I believe it's nicer to be able to _really_ recreate the artifact. At least I believe that's what most people in reproducible-builds.org are trying :)

cep-recipe-serialization.md
wolfv: Filename is more for renaming the file in the work directory. So yes, you could add a `file_name` argument if you want (and then it would be rendered).

cep-recipe-serialization.md
wolfv: Yeah this is to truncate history (`conda-build` has the same argument).

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1141 @@
...
+        name: ca-certificates
+        sha256: 49bc3439816ac72d0c0e0f144b8cc870fdcc4adec2e861407ec818d8116b2204
+        size: 155725
+        subdir: osx-arm64
+        timestamp: 1706844034242
+        version: 2024.2.2
+        fn: ca-certificates-2024.2.2-hf0a4a13_0.conda
+        url: https://conda.anaconda.org/conda-forge/osx-arm64/ca-certificates-2024.2.2-hf0a4a13_0.conda
+        channel: https://conda.anaconda.org/conda-forge/
+    run_exports:
```
wolfv, line 1103: @baszalmstra - we removed these here, right?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1141 @@
...
+        name: libzlib
+        sha256: 9c59bd3f3e3e1900620e3a85c04d3a3699cb93c714333d06cb8afdd7addaae9d
+        size: 47040
+        subdir: osx-arm64
+        timestamp: 1709241137619
+        version: 1.3.1
+        fn: libzlib-1.3.1-h0d3ecfb_0.conda
+        url: https://conda.anaconda.org/conda-forge/osx-arm64/libzlib-1.3.1-h0d3ecfb_0.conda
+        channel: https://conda.anaconda.org/conda-forge/
+    run_exports:
```
wolfv, line 1147: here again

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1141 @@
...
+      perl:
+        weak:
+          - perl >=5.32.1,<5.33.0a0 *_perl5
+        noarch:
+          - perl >=5.32.1,<6.0a0 *_perl5
+  host:
+    specs:
+      - source: raw
```
wolfv, line 1111: I think this needs to be updated as well

cep-recipe-serialization.md
wolfv: we made this change :)

wolfv (commented):
I just did another pass as well.

cep-recipe-serialization.md
baszalmstra: I did not!

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+
+- Type: object
+- Description: Dictionary with the variant configuration used for the build.
+
+##### `hash`
+
+- Type: object
+- Description: Contains information related to hashing.
+  - `hash`: The hash value (e.g., `60d57d3`).
+  - `hash_input`: The input values used for generating the hash.
```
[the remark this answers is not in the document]
baszalmstra: The `hash_input` has been completely removed.

cep-recipe-serialization.md
baszalmstra: We ended up running these tests, as claimed by zstd there was no difference so we removed this key.

cep-recipe-serialization.md
baszalmstra: I updated them in 00ad601

cep-recipe-serialization.md
baszalmstra: I updated them in 00ad601

cep-recipe-serialization.md
baszalmstra: Yep updated this in 00ad601

cep-recipe-serialization.md
wolfv: I added a `used_build_tool.json` file in the specification :)

jezdez:
@conda/steering-council

baszalmstra:
yes

wolfv:
Please use the following form to vote: 

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@kkraus14 (Keith Kraus)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [ ]  yes
- [ ]  no
- [ ]  abstain

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1190 @@
...
+```
+
+Will be copied into the following folder structure in the final package:
+
+```txt
+├── ...
+├── index.json
+├── used_build_tool.json  // new file!
+└── recipe
+    ├── build.bat
```
jaimergp, line 69: Huh, this has changed from `bld.bat`?

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1190 @@
...
+
+#### Recipe section
+
+This section is the fully rendered recipe in valid YAML format, with all Jinja2
+expressions evaluated and if/else logic resolved to the final values used during
+the build.
+
+Some key things to note:
+
+- Compiler expressions like ${{ compiler('c') }} are rendered to their final form immediately:
```
jaimergp, line 107:
```suggestion
- Compiler expressions like `${{ compiler('c') }}` are rendered to their final form immediately:
```

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1110 @@
...
+  - `hash_prefix`: Prefix for the hash, which can include values like `py311`
+    based on the variant configuration.
+
+##### `directories`
+
+- Type: object
+- Description: Contains directory paths used during the build process.
+  - `host_prefix`: The host prefix directory with placeholders for the host
+    environment.
+  - `build_prefix`: The build prefix directory for the build environment.
```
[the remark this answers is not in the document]
jaimergp: I assume merged build and host means that host_prefix == build_prefix, so both values are the same. Alternatively, host_prefix == null.

jaimergp (approved):
Just a few typos I found.

cep-recipe-serialization.md
wolfv: yep :)

wolfv:
@marcelotrevisani @jakirkham @cj-wright @mbargull last chance to vote!

chenghlee:
For the record: voted "no" because while I like the ideas presented in this CEP, I don't think the CEP as written is ready to be adopted _as a specification_. IMO, various unanswered questions/comments need to be addressed before adoption, and we should better separate behavioral specifications from implementation details.

mbargull:
I think it is good to have these parts of the build output/process specified which is why I support to have a proposal for it here.
I do agree with Cheng in that this is not yet ready to be put in, though, since the "Specification" section is a bit light in detail and the bigger part of the text, i.e., description of implementation details, are likely to change, so nothing we'd set in stone here.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1190 @@
...
+- pin_compatible: quarto
+  lower_bound: x.x    # not present if set to None
+  upper_bound: x.x.x  # not present if set to None
+  exact: true         # not present if false
+  spec: quarto >=1.4.550,<1.5
+- pin_subpackage: some-pkg-a
+  lower_bound: x.x    # not present if set to None
+  upper_bound: x.x.x  # not present if set to None
+  exact: true         # not present if false
+  spec: some-pkg-a >=1.0.0,<2
```
jakirkham, line 377: Am curious how these work together. We have `exact` and a more relaxed `spec` as well bounds. What takes precedence?

jakirkham (approved):
Thanks Wolf! 🙏

Think these are good improvements. Am voting yes with a few suggestions/questions (tried to use the checkbox, but had issues with it for some reason)

mbargull:
> (tried to use the checkbox, but had issues with it for some reason)

@jakirkham 
It's really odd that it doesn't like to accept your edit :/.
I checked "yes" on your behalf.

cep-recipe-serialization.md
```diff
@@ -0,0 +1,1190 @@
...
+      exact: true
+  ```
+
+- For the following:
+
+  ```yaml
+  - ${{ pin_subpackage('some-pkg-a', lower_bound="x.x.x") }}
+  - ${{ pin_compatible('python', upper_bound="x.x.x") }}
+  - ${{ pin_compatible('numpy', upper_bound="3.2.1", lower_bound="1.2.3") }}
+  ```
```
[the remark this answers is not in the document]
h-vetinari: This is intentionally unified in the recent set of CEPs, please see this [thread](https://github.com/conda/ceps/pull/71#discussion_r1609521754)

jakirkham:
Thanks Marcel! 🙏

Yeah this happened to me a couple times. So just used "approved" to mean the same thing. Think it is because I tried on my iPhone with [GitHub Mobile]( https://github.com/mobile ) (not sure why it can't check a box or why it makes an empty edit though 🤷‍♂️)

wolfv:

Total voters: 15 (valid: 13 = 86.67%)

Yes votes (11 / 84.62%):
- @xhochy (Uwe Korn)
- @goanpeca (Gonzalo Peña-Castellanos)
- @ocefpaf (Filipe Fernandes)
- @msarahan (Michael Sarahan)
- @jakirkham (John Kirkham)
- @jezdez (Jannis Leidel)
- @wolfv (Wolf Vollprecht)
- @kkraus14 (Keith Kraus)
- @baszalmstra (Bas Zalmstra)

No votes (2 / 15.38%)):
- @chenghlee (Cheng H. Lee)
- @mbargull (Marcel Bargull)

Abstain votes (0 / 0.00%):

Not voted (2):

Invalid votes (0):

Thus we reached quorum and enough YES votes to mark this as accepted. 🎉

cep-recipe-serialization.md
wolfv: They should not be allowed together, right?

jaimergp:
We also missed minting this one, but better late than never. Welcome CEP 40!

PR #75, opened 2024-04-30
CEP: sharded repodata

baszalmstra:
We propose a new solution to repodata downloading aimed at drastically reducing the time and memory required.

We implemented the use of this proposal in [rattler](https://github.com/mamba-org/rattler) and created [a mirror](https://github.com/prefix-dev/sharded-repo/) for popular channels containing the new files that are updated at a fixed interval.

[📝 Rendered](https://github.com/baszalmstra/ceps/blob/sharded-repodata/cep-16.md)

### Preliminary results

Requesting repodata records for a specific set of packages for linux-64 and noarch. Measured on a machine with a 200mbit internet connection.

```
| requested packages        | records | fresh | cached | JLAP  | sharded (cold) | sharded (hot) |
| ------------------------- | ------- | ----- | ------ | ----- | -------------- | ------------- |
| python + boto3 + requests | 6969    | 2.4 s | 0.58 s | 8.2 s | 0.34 s         | 0.015 s       |
| jupyterlab + detectron2   | 35524   | 2.4 s | 0.63 s | 8.7 s | 0.7 s          | 0.05 s        |
| rubin-env                 | 84254   | 2.9 s | 0.80 s | 8.8 s | 1 s            | 0.13 s        |
```

Note that with this approach the cache is not all or nothing. So even if the shards index is updated, most likely a user will still have a relatively hot cache.

### DIFF
--- a/cep-16.md
+++ b/cep-16.md
@@ -1,6 +1,8 @@
-# Sharded Repodata
+# CEP for Sparse Repodata
 
 We propose a new "repodata" format that can be sparsely fetched. That means, generally, smaller fetches (only fetch what you need) and faster updates of existing repodata (only fetch what has changed).
+
+We also change the encoding from JSON to MSGPACK for faster decoding.
 
 ## Motivation
 
@@ -31,7 +33,7 @@ Finally, the implementation of JLAP is quite complex which makes it hard to adop
 
 ### ZSTD compression
 
-A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% smaller (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice, this yields a file that is 1/5th the size (20-30 Mb for large cases). Although this is still quite a big file it's substantially smaller. 
 
 However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
 
@@ -73,7 +75,7 @@ The contents look like the following (written in JSON for readability):
 
 The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
 
-For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
+For a large case (conda-forge linux-64) this file is 670kb at the time of writing.
 
 We suggest serving the file with a short lived `Cache-Control` `max-age` header of 60 seconds to an hour but we leave it up to the channel administrator to set a value that works for that channel.
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -50,7 +50,7 @@ Although not explicitly required the server SHOULD support HTTP/2 to reduce the 
 
 ### Repodata shard index
 
-The shard index is a file that is stored under `<shard_base_url>/<subdir>/repodata_shards.msgpack.zst`. It is a zstandard compressed `msgpack` file that contains a mapping from package name to shard hash. The `<shard_base_url>` is defined in [Authentication](#authentication).
+The shard index is a file that is stored under `<subdir>/repodata_shards.msgpack.zst`. It is a zstandard compressed `msgpack` file that contains a mapping from package name to shard hash.
 
 The contents look like the following (written in JSON for readability):
 
@@ -59,6 +59,7 @@ The contents look like the following (written in JSON for readability):
   "version": 1,
   "info": {
     "base_url": "https://example.com/channel/subdir/",
+    "shards_base_url": "./shards/",
     "created_at": "2022-01-01T00:00:00Z",
     "...": "other metadata"
   },
@@ -71,6 +72,10 @@ The contents look like the following (written in JSON for readability):
 }
 ```
 
+The `shards_base_url` is the base url for the shards. 
+Both the `base_url` and the `shards_base_url` can be absolute or a relative url. 
+The urls are relative to the shard index.
+
 The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
 
 For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
@@ -79,7 +84,9 @@ We suggest serving the file with a short lived `Cache-Control` `max-age` header 
 
 ### Repodata shard
 
-Individual shards are stored under the URL `<shard_base_url>/<subdir>/shards/<sha256>.msgpack.zst`. Where the `sha256` is the lower-case hex representation of the bytes from the index. It is a zstandard compressed msgpack file that contains the metadata of the package. The `<shard_base_url>` is defined in [Authentication](#authentication).
+Individual shards are stored under the URL `<shards_base_url><sha256>.msgpack.zst`. 
+Where the `sha256` is the lower-case hex representation of the bytes from the index and `shards_base_url` is defined in the shard index. 
+It is a zstandard compressed msgpack file that contains the metadata of the package. 
 
 The files are content-addressable which makes them ideal to be served through a CDN. They SHOULD be served with `Cache-Control: immutable` header.
 
@@ -152,37 +159,15 @@ Implementers SHOULD ignore unknown keys, this allows adding additional keys to t
 
 Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remaing very small, e.g. 100s of bytes to a couple of kilobytes.
 
-## <a id="authentication"></a>Authentication
-
-To faciliate authentication and authorization we propose to add an additional endpoint at `<channel>/<subdir>/token` with the following content:
-
-```json
-{
-    "shard_base_url": "https://shards.prefix.dev/conda-forge/<subdir>/",
-    "token": "<bearer token>",
-    "issued_at": "2024-01-30T03:35:39.896023447Z",
-    "expires_in": 300,
-}
-```
-
-`shard_base_url` is an optional url to use as the base url for the `repodata_shards.msgpack.zst` and the individual shards. If the field is not specified it should default to `<channel>/<subdir>`.
-
-`token` is an optional field that if set MUST be added to any subsequent request in the `Authentication` header as `Authentication: Bearer <token>`. If the `token` field is not set sending the `Authentication` header is also not required. 
-
-The optional `issued_at` and `expires_in` fields can be used to verify the freshness of a token. If the fields are not present a client can assume that the token is valid for any subsequent request.
-
-For a simple implementor this endpoint could just be a static file with `{}` as the content.
-
 ## Fetch process
 
 To fetch all needed package records, the client should implement the following steps:
 
-1. Acquire a token (see: [Authentication](#authentication)). Acquiring a token can be done lazily as to only request a token when an actual network request is performed.
-2. Fetch the `repodata_shards.msgpack.zst` file. Standard HTTP caching semantics can be applied to this file.
+1. Fetch the `repodata_shards.msgpack.zst` file. Standard HTTP caching semantics can be applied to this file.
 3. For each package name, start fetching the corresponding hashes from the index file (for both arch & and noarch). 
     Shards can be cached locally and because they are content-addressable no additional round-trips to the server are required to check freshness. The server should also mark these with an `immutable` `Cache-Control` header.
 4. Parsing the requirements of the fetched records and add the package names of the requirements to the set of packages to fetch.
-5. Loop back to 2. until there are no new package names to fetch.
+5. Loop back to 1. until there are no new package names to fetch.
 
 ## Garbage collection
 
@@ -232,6 +217,13 @@ vs
 This did yield slightly better compression but we felt it makes it slightly harder to implement and adapt in the future which we deemed not worth the small size decrease.
 
 ## Future improvements
+
+### Authentication
+
+With this approach a client could do hundreds of requests concurrently. 
+Authenticating all these requests would impose a non-negligable overhead to each requests and will also require more server resources. 
+Initially a one-time authentication flow similar to OCI registry tokens was added to the CEP but this has been removed to reduce the scope of the CEP. 
+Another CEP will be introduced to amend this CEP with a performant authentication flow.
 
 ### Remove redundant keys
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -31,7 +31,7 @@ Finally, the implementation of JLAP is quite complex which makes it hard to adop
 
 ### ZSTD compression
 
-A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% smaller (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% of the original size (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
 
 However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -22,7 +22,7 @@ The current repodata format is a JSON file that contains all the packages in a g
 
 ### JLAP
 
-In a previous CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
+In a previously proposed CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
 With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
 
 However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
--- a/cep-16.md
+++ b/cep-16.md
@@ -1,8 +1,6 @@
-# CEP for Sparse Repodata
+# Sharded Repodata
 
 We propose a new "repodata" format that can be sparsely fetched. That means, generally, smaller fetches (only fetch what you need) and faster updates of existing repodata (only fetch what has changed).
-
-We also change the encoding from JSON to MSGPACK for faster decoding.
 
 ## Motivation
 
@@ -33,7 +31,7 @@ Finally, the implementation of JLAP is quite complex which makes it hard to adop
 
 ### ZSTD compression
 
-A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice, this yields a file that is 1/5th the size (20-30 Mb for large cases). Although this is still quite a big file it's substantially smaller. 
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% smaller (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
 
 However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
 
@@ -75,7 +73,7 @@ The contents look like the following (written in JSON for readability):
 
 The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
 
-For a large case (conda-forge linux-64) this file is 670kb at the time of writing.
+For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
 
 We suggest serving the file with a short lived `Cache-Control` `max-age` header of 60 seconds to an hour but we leave it up to the channel administrator to set a value that works for that channel.
 
@@ -85,7 +83,8 @@ Individual shards are stored under the URL `<shard_base_url>/<subdir>/shards/<sh
 
 The files are content-addressable which makes them ideal to be served through a CDN. They SHOULD be served with `Cache-Control: immutable` header.
 
-The shard contains the repodata information that would otherwise have been found in the `repodata.json` file. It is a dictionary that contains the following keys:
+The shard contains the repodata information that would otherwise have been found in the `repodata.json` file. 
+It is a dictionary that contains the following keys:
 
 **Example (written in JSON for readability):**
 
@@ -139,6 +138,9 @@ The shard contains the repodata information that would otherwise have been found
   }
 }
 ```
+
+The `sha256` and `md5` from the original repodata fields are converted from their hex representation to bytes. 
+This is done to reduce the overall file size of the shards.
 
 Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remaing very small, e.g. 100s of bytes to a couple of kilobytes.
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -1,3 +1,13 @@
+<table>
+<tr><td> Title </td><td> Sharded Repodata </td>
+<tr><td> Status </td><td> Accepted </td></tr>
+<tr><td> Author(s) </td><td> Bas Zalmstra &lt;bas@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> April 30, 2024</td></tr>
+<tr><td> Updated </td><td> July 22, 2024</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda-incubator/ceps/pull/75 </td></tr>
+<tr><td> Implementation </td><td>  </td></tr>
+</table>
+
 # Sharded Repodata
 
 We propose a new "repodata" format that can be sparsely fetched. That means, generally, smaller fetches (only fetch what you need) and faster updates of existing repodata (only fetch what has changed).
--- a/cep-16.md
+++ b/cep-16.md
@@ -42,7 +42,7 @@ Because the file is relatively big this means that often a large `max-age` is us
 We propose a "sharded" repodata format. It works by splitting the repodata into multiple files (one per package name) and recursively fetching the "shards".
 
 The shards are stored by the hash of their content (e.g. "content-addressable"). 
-That means that the URL of the shard is derived from the content of the shard. This allows for efficient caching and deduplication of shards on the client. Because the files are content-addressable no round-trip to the server is required to check fressness of individual shards.
+That means that the URL of the shard is derived from the content of the shard. This allows for efficient caching and deduplication of shards on the client. Because the files are content-addressable no round-trip to the server is required to check freshness of individual shards.
 
 Additionally an index file stores the mapping from package name to shard hash.
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -23,7 +23,7 @@ The current repodata format is a JSON file that contains all the packages in a g
 In a previously proposed CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
 With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
 
-However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
+However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the sheer amount of data involved.
 
 JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. This is often the case for CI runners.
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -31,7 +31,7 @@ Finally, the implementation of JLAP is quite complex which makes it hard to adop
 
 ### ZSTD compression
 
-A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% of the original size (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% of the original size (20-30 Mb for large cases). Although this is still quite a big file it's substantially smaller. 
 
 However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
 
@@ -78,7 +78,7 @@ The urls are relative to the shard index.
 
 The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
 
-For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
+For a large case (conda-forge linux-64) this file is 670kb at the time of writing.
 
 We suggest serving the file with a short lived `Cache-Control` `max-age` header of 60 seconds to an hour but we leave it up to the channel administrator to set a value that works for that channel.
 
@@ -157,7 +157,7 @@ This is done to reduce the overall file size of the shards.
 
 Implementers SHOULD ignore unknown keys, this allows adding additional keys to the format in the future without breaking old versions of tools.
 
-Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remaing very small, e.g. 100s of bytes to a couple of kilobytes.
+Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remain very small, e.g. 100s of bytes to a couple of kilobytes.
 
 ## Fetch process
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -27,7 +27,7 @@ With JLAP only the changes to an initially downloaded `repodata.json` file have 
 
 However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
 
-JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. THis is often the case for CI runners.
+JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. This is often the case for CI runners.
 
 Finally, the implementation of JLAP is quite complex which makes it hard to adopt for implementers. 
 
@@ -56,7 +56,7 @@ The shard index is a file that is stored under `<shard_base_url>/<subdir>/repoda
 
 The contents look like the following (written in JSON for readability):
 
-```json
+```js
 {
   "version": 1,
   "info": {
@@ -229,7 +229,7 @@ This did yield slightly better compression but we felt it makes it slightly hard
 
 ### Integrating additional data
 
-With the total size of the repodata reduced it becomes feasible to add additional fields directly to the repodata records. Exampels are:
+With the total size of the repodata reduced it becomes feasible to add additional fields directly to the repodata records. Examples are:
 
 - add `purl` as a list of strings (Package URLs to reference to original source of the package) (See: https://github.com/conda-incubator/ceps/pull/63)
 - add `run_exports` as a list of strings (run-exports needed to build the package) (See: https://github.com/conda-incubator/ceps/pull/51)
@@ -247,4 +247,4 @@ They will contain the same format as the `repodata_shards.msgpack.zst` file but 
 
 ### Store `set(dependencies)` at the start of the shards or in a header
 
-To reduce the time it takes to parse a shard and start fetching its dependencies we could also store the set of all dependencies in the file at the start of the shard or in a seperate header. This could enable fetching recursive dependencies while still parsing the records.
+To reduce the time it takes to parse a shard and start fetching its dependencies we could also store the set of all dependencies in the file at the start of the shard or in a separate header. This could enable fetching recursive dependencies while still parsing the records.
--- a/cep-16.md
+++ b/cep-16.md
@@ -88,8 +88,9 @@ It is a dictionary that contains the following keys:
 
 **Example (written in JSON for readability):**
 
-```json
+```js
 {
+  // dictionary of .tar.bz2 files
   "packages": {
     "rich-10.15.2-pyhd8ed1ab_1.tar.bz2": {
       "build": "pyhd8ed1ab_1",
@@ -114,6 +115,7 @@ It is a dictionary that contains the following keys:
       "version": "10.15.2"
     }
   },
+ // dictionary of .conda files
   "packages.conda": {
     "rich-13.7.1-pyhd8ed1ab_0.conda": {
       "build": "pyhd8ed1ab_0",
@@ -135,7 +137,11 @@ It is a dictionary that contains the following keys:
       "timestamp": 1709150578093,
       "version": "13.7.1"
     }
-  }
+  },
+  // list of strings of keys (filenames) that were removed from either packages or packages.conda
+  "removed": [
+    "rich-10.15.1-pyhd8ed1ab_1.tar.bz2"
+  ]
 }
 ```
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -253,7 +253,7 @@ For this we propose to add the following two files:
 - `<channel>/<subdir>/repodata_shards_daily.msgpack.zst`
 - `<channel>/<subdir>/repodata_shards_weekly.msgpack.zst`
 
-They will contain the same format as the `repodata_shards.msgpack.zst` file but only contain the packages that have been updated in the last day or week respectively. The `created_at` field in the index file can be used to determine which file to fetch to make sure that the client has the latest information.
+They will contain the same format as the `repodata_shards.msgpack.zst` file but only contain the packages that have been updated in the last day or week respectively. `null` is used for keys that have been removed. The `created_at` field in the index file can be used to determine which file to fetch to make sure that the client has the latest information.
 
 ### Store `set(dependencies)` at the start of the shards or in a header
 
--- a/cep-16.md
+++ b/cep-16.md
@@ -1,0 +1,248 @@
+# Sharded Repodata
+
+We propose a new "repodata" format that can be sparsely fetched. That means, generally, smaller fetches (only fetch what you need) and faster updates of existing repodata (only fetch what has changed).
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


wolfv:
Note that in fairness we never got around to implement the faster "2-file" JLAP variant. However, the sharded repodata is a lot "simpler" in terms of updates, so we believe that in `rattler` this will supersede the JLAP-efforts.

cep-16.md
```diff
@@ -0,0 +1,250 @@
...
+5. **Easy to implement**: It MUST be relatively straightforward to implement to ease the adoption in different tools.
+6. **Client-side cachable**: If a user has a hot cache the user SHOULD only have to download small incremental changes. Preferably as little communication as possible with the server should be required to check freshness of the data.
+7. **Bandwidth optimized**: Any data that is transferred SHOULD be as small as possible.
+
+## Previous work
+
+### JLAP
+
+In a previous CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
+With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
```
jezdez, line 26: Note that the JLAP CEP draft did _not_ pass, so it's still a draft and it would be good to indicate that.

jezdez: Perfect, thank you!

cep-16.md
```diff
@@ -0,0 +1,250 @@
...
+## Previous work
+
+### JLAP
+
+In a previously proposed CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
+With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
+
+However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
+
+JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. THis is often the case for CI runners.
```
jaimergp, line 30:
```suggestion
JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. This is often the case for CI runners.
```

cep-16.md
```diff
@@ -0,0 +1,250 @@
...
+
+Although not explicitly required the server SHOULD support HTTP/2 to reduce the overhead of doing a massive number of requests. 
+
+### Repodata shard index
+
+The shard index is a file that is stored under `<shard_base_url>/<subdir>/repodata_shards.msgpack.zst`. It is a zstandard compressed `msgpack` file that contains a mapping from package name to shard hash. The `<shard_base_url>` is defined in [Authentication](#authentication).
+
+The contents look like the following (written in JSON for readability):
+
+```json
```
jaimergp, line 59:
```suggestion
```js
```

For better highlighting.

cep-16.md
```diff
@@ -0,0 +1,250 @@
...
+For this we propose to add the following two files:
+
+- `<channel>/<subdir>/repodata_shards_daily.msgpack.zst`
+- `<channel>/<subdir>/repodata_shards_weekly.msgpack.zst`
+
+They will contain the same format as the `repodata_shards.msgpack.zst` file but only contain the packages that have been updated in the last day or week respectively. The `created_at` field in the index file can be used to determine which file to fetch to make sure that the client has the latest information.
+
+### Store `set(dependencies)` at the start of the shards or in a header
+
+To reduce the time it takes to parse a shard and start fetching its dependencies we could also store the set of all dependencies in the file at the start of the shard or in a seperate header. This could enable fetching recursive dependencies while still parsing the records.
```
jaimergp, line 250:
```suggestion
To reduce the time it takes to parse a shard and start fetching its dependencies we could also store the set of all dependencies in the file at the start of the shard or in a separate header. This could enable fetching recursive dependencies while still parsing the records.
```

cep-16.md
```diff
@@ -0,0 +1,250 @@
+# CEP for Sparse Repodata
+
+We propose a new "repodata" format that can be sparsely fetched. That means, generally, smaller fetches (only fetch what you need) and faster updates of existing repodata (only fetch what has changed).
+
+We also change the encoding from JSON to MSGPACK for faster decoding.
```
jaimergp, line 5: Could this be considered a implementation detail? It imposes an additional dependency on implementors, but I think this should be mostly about the strategy of sharding, their path locations and the schema expecting. Maybe we can standardize things like `<path>/<filename>.<encoding>.<compression>`. Or at least standardize that a JSON counterpart is also available.

wolfv: Initially we used to hash the underlying data. Unfortunately that would not map well to the OCI registry. So we decided to use the hash of the compressed blob.

baszalmstra: And here are some basic numbers.

I generated the repodata shard index for conda-forges noarch subdir in both json and msgpack format. And I took the largest shard (`conda-forge-pinning`) for comparison as well.

```
| which               | size compressed | size uncompressed | deserialization time |
| ------------------- | --------------- | ----------------- | -------------------- |
| json shard index    | 705 KB          | 1404 KB           | 4.8 ms               |
| msgpack shard index | 666 KB (94%)    | 788 KB (56%)      | 3.2 ms (66%)         |
| json shard          | 298 KB          | 1926 KB           | 12.8 ms              |
| msgpack shard       | 313 KB (105%)   | 1567 KB (81%)     | 10.2 ms (83%)        |
```

In terms of size both JSON and msgpack files become roughly the same size. The biggest win seems to be in deserialization time which also takes up a significant amount of time in real-world examples.

Edit: If I change the format of the hashes in the shards from hex string to bytes, the msgpack files are always smaller (290 KB for the conda-forge-pinning case).

baszalmstra: This is the size of all shards of the noarch conda-forge channel (with the above edit applied):

```
|                      | size    |
| -------------------- | ------- |
| repodata.json.zst    | 13.5 MB |
| shards/*.json.zst    | 18   MB |
| shards/*.msgpack.zst | 17.1 MB |
```

wolfv:
One other thing: 

`msgpack2json -d -b -i /Users/wolfv/Library/Caches/rattler/cache/repodata/shards-v1/7336fbc738626810350ff13c195012deb91b794840c81f59ffd8a74559102ec3.msg` works nowadays for looking at the shards in JSON :)

jezdez:
> Note that in fairness we never got around to implement the faster "2-file" JLAP variant. However, the sharded repodata is a lot "simpler" in terms of updates, so we believe that in `rattler` this will supersede the JLAP-efforts.

Thinking about this a little more, I think your statement that this proposal is "simpler" needs some elaboration, "simpler" in what areas? Do you mean "faster"?

In my experience, "simplicity" is easier said than done (read: is hard) and given that this introduces a whole new format, uses a different encoding type and seems to be focused on a non-standard hosting type (OCI, from reading your comments), I'm hoping you can provide context about this. It's.. a lot for one CEP, while trying to imagine a way to implement this in conda (which remains a goal with these CEPs after all).

I would appreciate hearing your thoughts on how a rollout would look like for existing conda hosting (e.g. conda-forge) and how current conda users would benefit from this.

wolfv:
Note that in fairness we never got around to implement the faster "2-file" JLAP variant. However, the sharded repodata is a lot "simpler" in terms of updates, so we believe that in rattler this will supersede the JLAP-efforts.

> Thinking about this a little more, I think your statement that this proposal is "simpler" needs some elaboration, "simpler" in what areas? Do you mean "faster"?

Simpler in that no "patching" is needed (you get the real repodata right away). There is also no "state" that needs to be kept - just update the index file and make sure that all the shards are there, which is really straightforward.

It should be faster, too, because, again, we don't need the whole index in the first place. So the download is only ~1 Mb (for linux + noarch). I can download all shards for `jupyterlab` in 350 ms.

> In my experience, "simplicity" is easier said than done (read: is hard) and given that this introduces a whole new format, uses a different encoding type and seems to be focused on a non-standard hosting type (OCI, from reading your comments), I'm hoping you can provide context about this. It's.. a lot for one CEP, while trying to imagine a way to implement this in conda (which remains a goal with these CEPs after all).

We're not at all focused on a non-standard hosting type - we just want to make sure that that works 100% fine. In fact, we host our "repodata-shards" on a traditional S3-style bucket, but a regular file server would work just as well. One just needs to add the `repodata_shards.msgpack.zst` file and all the shards under `/shards/...`.

I don't think there is anything to worry about with regards to the implementation in conda. We are using `msgpack-python` from Python to create the index (the API is the same as `json.load`/`json.dumps`).

> I would appreciate hearing your thoughts on how a rollout would look like for existing conda hosting (e.g. conda-forge) and how current conda users would benefit from this.

You could do it the same way as we do, and host an alternative index on a different subdomain for a testing period? What we do is that our Github action is ingesting existing repodata, creates shards, and stores the (new) shards on the bucket. It's all open source (in the linked repository). So you can try it out today! :)

With the `base_url` in the `repodata_shards.msgpack.zst` file we link back to the "original" conda-forge, by the way (so that packages would be downloaded from the official `conda.anaconda.org`.

cep-16.md
```diff
@@ -0,0 +1,254 @@
...
+      "md5": "ba445bf767ae6f0d959ff2b40c20912b",
+      "name": "rich",
+      "noarch": "python",
+      "sha256": "2b26d58aa59e46f933c3126367348651b0dab6e0bf88014e857415bb184a4667",
+      "size": 184347,
+      "subdir": "noarch",
+      "timestamp": 1709150578093,
+      "version": "13.7.1"
+    }
+  }
```
wolfv, line 138: We are adding a `removed` key here for backwards compatibility.

wolfv:
I just tested the sharded repodata vs regular repodata on a slow internet connection. It took 15s for noarch + osx-arm64 in the traditional case, and 900ms for the sharded repodata. Very nice!

wolfv:
I also wanted to point out that this is an infrastructure improvement that should dramatically reduce indexing times and thereby hopefully pave the way to an almost realtime experience when publishing new packages. Especially when we start using the `daily` and `weekly` files.

The changes will be much more minimal and small. I hope you guys are also excited about that prospect :) @jezdez

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
+
+Because the file is relatively big this means that often a large `max-age` is used for caching which means it takes more time to propagate new packages through the ecosystem.
+
+## Proposal
+
+We propose a "sharded" repodata format. It works by splitting the repodata into multiple files (one per package name) and recursively fetching the "shards".
+
+The shards are stored by the hash of their content (e.g. "content-addressable"). 
+That means that the URL of the shard is derived from the content of the shard. This allows for efficient caching and deduplication of shards on the client. Because the files are content-addressable no round-trip to the server is required to check fressness of individual shards.
```
jezdez, line 45:
```suggestion
That means that the URL of the shard is derived from the content of the shard. This allows for efficient caching and deduplication of shards on the client. Because the files are content-addressable no round-trip to the server is required to check freshness of individual shards.
```

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+## Motivation
+
+The current repodata format is a JSON file that contains all the packages in a given channel. Unfortunately, that means it grows with the number of packages in the channel. This is a problem for large channels like conda-forge, which has over 150,000+ packages. It becomes very slow to fetch, parse and update the repodata.
+
+## Design goals
+
+1. **Speed**: Fetching repodata MUST be very fast. Both in the hot- and cold-cache case.
+2. **Easy to update**: The channel MUST be very easy to update when new packages become available.
+3. **CDN friendly**: A CDN MUST be usable to cache the majority of the data. This reduces the operating cost of a channel.
+4. **Support authN and authZ**: It MUST be possible to implement authentication and authorization with little extra overhead.
```
jezdez, line 14: That seems like an odd duck, authorization shouldn't be attached to the repodata format as it's part of the transport layer, not the application layer.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+## Proposal
+
+We propose a "sharded" repodata format. It works by splitting the repodata into multiple files (one per package name) and recursively fetching the "shards".
+
+The shards are stored by the hash of their content (e.g. "content-addressable"). 
+That means that the URL of the shard is derived from the content of the shard. This allows for efficient caching and deduplication of shards on the client. Because the files are content-addressable no round-trip to the server is required to check fressness of individual shards.
+
+Additionally an index file stores the mapping from package name to shard hash.
+
+Although not explicitly required the server SHOULD support HTTP/2 to reduce the overhead of doing a massive number of requests.
```
jezdez, line 49: Given the sheer number of files, I think realistically this is a MUST.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+    "...": "other packages"
+  }
+}
+```
+
+The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
+
+For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
+
+We suggest serving the file with a short lived `Cache-Control` `max-age` header of 60 seconds to an hour but we leave it up to the channel administrator to set a value that works for that channel.
```
jezdez, line 78: Why do you suggest 60 seconds TTL?

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Repodata shard
+
+Individual shards are stored under the URL `<shard_base_url>/<subdir>/shards/<sha256>.msgpack.zst`. Where the `sha256` is the lower-case hex representation of the bytes from the index. It is a zstandard compressed msgpack file that contains the metadata of the package. The `<shard_base_url>` is defined in [Authentication](#authentication).
+
+The files are content-addressable which makes them ideal to be served through a CDN. They SHOULD be served with `Cache-Control: immutable` header.
```
jezdez, line 84: Why SHOULD and not MUST?

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+  "removed": [
+    "rich-10.15.1-pyhd8ed1ab_1.tar.bz2"
+  ]
+}
+```
+
+The `sha256` and `md5` from the original repodata fields are converted from their hex representation to bytes. 
+This is done to reduce the overall file size of the shards.
+
+Implementers SHOULD ignore unknown keys, this allows adding additional keys to the format in the future without breaking old versions of tools.
```
jezdez, line 151: [CEP 15](https://github.com/conda/ceps/blob/main/cep-15.md) adds a `repodata_version` field incrementing it to `2` when a base_url is present. Since you're willing to update a number of fields, this should be incremented to `3`, so clients don't have to implement forever backwards-compatible repodata loaders.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remaing very small, e.g. 100s of bytes to a couple of kilobytes.
```
jezdez, line 153: ❤️

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+}
+```
+
+`shard_base_url` is an optional url to use as the base url for the `repodata_shards.msgpack.zst` and the individual shards. If the field is not specified it should default to `<channel>/<subdir>`.
+
+`token` is an optional field that if set MUST be added to any subsequent request in the `Authentication` header as `Authentication: Bearer <token>`. If the `token` field is not set sending the `Authentication` header is also not required. 
+
+The optional `issued_at` and `expires_in` fields can be used to verify the freshness of a token. If the fields are not present a client can assume that the token is valid for any subsequent request.
+
+For a simple implementor this endpoint could just be a static file with `{}` as the content.
```
jezdez, line 174: As stated above, authnz is a transport layer thing and doesn't fit into this application layer metadata standard. This should be moved into an own CEP.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+## Fetch process
+
+To fetch all needed package records, the client should implement the following steps:
+
+1. Acquire a token (see: [Authentication](#authentication)). Acquiring a token can be done lazily as to only request a token when an actual network request is performed.
```
jezdez, line 180: Authzn isn't part of the repodata standard and should be evolved in a separate CEP.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+### Remove redundant keys
+
+`platform` and `arch` can be removed because these can be inferred from `subdir`.
+
+### Integrating additional data
+
+With the total size of the repodata reduced it becomes feasible to add additional fields directly to the repodata records. Examples are:
+
+- add `purl` as a list of strings (Package URLs to reference to original source of the package) (See: https://github.com/conda-incubator/ceps/pull/63)
+- add `run_exports` as a list of strings (run-exports needed to build the package) (See: https://github.com/conda-incubator/ceps/pull/51)
```
jezdez, line 245: How does this related to [CEP-15](https://github.com/conda/ceps/blob/main/cep-15.md)?

cep-16.md
baszalmstra: We have been running this with HTTP/1 by accident for a while, its definitely slower but is still very acceptable. I left this as a should because everything still works perfectly fine without HTTP/2.

cep-16.md
baszalmstra: I feel like this provides a sane tradeoff between limiting the number of requests to the server while also reducing the time it takes to get new packages. But TBH it's quite random.

cep-16.md
baszalmstra: An implementation doesnt care about Cache-Control headers, it just uses the content-address. However, for CDNs having the immutable header ensures that the content is cached on the edge for as long as possible. But it is not a requirement.

cep-16.md
baszalmstra: The downside of that though is that a client that doesn't support version 3 will not be able to parse the repodata, effectively breaking it. Adding keys is forward-compatible as long as the client is not strict about unknown keys.

cep-16.md
baszalmstra: Im not sure what you mean? The `base_url` from CEP-15 is part of the index file.

cep-16.md
baszalmstra: Or did I misunderstand and is the `repodata_version` field forward-compatible?

cep-16.md
baszalmstra: Im not sure why this CEP cannot make changes to both? Authenticating requests can add serious overhead to the server-side implementation which in turn affects the performance on the client. Having a way to effectively "cache" the authentication drastically reduces the work on the server which given the shear number of requests introduced by this CEP I think makes sense to address as well. Note that the approach taken here is very similar to how OCI registries operate.

wolfv: I would also argue that we're proposing more of a transport protocol here where the authentication fits in pretty well as a key part of the whole thing.

wolfv: Lastly, you can also use other (more traditional) means by not replying with a token on the endpoint and use traditional means instead (depending on how your authentication works, it might be slower).

cep-16.md
dholth: Does pip do something similar with the simple index, requesting one per package name

baszalmstra: Yes there is one index per package name e.g. https://pypi.org/simple/pinject .

But this index does not store any dependency information for any of the artifacts. An additional request is required PER artifact to get the dependecy information. 

In this proposal we add the dependecy information of individual packages directly to the per package name shard.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+### JLAP
+
+In a previously proposed CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
+With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
+
+However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
+
+JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. This is often the case for CI runners.
+
+Finally, the implementation of JLAP is quite complex which makes it hard to adopt for implementers.
```
dholth, line 30: IMO the complexity is the least fair criticism of JLAP, it can yield a small implementation.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+7. **Bandwidth optimized**: Any data that is transferred SHOULD be as small as possible.
+
+## Previous work
+
+### JLAP
+
+In a previously proposed CEP, [JLAP](https://github.com/conda-incubator/ceps/pull/20) was introduced. 
+With JLAP only the changes to an initially downloaded `repodata.json` file have to be downloaded which means the user drastically saves on bandwidth which in turn makes fetching repodata much faster.
+
+However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the shear amount of data involved.
```
dholth, line 26:
```suggestion
However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the sheer amount of data involved.
```

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+### Update optimization
+
+We could implement support for smaller index update files. This can be done by creating a rolling daily and weekly index update file that can be used instead of fetching the whole `repodata_shards.msgpack.zst` file. The update operation is very simple (just update the hashmap with the new entries).
+
+For this we propose to add the following two files:
+
+- `<channel>/<subdir>/repodata_shards_daily.msgpack.zst`
+- `<channel>/<subdir>/repodata_shards_weekly.msgpack.zst`
+
+They will contain the same format as the `repodata_shards.msgpack.zst` file but only contain the packages that have been updated in the last day or week respectively. The `created_at` field in the index file can be used to determine which file to fetch to make sure that the client has the latest information.
```
dholth, line 256: plus null if a shard has been removed?

cep-16.md
baszalmstra: IMHO in theory its pretty straightforward but to integrate properly there are a number of additional complexities that make this statememt true. To name a few: it requires maintaining extra state on disk, http range requests, we need to make sure that the repodata.json is still the previously stores repodata, indexing requires the previous state, and the problem that you need an additional overlay file to make it perform optimally.

But in fairness this is just my personal experience while implementing it.

wolfv: @dholth I think we can claim that we are the only ones shipping JLAP in production and it's been pretty rough, sadly. We could improve with the two file approach but I think we make a good point about why this proposal is simpler.

Also, one thing we gain with our proposal here is complete hash-integrity. Something that is not possible with the JLAP two-file approach. We want to enable signed repodata and packages (TUF style) sooner or later and I don't see a way to do that with JLAP's two-file approach.

With our approach, we can at any time prove the integrity of the whole thing.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+However, in practice patching the original repodata can be a very expensive operation, both in terms of memory and in terms of compute because of the sheer amount of data involved.
+
+JLAP also does not save anything with a cold cache because the initial repodata still has to be downloaded. This is often the case for CI runners.
+
+Finally, the implementation of JLAP is quite complex which makes it hard to adopt for implementers. 
+
+### ZSTD compression
+
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% smaller (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller.
```
dholth, line 34: 20% smaller than Content-Encoding: gzip?

cep-16.md
dholth: Is there a non-content-addressable variant that might be suitable for a package server with per-user, per-package permissions?

cep-16.md
baszalmstra:
```suggestion
They will contain the same format as the `repodata_shards.msgpack.zst` file but only contain the packages that have been updated in the last day or week respectively. `null` is used for keys that have been removed. The `created_at` field in the index file can be used to determine which file to fetch to make sure that the client has the latest information.
```

cep-16.md
baszalmstra:
```suggestion
A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% of the original size (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
```

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+### Integrating additional data
+
+With the total size of the repodata reduced it becomes feasible to add additional fields directly to the repodata records. Examples are:
+
+- add `purl` as a list of strings (Package URLs to reference to original source of the package) (See: https://github.com/conda-incubator/ceps/pull/63)
+- add `run_exports` as a list of strings (run-exports needed to build the package) (See: https://github.com/conda-incubator/ceps/pull/51)
+
+### Update optimization
+
+We could implement support for smaller index update files. This can be done by creating a rolling daily and weekly index update file that can be used instead of fetching the whole `repodata_shards.msgpack.zst` file. The update operation is very simple (just update the hashmap with the new entries).
```
dholth, line 249: This could be done with a jlap-like rolling checksum / changelog file where the client would look for the hash of their current index, and json merge patch instead of the more complicated json patch.

wolfv: Yeah I like the idea of having a hash of the "expected" file. We don't need any proper "patching" as it shoudl be as simple as 

```
old_repodata["shards"].update(new_repodata["shards"])
```

(with some extra handling of null keys potentially).

We would also store a timestamp in the `info` field that we could use to evaluate wether we can use the `daily` or `weekly` update.

dholth: "json merge patch" is similar to ".update()". We could simulate the size of the proposed file using https://github.com/dholth/conda-test-data, probably it is pretty small based on what I've observed about jlap.

wolfv: Yes, we are logging what percentage of shards changes every 6 hours in the indexing pipeline. It's usually < 1 percent so it's pretty small. 

I think adding json-merge-patch isn't going to add anything over the proposed `update` of dictionary keys - but the principle is definitely the same. We could also define a hash, but I would argue that we should do a "content" hash (ie. one that does not rely on the specific formatting of a "JSON" file. Although that might also be less of an issue since we're using `msgpack` for serialization in this proposal :)

dholth: `msgpack` may have more options for formatting than json, e.g. "serializers SHOULD use the format which represents the data in the smallest number of bytes." but does our implementation? JSON with sorted keys and no whitespace is predictable. But we won't think about implementing https://www.rfc-editor.org/rfc/rfc8785

cep-16.md
xhochy: I also feel that authorization should be split off. You can do this in multiple ways that are independent of the how shards are actually implemented.

cep-16.md
xhochy: I would also expect that HTTP/1 gives already a reasonable improvement. Not all corporate proxies (but most) support HTTP/2 yet and thus I'm always happy to have enhancements that can fall back to HTTP/1.

cep-16.md
xhochy: > Is there a non-content-addressable variant that might be suitable for a package server with per-user, per-package permissions?

Most permission-based solutions I have yet encountered (I have not seen Anaconda Enterprise) all managed permissions on the basis of whole channels. This would stay compatible with the current approach and was something that was also compatible with the old approach.

For giving access to a subset of packages in a channel, you would need to serve copies of `repodata.json` depending on the user. As the new approach (new = sharding) would still also serve the old `repodata.json`, I think this can only be supported with a backend that check permissions on the request basis and doesn't do authentication purely based on the URL.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+The `sha256` and `md5` from the original repodata fields are converted from their hex representation to bytes. 
+This is done to reduce the overall file size of the shards.
+
+Implementers SHOULD ignore unknown keys, this allows adding additional keys to the format in the future without breaking old versions of tools.
+
+Although these files can become relatively large (100s of kilobytes) typically for a large case (conda-forge) these files remaing very small, e.g. 100s of bytes to a couple of kilobytes.
+
+## <a id="authentication"></a>Authentication
```
xhochy, line 155: I'm -1 on adding this specific bit of authentication to the sharding proposal. We should be able to implement the sharding approach independently of the authentication mechanism. 

I don't see in some setups how we would be able to provide such an endpoint. This ties the CEP to a certain kind of authentication and removes the flexibility that we currently have in also using HTTP Basic Auth instead of token based auth.

xhochy (changes requested):
I really like sharding approach, but I'm very heavy -1 on mixing in authentication into it. This will conflict with existing implementations and thus will make the feature not usable there.

cep-16.md
baszalmstra: I am open to opening another CEP specifically for authentication of the sharded index but let me first explain the reasoning behind why we added this.

The reason we added this is that we have noticed that adding support for generic authentication methods provides a very significant overhead to fetching the individual shards and a large infrastructure cost. 

Take as an example our prefix.dev backend. When requesting private repodata a user requests `https://prefix.dev/mychannel/win-64/repodata.json`, our backend server authenticates the request using either a conda token, basic auth or an API key and redirects the request to a presigned-url. If we would do something similar for the shards each request would have to go through this dynamic end-point. This adds additional delay to the request and incurs an infrastructure cost due to the required compute. And we are talking about hundreds to thousands of requests when a user fetching repodata. We have noticed that this significantly impacts the performance.

Ideally, we want to have a solution where the "presigning" happens just once and we can reuse this information on subsequent requests.

I am open to alternative solutions, but I think being able to authenticate once for all subsequent requests is very important to how effective this proposal is. Standardizing this method is also important to ensure that every client will be able to access different server implementations. 

Then with regards to our specific solution using the `token` endpoint. The `token` endpoint only really needs to be dynamic when the authentication mechanism used is dynamic. In any other case, the token end-point could just be an empty static file. Basic authentication or even conda-token-based authentication can still be applied to both the individual shards and the shard index.

As an alternative though, we could also look at returning a set of headers from the initial request to the shard index instead of requesting a separate file.

FYI @jezdez

wolfv: I think what we are not making clear in the CEP is that "traditional" auth methods should continue to work just as well. For example, if the server uses basic HTTP authentication or conda token authentication, this should continue to work (and depending on the server implementation should also not negatively affect performance).

We are just interested in adding a mechanism to the protocol in order to facilitate really fast performance for large public channels.

I think we can make this an "add-on" behavior when the request to `repodata_shards.msgpack.zst` returns specific header values such as:

```
X-Shard-BaseUrl: https://fast.prefix.dev/conda-forge/
X-Shard-Token: pfx_secret_token
X-Shard-Token-ExpiresAt: TIMESTAMP
```

The client would then use these values for subsequent requests. If these headers are not returned, it will just use the "regular" URL with regular authentication.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+A notable improvement is compressing the `repodata.json` with `zst` and serving that file. In practice this yields a file that is 20% of the original size (20-30 Mb for large cases). Although this is still quite a big file its substantially smaller. 
+
+However, the file still contains all repodata in the channel. This means the file needs to be redownloaded every time anyone adds a single package (even if a user doesnt need that package).
+
+Because the file is relatively big this means that often a large `max-age` is used for caching which means it takes more time to propagate new packages through the ecosystem.
+
+## Proposal
+
+We propose a "sharded" repodata format. It works by splitting the repodata into multiple files (one per package name) and recursively fetching the "shards".
```
aovasylenko, line 42: I reminds me about current `simple` pypi index, which (in case all python ecosystem of 500K+ packages leads to this number of shards)
Were there any ideas about evaluating smaller number of shards, for example by having prefix-based rule or range?
My worry that based on server-side implementation initial pulling of this information will generate massive number of requests (even with using http2) on API/DB level each request could be processed separately.

wolfv: We haven't really discussed different sharding strategies. However, our implementation uses a CDN + static files and that seems to work relatively well. It also seems to work well enough for PyPI and crates.io so I don't think we're particularly concerned.

cep-16.md
```diff
@@ -0,0 +1,260 @@
...
+
+The index is still updated regularly but the file does not increase in size with every package added, only when new package names are added which happens much less often.
+
+For a large case (conda-forge linux-64) this files is 670kb at the time of writing.
+
+We suggest serving the file with a short lived `Cache-Control` `max-age` header of 60 seconds to an hour but we leave it up to the channel administrator to set a value that works for that channel.
+
+### Repodata shard
+
+Individual shards are stored under the URL `<shard_base_url>/<subdir>/shards/<sha256>.msgpack.zst`. Where the `sha256` is the lower-case hex representation of the bytes from the index. It is a zstandard compressed msgpack file that contains the metadata of the package. The `<shard_base_url>` is defined in [Authentication](#authentication).
```
dholth, line 82: What if we got rid of the implicit /shards/ subdir?

wolfv: Any reasons for getting rid of it?

wolfv: There are some benefits I can think of:

- makes it easy for the backend to redirect traffic to another bucket or other rewrite rules
- separates the data, makes it easy to purge the whole `shards/` folder if necessary
- doesn't clutter the folder with the packages

But I would be happy to hear your reasons for removing it.

dholth: It's something else to look up or remember and we have some upload code that assumes everything for a subdir is in the same folder. Should the shards folder be part of the index file?

According to google some people call .msgpack .mpk but others don't want to advertise "generic msgpack" as a thing and instead use an application extension.

cep-16.md
dholth: [This code plays with the codec and compression.](https://github.com/dholth/conda-test-data/blob/main/test_change_strategy.py#L190-L193)

cep-16.md
wolfv: @jezdez was asking to not optimize for a specific way of distributing these things :) some people might still run on traditional filesystems ...

dholth: I could mention splitting the hash into pairs and having a tree of ab/cd/ef/abc123.zst
But we probably have to assume that our filesystems are OK with large numbers of files in the same folder since the packages are there.

dholth: Is there a reason not to include "shard_base_url": "https://shards.prefix.dev/conda-forge/<subdir>/shards/" (including a now-explicit /shards/ subdirectory) in the index file?

baszalmstra: Nope, I think that makes more sense indeed.

cep-16.md
dholth: What if we allow a shard to contain the metadata of more than one package, so that multiple package names could point to the same shard? It is a nuisance to download many 300-byte shards.

wolfv: Yeah we could, but it comes with some downsides:

- cache might not be as efficient as multiple packages might change at different frequencies and cache becomes invalid more often
- it's more complicated on the implementation as packages don't map 1-1 to shards, but that would not be too bad.

We could keep the current repodata_shards index structure and use the same hash for multiple package names (that would probably compress decently). 

However, I am not sure if the tradeoff is worth it. 

Also it will start to look a bit like `zchunk` then :)

dholth: I tried [using a buffer to combine sequential (ordered by package name) shards](https://github.com/dholth/conda-test-data/commit/d5e4c74aae07125055640a537473e8116fd4454a) when the combined size is not > 8192 bytes compressed. So we still wind up with plenty of ~300 byte single package shards if it sees (large shard, small shard, large shard) but it's all in order.

That scheme gives 7987 files and 10883 package names on a conda-forge/linux-64 snapshot, probably taking 75% of the time to download if we happen to want to download everything. Unlike zchunk we have the option of not maintaining a complete local copy of the data.

The most packages combined into a single shard are these fifteen.

```
['perl-bit-vector-7.4-pl5321h166bdaf_0.tar.bz2',
 'perl-capture-tiny-0.48-pl5321ha770c72_1.tar.bz2',
 'perl-carp-assert-0.21-pl526_0.tar.bz2',
 'perl-class-method-modifiers-2.13-pl5321ha770c72_0.tar.bz2',
 'perl-clone-0.46-pl5321h166bdaf_0.tar.bz2',
 'perl-compress-raw-bzip2-2.201-pl5321h166bdaf_0.tar.bz2',
 'perl-compress-raw-zlib-2.202-pl5321h166bdaf_0.tar.bz2',
 'perl-data-dumper-2.183-pl5321h166bdaf_0.tar.bz2',
 'perl-b-hooks-endofscope-0.26-pl5321ha770c72_0.conda',
 'perl-b-hooks-endofscope-0.28-pl5321ha770c72_0.conda',
 'perl-class-data-inheritable-0.09-pl5321ha770c72_0.conda',
 'perl-class-load-0.25-pl5321ha770c72_0.conda',
 'perl-class-load-xs-0.10-pl5321h0b41bf4_0.conda',
 'perl-class-method-modifiers-2.14-pl5321ha770c72_0.conda',
 'perl-class-method-modifiers-2.15-pl5321ha770c72_0.conda',
 'perl-cpan-meta-check-0.014-pl5321ha770c72_0.conda',
 'perl-data-dumper-2.183-pl5321hd590300_0.conda',
 'perl-data-dumper-concise-2.023-pl5321ha770c72_0.conda',
 'perl-data-optlist-0.112-pl5321ha770c72_0.conda',
 'perl-data-optlist-0.113-pl5321ha770c72_0.conda',
 'perl-data-optlist-0.114-pl5321ha770c72_0.conda']
```

cep-16.md
dholth: I tried doing an http/1 fetch of all ~10k files from the prefix test server using a 20-thread (and 10 connections at once pool) with Python requests. It takes about 20 seconds to fetch everything with a hot cache, and the spec is designed for us to fetch much less than everything.

cep-16.md
wolfv: I am wondering, if you need everything, wouldn't you rather download the `repodata.json` file instead? I think for us that wasn't the use case we optimized for.

dholth: I'm imagining a situation in which there is only repodata_shards. Suppose we are in the middle of processing a bunch of queries on the SubdirData and suddenly we encounter a query for "any package name, md5=x". Are we more likely to reject the query, download everything, or eject to downloading an alternate format?

wolfv: Why not deprecate it, and grab the repodata.json in the meantime? Tell people to use proper lockfiles? :)

dholth: What if we simplified the implementation, and included one repodata shard with all the packages in it, and an index that pointed to that singular shard 10k times 🤣

cep-16.md
baszalmstra: Have you drawn any conclusions from this that we should incorporate @dholth ?

cep-16.md
baszalmstra: I removed authentication from the CEP and marked it as something we should investigate in another CEP. I still left it in the design goals because I still think we shouldnt implement something that would rule this out.

cep-16.md
baszalmstra: I added this to the CEP, implemented it [in our mirror script](https://github.com/prefix-dev/sharded-repo/pull/6), and added [this to rattler](https://github.com/mamba-org/rattler/pull/747).

baszalmstra:
@jezdez @xhochy @dholth @wolfv I updated the CEP to:

* Add `shards_base_url`
* Remove authentication flow from the CEP (we will address this in a future CEP).

Please give it another read, I would like to put this to a vote soon.

xhochy (approved):
Fine with me now.

jezdez:
@conda/steering-council

baszalmstra:
yes

wolfv:
Please use the following form to vote!

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [x]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@kkraus14 (Keith Kraus)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [x]  yes
- [ ]  no
- [ ]  abstain

cep-16.md
mariusvniekerk: Another form of permission for content-addressed blob store systems I've seen is to store some additional tag style header attributes with the content blob and use that to drive authz.  gcp and aws both support setting this kind of metadata and you can perform the auth part with a proxy in front of the blob store

jaimergp (approved):
Just a few typos I found while reading through the whole thing.

jakirkham:
Weird. Had checked the box above a few days ago and it appears to have disappeared. Rechecked it

wolfv:
@jakirkham strange indeed. I can see your edit in the edit history, but it didn't seem to "tick" the box...

wolfv:
@chenghlee @CJ-Wright @mbargull last chance to vote :)

mbargull:
I am thankful that someone is working on the repodata implementation since the current one (one big blob per channel/subdir) became unwieldy a good while ago (at least in resource-constrained environments).
As far as I can tell, this proposal is a good starting point but still needs refinement (which I expect to get clearer once more people test it out in rattler and conda-index, etc.).
Unfortunately, I was not able to take the time to do a proper review and consult in the refine process here and as such opted to abstain.

PR #81, opened 2024-06-04
[CEP 24] Standardize `environment.yml`

jaimergp:
Standardizing the popular `environment.yml` input files.

📝👓 [Markdown preview](https://github.com/jaimergp/ceps/blob/environment-yml/cep-XXXX.md)


cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+dependencies:
+  - numpy
+```
+
+With a name:
+
+```yaml
+name: test
+dependencies:
+  - numpy >=1.10
```
skupr-anaconda, line 89: What about conditions if a dependency is used on different platforms, python versions, implementations, etc.?

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+## Nomenclature
+
+The `environment.yml` file format was introduced by `conda env`. It has been traditionally referred to as "the YAML input file", but other YAML files exist in the ecosystem now. To avoid confusion, in this CEP we will only use the term `environment.yml`. Do note that the actual filename of an input file using this format might be a different one (e.g. `env.yml`).
+
+## Specification
+
+`environment.yml` files are YAML documents that encode the necessary information to create a new conda environment. They can contain up to five top-level keys:
+
+- `name`: The preferred name for the environment.
+- `prefix`: The preferred full path for the environment. Often ignored.
```
marcoesters, line 28:
```suggestion
- `prefix`: The preferred full path to the environment. Often ignored.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+- `dependencies`: List of package specifications that MUST be installed in the new environment.
+- `channels`: conda channels that will be used to solve the dependencies
```
marcoesters, line 30:
```suggestion
- `channels`: conda channels that will be used to resolve the dependencies.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+The file name is not standardized, but the extension MUST be `yml` or `yaml`.
+
+### `name` and `prefix`
+
+Optional, `str`.
+
+Both fields refer to the _preferred_ name or path for the newly created environment. Tools SHOULD allow these suggestions to be overridden by the user with additional CLI flags or equivalent. If the proposed environment path exists, tools MUST NOT overwrite silently by default.
+
+Special names `base` and `root` SHOULD not be accepted. Prefixes targetting protected system paths SHOULD be rejected. Paths can contain tildes (`~`) and environment variables, and they MUST be expanded when present.
```
marcoesters, line 43:
```suggestion
Special names `base` and `root` SHOULD NOT be accepted. Prefixes targetting protected system paths SHOULD be rejected. Paths can contain tildes (`~`) and environment variables, and they MUST be expanded when present.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+The name of the environment (and the last component of the expanded`prefix` path) MUST NOT contain any of these characters: `/`, ` `, `:`, `#`.
```
marcoesters, line 45:
```suggestion
The name of the environment (and the last component of the expanded `prefix` path) MUST NOT contain any of these characters: `/`, ` `, `:`, `#`.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+### `dependencies`
+
+Required, `list[str | dict[str, Any]]`.
+
+The simplest form for this section is a list of `str` encoding `MatchSpec`-compatible requirements. These items MUST be processed as conda requirements and submitted to the solver (along with `channels`) to obtain the contents for the resulting conda environment.
+
+This section can also contain "subsection" dictionaries that map `str` to arbitrary values. Each key refers to a non-conda installer tool that will process the associated contents as necessary. The additional subsections MUST be processed after the conda requirements. They SHOULD only add new contents. These keys SHOULD NOT overwrite existing contents.
+
+Currently known subsections include `pip`, the contents of which encode a list of `str` referring to PyPI requirements. How to process this list is left as an implementation detail, but common approaches involve invoking the `pip` command-line directly.
```
marcoesters, line 55:
```suggestion
Currently known subsections include `pip`, the contents of which encode a list of `str` referring to PyPI requirements. How this list is processed is left as an implementation detail, but common approaches involve invoking the `pip` command-line directly.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+Additional subsections are allowed. The conda client MUST error out if it cannot process unknown sections.
+
+### `channels`
+
+Optional, `list[str]`.
+
+These are the conda channels that will be queried to solve the requirements added in `dependencies`. They can be expressed with names, URLs and local paths.
+
+If not specified, the conda client MUST use the default configuration. When specified, these channels MUST be used to populate the channel list passed to the solver. Then default channel configuration MUST be appended, unless the special name `nodefaults` is present. When this is the case, the default configuration MUST no be appended. The `nodefaults` name MUST not be passed to the solver.
```
marcoesters, line 65:
```suggestion
If not specified, the conda client MUST use the default configuration. When specified, these channels MUST be used to populate the channel list passed to the solver. Then default channel configuration MUST be appended, unless the special name `nodefaults` is present. When this is the case, the default configuration MUST no be appended. The `nodefaults` name MUST NOT be passed to the solver.
```

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+Special names `base` and `root` SHOULD not be accepted. Prefixes targetting protected system paths SHOULD be rejected. Paths can contain tildes (`~`) and environment variables, and they MUST be expanded when present.
+
+The name of the environment (and the last component of the expanded`prefix` path) MUST NOT contain any of these characters: `/`, ` `, `:`, `#`.
+
+### `dependencies`
+
+Required, `list[str | dict[str, Any]]`.
+
+The simplest form for this section is a list of `str` encoding `MatchSpec`-compatible requirements. These items MUST be processed as conda requirements and submitted to the solver (along with `channels`) to obtain the contents for the resulting conda environment.
```
marcoesters, line 51: Should platform selectors be mentioned here or is that implied with `MatchSpec`-compatible?

marcoesters (commented):
Just a few suggestions (no blockers if you don't agree though) and a question about platform selectors.

cep-??.md
jaimergp: AFAIK, `environment.yml` doesn't support platform selectors like `meta.yaml`, doesn't it?

baszalmstra (approved):

cep-??.md
```diff
@@ -0,0 +1,134 @@
...
+
+`environment.yml` files are YAML documents that encode the necessary information to create a new conda environment. They can contain up to five top-level keys:
+
+- `name`: The preferred name for the environment.
+- `prefix`: The preferred full path to the environment. Often ignored.
+- `dependencies`: List of package specifications that MUST be installed in the new environment.
+- `channels`: conda channels that will be used to resolve the dependencies.
+- `variables`: Environment variables that SHOULD be added to the `conda-meta/state` file in the resulting environment.
+
+Additional sections can be present, but they MUST be ignored and the user SHOULD receive an informative warning about them.
```
beckermr, line 33: conda-lock cannot abide by this CEP because it includes a `platforms` key and one other. Any ideas on how to make that work?

jaimergp: Thanks! I added `category` and `platforms` as optional values now.

beckermr:
@jaimergp We need to go through the conda-lock docs to ensure we are not excluding features conda-lock uses. For example, selectors are supported by conda-lock right now: https://conda.github.io/conda-lock/src_environment_yml/#preprocessing-selectors

jaimergp:
Oof. I really don't want to open that can of worms. Best compromise I can think of is documenting it as a series of `SHOULD NOT`, or `MAY` at best.

jaimergp:
> We may not like selectors, however.

Precisely 😬 

> Why write in an optional feature to have recipe v1-style if...then...else mappings and then allow conda-lock to implement that later?

Yes, I want to do that for v2.

jaimergp:
`micromamba` also has (or had?) selectors, IIRC :/

beckermr:
I don't see a reason to wait for another CEP round though. We can add the optional feature now and move ahead. Otherwise it is just more voting and writing, all of which cost us time.

beckermr:
Ah I missed the point of this PR. The CEP claims to document what we have. The title of the PR threw me off. We should document that some tools use selectors for sure. Seems like the "MAY use selectors with platform or OS specific information" is the way to go then.

beckermr:
What else do you want to add to v2? I'd be ok doing both at the same time in a single CEP.

jaimergp:
Added a few sections which hopefully address your concerns.

jaimergp:
> What else do you want to add to v2? I'd be ok doing both at the same time in a single CEP.

A few more things. You can see some ideas in my [draft in this gist](https://gist.github.com/jaimergp/4209c4c90d51b1bb07fe7293095f7c70). We still need to prototype things in `conda` to see what else we need there.

cep-XXXX.md
```diff
@@ -0,0 +1,267 @@
...
+- `dependencies`: List of package specifications that MUST be installed in the new environment.
+- `channels`: conda channels that will be used to resolve the dependencies.
+- `variables`: Environment variables that SHOULD be added to the `conda-meta/state` file in the
+  resulting environment.
+
+The file name is not standardized, but the extension MUST be `yml` or `yaml`.
+
+Notably, this version of the specification does not feature a `version` (or equivalent) top-level
+key. Future revisions MAY include such a field. As such, the version information of this revision
+is omitted and considered to be `1`.
```
jaimergp, line 45: The community has talked about a [v2](https://github.com/conda/conda/issues/11341) format for a while, so I'm calling this one v1, but open for suggestions.

cep-XXXX.md
```diff
@@ -24,21 +24,37 @@ The `environment.yml` file format was introduced by `conda env`. It has been tra
...
 
-Additional sections can be present, but they MUST be ignored and the user SHOULD receive an informative warning about them.
+Preprocessing selectors MAY be available to specify line filters to allow platform-specific values,
+among other uses. There are two types: comment-based and dictionary-based. Only one type of filter
+SHOULD be used in the same document.
 
-The file name is not standardized, but the extension MUST be `yml` or `yaml`.
+### Other top-level keys
+
+Additional top-level keys can be present, but they MUST be ignored and the user SHOULD receive an informative warning about them when the conda client does not support them.
```
beckermr: Do we mean to explicitly forbid any more extensions to the v0 format?

cep-XXXX.md
```diff
@@ -88,13 +104,46 @@ valid `subdirs`, with one exception: `noarch` MUST be rejected as a valid platfo
...
+The expressions MUST be composed of one or more boolean variables, and can be chained with `and`
+and `or` logic operators. Precedence can be specified with parentheses.
+
+The supported boolean variables are the OS- and architecture-specific values collected in
+[conda-build's 25.1.x documentation][selectors], plus `osx64` (equivalent to `osx and x64`). More
+specifically,`py`, `py*`, `np` and `build_platform` are NOT supported.
+
+These are used by `conda-lock`.
+
+#### Dictionary-based selectors
```
beckermr, line 168: These are used by micromamba?

cep-XXXX.md
jaimergp: If they are, they will be non-standard (as the currently listed extensions "were"). I really hope no one needs more from this til-now unspecified format.

cep-XXXX.md
jaimergp: Yes, only docs about it are in https://github.com/mamba-org/provision-with-micromamba#extra-specs, but you can also see it in the code https://github.com/mamba-org/mamba/blob/26785a6b50bea434cf8b7d964f92f0f407b34b0a/libmamba/tests/src/core/test_env_file_reading.cpp#L18. Let me add a note.

cep-??.md
jaimergp: Well, in the end I did address these. Resolving this thread.

cep-??.md
jaimergp: Added a section about these now because they are used in conda-lock.

beckermr:
Also please call the existing format `v0`. Otherwise we are going to really confuse ourselves down the road. 

I wrote another CEP to forbid calling an existing format `v1`.

https://github.com/conda/ceps/pull/119

cep-XXXX.md
beckermr: This is going to confuse everything since recipe formats go from v0 to v1. Let's call this one `v0`.

beckermr:
That gist is cool. How much does this look like a pixi lock file and should we be trying to aim for a single format?

jaimergp:
> How much does this look like a pixi lock file and should we be trying to aim for a single format?

If you mean `pixi.toml` (declarative input file, not the resulting output lockfile), yes, I am borrowing some ideas from their design (groups, for example). However, I don't think we can fit all the use cases of `pixi.toml` in `environment.[v2.]yml`. Per-project features like `tasks`, or Python specific ones like `pypi-dependencies` are not a goal of the revisions I have in mind.

Also, even if we decided to create a CEP for `pixi.toml`'s schema, what will happen when Pixi keeps inevitably adding features to the file format. It will drift at some point.

My vision for the new environment.yml is to provide a reasonable standard that supports common use cases in the conda ecosystem that folks have been asking for a long time, but it should remain application-agnostic whenever possible and thus serve as a good interoperability format across clients.

cep-XXXX.md
jaimergp: I've thought more about this and after finding some examples, it sounds like `v0` for `meta.yaml` is more the exception, not the norm. See rationale in https://github.com/conda/ceps/pull/119#issuecomment-2732831737 and following comments.

beckermr:
That makes a ton of sense. This matches well onto the scope for conda-lock, which is focusing on a generic, non-project focused locking tool.

jaimergp:
@conda/steering-council, I'd like to call a vote for this CEP in two weeks (Wednesday, Apr 2nd 2025). Please let me know if there are any comments in the meantime. Thank you!

beckermr (approved):

cep-XXXX.md
```diff
@@ -0,0 +1,291 @@
...
+
+Optional, `str`.
+
+Both fields refer to the _preferred_ name or path for the newly created environment. Tools SHOULD
+allow these suggestions to be overridden by the user with additional CLI flags or equivalent. If
+the proposed environment path exists, tools MUST NOT overwrite silently by default.
+
+Special names `base` and `root` SHOULD NOT be accepted. Prefixes targeting protected system paths
+SHOULD be rejected. Paths can contain tildes (`~`) and environment variables, and they MUST be
+expanded when present.
```
chenghlee, line 78: > Paths can contain tildes (`~`) and environment variables, and they MUST be
expanded when present.

Should we clarify this statement and say `~` MUST BE expanded only when it occurs as the first character? Most *nix shells won't expand, e.g., `/tmp/~/subdir-a`.

marcelotrevisani (approved):

beckermr (approved):

xhochy (approved):

cep-XXXX.md
```diff
@@ -0,0 +1,291 @@
...
+channels:
+  - conda-forge
+dependencies:
+  - python
+  - sel(win): pywin32
+```
+
+## Reference
+
+- [Allowed extensions](https://github.com/conda/conda/blob/24.5.0/conda/env/specs/yaml_file.py)
```
Hind-M, line 279: Is this referring to extensions as in `.yml` and `.yaml`?
In that case, maybe specify the [line](https://github.com/conda/conda/blob/24.5.0/conda/env/specs/yaml_file.py#L11) instead as in the following references below?

jaimergp:
@xhochy @CJ-Wright @mariusvniekerk @chenghlee @mbargull @jakirkham @baszalmstra, this is your mid-period reminder to vote (please mark the checkbox in the comment above for an easier count). Thanks!

xhochy:
Somehow it doesn't mark it for me on multiple tries. My approval is already here via the review.

baszalmstra:
One final question I have is: if this CEP is accepted will conda and other clients also have to support “Dictionary-based selectors”? That wasnt clear to me.

jaimergp:
> if this CEP is accepted will conda and other clients also have to support “Dictionary-based selectors”? 

No, they are "extensions", so while recognized in the CEP, they are not required.

jaimergp:

* Total possible voters: 16. 
* Valid votes: 13 / 16 = 81.25%.

Yes votes (12 / 92.30%):

1. @xhochy
1. @chenghlee
1. @ocefpaf
1. @marcelotrevisani
1. @msarahan
1. @jezdez
1. @wolfv
1. @jaimergp
1. @baszalmstra
1. @beckermr
1. @hind-m
1. @trallard

No votes (0 / 0.00%).

Abstain votes (1 / 7.70%):

1. @mbargull

Not voted (2):

1. @cj-wright
1. @mariusvniekerk
1. @jakirkham

Invalid votes (0).

13/16 votes = 81.25% meets the minimum quorum. 
12/13 YES votes meets the approval threshold (60%).

PR #84, opened 2024-06-17
CEP: test serialization and package content tests in the new recipe format

wolfv:
Rendered version can be found here: https://github.com/wolfv/ceps/blob/cep-recipe-tests/cep-recipe-tests.md

Supersedes https://github.com/conda/ceps/pull/57


jezdez:
@wolfv Have you included feedback from #57 in this? Should we ask reviewers of that to comment here again?

wolfv:
I've included most feedback but I should do another pass.

PR #86, opened 2024-07-01
CEP: support for ABI3 packages

isuruf:


cep-abi3.md
```diff
@@ -0,0 +1,245 @@
+# Support for abi3 python packages
+
+<table>
+<tr><td> Title </td><td> A new recipe format </td>
```
jaimergp, line 4: This looks like a leftover from the template

isuruf:
```suggestion
<tr><td> Title </td><td> Support for abi3 python packages </td>
```

cep-abi3.md
```diff
@@ -0,0 +1,245 @@
...
+```
+to imply a python version independent package and set `noarch: python`
+in `info/index.json`.
+
+### `package_metadata_version = 2`
+
+In this method, we require additional support from install tools.
+
+For `abi3` and `noarch: python` pacakges, we record the `entry_points` and
+the pure python files in `info/link,json`.
```
pavelzw, line 193:
```suggestion
For `abi3` and `noarch: python` packages, we record the `entry_points` and
the pure python files in `info/link.json`.
```

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+&nbsp;&nbsp;<strong>C3</strong>:
+  `A2, A3, A4` are applied.
+
+This is compatible with `conda/mamba/micromamba` install tools
+currently. This requires support from build tools to set `subdir: <platform>`
+only.
+
+In particular an option:
+```
+build:
```
wolfv, line 182: I think for V1 recipes we would use:

```yaml
build:
  python:
    version_independent: true/false
```

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+  python_version_independent: true
```
wshanks, line 183: Is `version_independent` better than `abi3`? It seems possibly misleading to me because a package can still set a minimum Python version and so be Python version-dependent.

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+  They have `noarch: python`.
+
+&nbsp;&nbsp;<strong>C3</strong>:
+  `A2, A3, A4` are applied.
+
+This is compatible with `conda/mamba/micromamba` install tools
+currently. This requires support from build tools to set `subdir: <platform>`
+only.
+
+In particular an option:
```
wshanks, line 180: Is it planned that the conda-forge tooling would check for this option and then set up Windows/Mac/Linux builds even though `noarch: python` is specified where it would otherwise only build on one arch?

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+### `package_metadata_version = 1`
+
+We require the following attributes in abi3 packages:
+
+&nbsp;&nbsp;<strong>C1</strong>:
+  They have `subdir: <platform>` where `<platform>` is the subdir
+  that the package was built for.
+
+&nbsp;&nbsp;<strong>C2</strong>:
+  They have `noarch: python`.
```
wshanks, line 171: Should there be a suggestion for a new way to specify abi3 packages in this proposal? I understand how `noarch: python` almost works already and so can be used to get abi3 packages working with existing tools, but it is a misleading label since these packages are still architecture dependent.

Also, I had tried to test with `post-link.sh` and `pre-unlink.sh` scripts previously [here](https://github.com/conda-forge/qiskit-terra-feedstock/pull/41) but I didn't try `noarch: python` and got stuck on the run exports of Python. Maybe I just didn't know the right syntax, but it seemed like when you don't use `noarch: python` `conda-build` pins the Python version in a way that is hard to override by setting the Python version in the recipe requirements. That might be worth mentioning as another important behavior of `noarch: python` (if I was not missing something).

isuruf: > Should there be a suggestion for a new way to specify abi3 packages in this proposal? I understand how noarch: python almost works already and so can be used to get abi3 packages working with existing tools, but it is a misleading label since these packages are still architecture dependent.

This is `package_metadata_version = 1` which is for existing tool compatibility. Please see the next section for new ways.

> Also, I had tried to test with post-link.sh and pre-unlink.sh scripts previously https://github.com/conda-forge/qiskit-terra-feedstock/pull/41 but I didn't try noarch: python and got stuck on the run exports of Python. Maybe I just didn't know the right syntax, but it seemed like when you don't use noarch: python conda-build pins the Python version in a way that is hard to override by setting the Python version in the recipe requirements. That might be worth mentioning as another important behavior of noarch: python (if I was not missing something).

That should be fixable by using `ignore_run_exports`. This is only specific to conda-forge which has `run_exports` on `python` package and defaults does not AFAIK. I wanted to keep the CEP more generic.

cep-abi3.md
isuruf: Yes, we can do that.

cep-abi3.md
isuruf: `abi3` packages with `package_metadata_version = 2` should not be that different from any other package. It just has more things in `info/link.json`.

mariusvniekerk: i think something like

```
python:
   abi: abi3
```

leaves open enough space to support future expansion of other abis that may require special behavior

for example

```
python:
   abi: hpy
```

isuruf: @mariusvniekerk, both those options need the exact same thing. (B1-B4)

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+
+Conda install tools have slightly different behaviour.
+
+Conda:
+1. Actions `B1, B2, B3` are applied for packages with `A3`.
+2. Action `B4` is applied for packages with `A4`.
+
+Micromamba:
+1. Actions `B1, B2, B3` are applied for packages with both `A2, A3`.
+2. Action `B4` is applied for packages with both `A2, A4`.
```
baszalmstra, line 155: For reference, rattler follows the same behavior as micromamba.

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+
+&nbsp;&nbsp;<strong>D2</strong>:
+  Apply actions `B2, B3` if `python: py_compile` is present in `info/link.json`.
+
+&nbsp;&nbsp;<strong>D3</strong>:
+  Provide a `__supports_package_metadata_version=2` virtual package.
+
+Note that we do not require `B1` as package authors should depend on a python
+version that has a custom `site.py` that adds `lib/python/site-packages` to
+the path.
```
baszalmstra, line 229: Doesnt this imply that older python versions available on for instance conda-forge will never have the ability to use abi3 packages? If we are modifying the installer we could just as well move the Python files to the python specific site-packages directory.

isuruf: Sure, but we can also have the `site.py` in a separate conda package as well. I'm trying to reduce all these special things that a installer has to do. With the files in `lib/python/site-packages`, the only thing that a installer has to do for an abi3 package is to compile the pyc files which is an optional thing anyway.

isuruf: I'm hoping we can add support in defaults too.

baszalmstra: I would not prefer this because it means the proper functioning of the installer is dependent on a specific package. I'd rather encode the behavior in the installer itself.

isuruf: `lib/python3.10/site-packages` is a CPython specific directory and we had to use that for PyPy too because the installer didn't know about PyPy.

Can you explain why you want the installer to do special things? I would think that it's better for the installer to do less special case stuff.

baszalmstra: I agree that the installer should do less special case stuff. But I also don't think that it should depend on specific packages to function properly.

Even with this proposal the installer does already have special case behavior. However, without the dependency on a package that introduces a `site.py` file it will leave the package in a broken state. To me that adds complexity for the package maintainer.

However, one could argue that the same holds for depending on `python` itself.. 

It would be ideal if a recipe author doesn't have to care about these details.

isuruf: We could make the build tool add this package, then the installer does not have to care about it.

baszalmstra: Do we do this for other packages? Because then we tie the build-tool to specific package names which might also not be ideal.

isuruf: We take the C compiler name from `c_compiler` variable in the config. We could do the same.

isuruf: Oh I forgot about `run_exports: noarch`. We can extend that in the python recipe.

cep-abi3.md
isuruf: Here's what a conda-build PR would look like https://github.com/conda/conda-build/pull/5456

cep-abi3.md
jezdez: I'm interested what your suggestion is for this mismatch in behavior, should conda be adopted to follow mamba and rattler in this case?

cep-abi3.md
```diff
@@ -0,0 +1,269 @@
...
+```
+build:
+  python_version_independent: true
+```
+to imply a python version independent package and set `noarch: python`
+in `info/index.json`.
+
+To test that an install tool works with this scheme, you can try
+
+    conda/micromamba create -n a sympy python=3.10 -c isuruf/label/abi3 -c conda-forge
```
jezdez, line 190: Should we hard-code this into the CEP which may prove to not work anymore at some point?

cep-abi3.md
isuruf: My suggestion is to keep the existing mismatched behaviour in `package_metadata_version = 1` and make this uniform in `package_metadata_version = 2`

cep-abi3.md
isuruf: No, I'll remove it

jaimergp:
@conda/steering-council - @isuruf would like to call a vote on this soon. I'll be championing this request. In the last community call, we decided to give another two weeks for comments, and then call a vote starting December the 4th.

So, if you'd like to make any CEP-altering comments, this is your chance. Once the vote starts, we should only make editorial corrections. Thanks!

cep-abi3.md
rgommers: Note that it's fairly likely that `abi4` will become a thing in 2025 or 2026, since it seems necessary for free-threading to be able to support the Limited C API. `python-abi3` in host requirements seems to cover it though.

cep-abi3.md
```diff
@@ -0,0 +1,235 @@
...
+which would make the build tool
+
+&nbsp;&nbsp;<strong>D1</strong>:
+Set `noarch: python` in `info/index.json`.
+
+Note that `python-abi3` would set the runtime requirements.
+This is explicitly required from recipe authors so that we do not
+restrict this CEP to ABI3 packages and allow the possibility for ABI4 etc.
+
+An example `python-abi3=3.8` package would set itself in its
```
rgommers, line 200: Minor: perhaps this should use the same `{{ python_min }}` syntax as was recently introduced for noarch packages?

isuruf: Unlike python which needs multiple values for arch builds and a single value for noarch builds, we only need a single value for abi3 builds. So we can set
```yaml
python_abi3:
  - 3.8
```
in our global pinning file in conda-forge and anaconda.

rgommers: Okay, that makes sense. Some of the "automatically bump" rationale of `python_min` probably still applies, but this dependency is indeed easier to handle than `python`.

h-vetinari: > we only need a single value for abi3 builds.

That's not guaranteed AFAIU. Despite the ABI being stable, new features get added in newer python versions, and if anything depends on those features, you might need both the oldest and a newer abi3 build.

For example, cryptography [builds](https://github.com/pyca/cryptography/blob/644dcafecf47dfd598302b35dbd53c6af3189fca/.github/workflows/wheel-builder.yml#L68-L69) for both the py37 & py39 limited API. In the future, it's quite likely that other features (e.g. nogil) will similarly require a newer baseline for some features.

Of course, individual feedstocks will be able to override `python_abi3` where necessary (and we may get away to only have a single value in the global pinning), but conceptually, I don't think we can assume it'll be one value.

isuruf: Sure, but what I want to emphasize is that we don't need different requirements based on `arch` vs `noarch`.

wolfv:
I just re-read the CEP and it looks great to me. Happy that we can achieve that with the existing metadata.

cep-abi3.md
```diff
@@ -0,0 +1,235 @@
+# Support for abi3 python packages
+
+<table>
+<tr><td> Title </td><td> Support for abi3 python packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Isuru Fernando &lt;ifernando@quansight.com&gt;</td></tr>
+<tr><td> Created </td><td> July 01, 2023</td></tr>
```
jaimergp, line 7:
```suggestion
<tr><td> Created </td><td> July 01, 2024</td></tr>
<tr><td> Updated </td><td> Nov 26, 2024</td></tr>
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/86 </td></tr>
<tr><td> Implementation </td><td> https://github.com/conda/conda-build/pull/5456 </td></tr>
```

jaimergp:
The proposal looks fine! I like how easy is to follow attributes, behaviors and actions with the `A1`, `B2`, `C3` approach. 👏 

I'm echoing @wshanks thoughts here too, though: it feels weird to use a key named `noarch` to signal something that is indeed arch-specific. I understand we are leveraging the "version-independence" of `noarch: python` for this, and it's nice that we can get there with arguably little changes, but it still feels odd to keep adding unintuitive semantics to the pile. That said, I don't think it should block this PR.

Also, I'd like to make some editorial changes that don't convey meaning but I'll send those in a separate PR to your branch so this one doesn't get too noisy.

cep-abi3.md
```diff
@@ -0,0 +1,235 @@
...
+  run:
+    - cpython >=3.8
+    - python-gil
+```
+
+## Alternatives considered
+
+### Apply all actions in a `post-link.sh` script.
+
+A draft work provided at [python-feedstock](python-pr-669)
```
jaimergp, line 213:
```suggestion
A draft work provided at [python-feedstock][python-pr-669]
```

cep-abi3.md
```diff
@@ -0,0 +1,235 @@
...
+`abi3.so`. For eg:
+
+  foo.abi3.so
+
+These extensions only support the platform it was built for (for eg:
+`x86_64-linux-gnu`), but is not specified in the extension suffix.
+
+Note that the stable ABI is only specific to CPython and is not compatible with
+PyPy or other Python implementations. For a Python implementation independent
+ABI, see the [HPy project](HPy).
```
jaimergp, line 41:
```suggestion
ABI, see the [HPy project][HPy].
```

cep-abi3.md
```diff
@@ -0,0 +1,235 @@
...
+version. These extensions are identified by the extension suffix.
+For eg:
+
+  foo.cpython-310-x86_64-linux-gnu.so
+
+is an extension that support only CPython 3.10 on x86_64-linux-gnu platform.
+
+However some symbols are available in all python major.minor versions with some
+lower bound on the python version. These symbols are part of the
+[limited C API]([C_API_Stability]). It is guaranteed that the symbols Stable ABI
```
jaimergp, line 29:
```suggestion
[limited C API][C_API_Stability]. It is guaranteed that the symbols Stable ABI
```

jaimergp:
Hello @conda/steering-council. Requesting a vote for this CEP.

To vote, please use the following form to vote.

**This vote will end on 2024-12-18, End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@kkraus14 (Keith Kraus)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@beckermr (Matthew R. Becker)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@hind-m (Hind Montassif)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@trallard (Tania Allard)
- [ ]  yes
- [ ]  no
- [ ]  abstain

cep-abi3.md
```diff
@@ -0,0 +1,241 @@
...
+
+- **C1**:
+ They have `subdir: <platform>` where `<platform>` is the subdir
+ that the package was built for.
+
+- **C2**:
+ They have `noarch: python`.
+
+- **C3**:
+  `A2, A3, A4` are applied.
```
JeanChristopheMorinPerso, line 171: Minor: A2 is already covered by C3 if I'm not mistaken. Do I get this right?

jaimergp:

* Total possible voters: 17. 
* Valid votes: 11 / 17 = 64.7%. 

Yes votes (11 / 100.00%):

1. @xhochy
1. @chenghlee
1. @ocefpaf
1. @jezdez
1. @wolfv
1. @jaimergp
1. @baszalmstra
1. @kkraus14
1. @beckermr
1. @Hind-M
1. @trallard

No votes (0 / 0.00%).

Abstain votes (0 / 0.00%).

Not voted (6):

1. @CJ-Wright
1. @jakirkham
1. @mariusvniekerk
1. @marcelotrevisani
1. @msarahan
1. @mbargull

Invalid votes (0).

11/17 votes = 64.7% meets the minimum quorum. 
11/11 YES votes meets the approval threshold (60%).

PR #90, opened 2024-09-13
CEP for specifying the location of site-packages in repodata

jjhelmus:
### Description

This CEP proposed to introduce an optional `python_site_packages_path` field in repodata that `python` packages can use to explicitly specify the location of the site-packages directory. This allows for `noarch: python` packages to be linked into the correct location when this path does not match the on used by standard builds of CPython. This includes alternative implementation of the Python programming langauge (PyPy, GraalPy) and free-threading builds of CPython.

The motivation and request for this CEP came from conda/conda#14053, which discusses `noarch: python` support in conda in environments with the experimental free-threading configuration of CPython 3.13.

### DIFF
--- a/cep-999.md
+++ b/cep-999.md
@@ -1,0 +1,162 @@
+<table>
+<tr><td> Title </td><td> Optional python site-packages path in repodata </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Jonathan Helmus &lt;jjhelmus@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Sep 13, 2024</td></tr>
+<tr><td> Updated </td><td> Sep 13, 2024</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+To avoid confusion this document will use a stylized `python` to indicate a conda package with that name.
+Other uses of “Python” will be followed by a qualifier, like "programming language" or "interpreter".
+Specific implementations of the Python programming language will be referred to by name, such as CPython or PyPy.
+
+## Abstract
+
+We propose adding a new optional field in repodata.json that `python` packages can use to specify the destination path when installing `noarch: python` packages.
+
+## Background
+
+In the conda ecosystem the package named `python` provides, either directly or via a dependency, an interpreter for the Python programming language.
+This interpreter is often CPython but can also be alternative implementations such as PyPy or GraalPy.
+
+Tools like conda, mamba and pixie support installing `noarch: python` packages into conda environments.
+Whereas the content of other conda packages are linked into the target environment based upon their path within the package, files within the "site-packages" directory of `noarch: python` packages are linked into the site-packages directory of the target environment.
+
+In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is "Lib/site-packages" on Windows and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
+These paths are the defaults for CPython but are not the correct paths for alternative implementation of the Python programming language or for some configurations of CPython.
+For example PyPy uses "lib/pypyX.Y/site-packages" on POSIX systems.
+
+In these cases kludgy workarounds, [often a symlink](https://github.com/conda-forge/pypy3.6-feedstock/blob/ac4201e80432469971626d277718edf80b22d6ef/recipe/build.sh#L102), are used in the `python` package to support the installation of `noarch: python` packages.
+In the next section a new field will be proposed that allows `python` package to explicitly declare the location of the site-packages directory to avoid the need for these work-arounds.
+
+## Specification
+
+An optional field `python_site_packages_path` may be included in a conda package’s info/index.json file as a string and in the corresponding repodata.json entry for the package.
+When defined, this field specifies the path of the Python interpreter’s site-packages directory relative to the root of the environment.
+If a `python` package that includes this field is installed into an environment, all `noarch: python` packages that are installed into the environment will have the files in their "site-packages" directory linked into the path specified by this field.
+
+A value of "null" can be used to indicate that this field is not specified.
+It is acceptable to not include the field in repodata.json in this case.
+
+If this field is missing or specified as "null" a default site-packages path will be used.
+This path is "Lib/site-packages" if the environment targets a Windows platform and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
+This matches the current behavior of conda, mamba and pixie.
+
+This field should only be included in the metadata for `python` packages but tools must not fail if packages with other names include this field, rather the entry should be ignored.
+
+To avoid unnecessarily increasing the size of repodata.json, it is highly recommended that this field only be included in `python` packages where it is required, that is where the site-packages directory is not the default value.
+Packages with other names or `python` packages that use the default site-packages path should not include this field.
+
+This path must not point to a location outside of the root of the environment (e.g. it cannot navigate up a directory or be an absolute path).
+If a package specifies a path which violates this requirement the package must not be installed and an appropriate error should be shown to the user.
+
+A possible way to check the validity of the `python_site_packages_path` field is with this function:
+
+``` Python
+def is_valid(target_prefix: str, python_site_packages_path: str) -> bool:
+    target_prefix = os.path.realpath(target_prefix)
+    full_path = os.path.realpath(os.path.join(target_prefix, python_site_packages_path))
+    test_prefix = os.path.commonpath((target_prefix, full_path))
+    return test_prefix == target_prefix
+```
+
+When a package which includes this optional field is indexed the `python_site_packages_path` field will be included in the repodata entry for the package.
+For example the repodata entry for a python-3.13.0rc1 package using the free-threading build configuration might look as follows:
+
+```
+    "python-3.13.0rc1-haa6bb3f_0_cpython_cp313t.tar.bz2": {
+      "build": "haa6bb3f_0_cpython_cp313t",
+      "build_number": 0,
+      "depends": [
+        "bzip2 >=1.0.8,<2.0a0",
+        "expat >=2.6.2,<3.0a0",
+        "libffi >=3.4,<4.0a0",
+        "ncurses >=6.4,<7.0a0",
+        "openssl >=3.0.14,<4.0a0",
+        "readline >=8.1.2,<9.0a0",
+        "sqlite >=3.45.3,<4.0a0",
+        "tk >=8.6.14,<8.7.0a0",
+        "tzdata",
+        "xz >=5.4.6,<6.0a0",
+        "zlib >=1.2.13,<1.3.0a0"
+      ],
+      "license": "PSF-2.0",
+      "license_family": "PSF",
+      "timestamp": 1722610680768,
+      "track_features": "free-threading",
+      "md5": "c09289eb86239e1221533457d861f1a3",
+      "name": "python",
+      "size": 16332720,
+      "subdir": osx-arm64,
+      "version": "3.13.0rc1",
+      "sha256": "fa0ae22c13450fe6c30c754ee5efbd7fe7e7533b878d7be96e74de56211d19df",
+      "python_site_packages_path": "lib/python3.13t/site-packages"
+    },
+```
+
+This field will also be included in repodata derivatives, like current_repodata.json or sharded repodata when appropriate.
+
+Because this field is present in repodata.json it can be hot-fixed to correct mistakes or omissions.
+This allows existing `python` packages to retroactively specify the locations of their site-packages directory.
+
+## Motivation
+
+This proposal was motivated by the free-threading configuration of CPython 3.13 which uses a different site-packages location on POSIX systems.
+Specifically, the free-threading build uses "lib/python3.13t/site-packages".
+
+A symlink between these paths or logic in [sitecustomize.py](https://docs.python.org/3/library/site.html#module-sitecustomize) allows packages installed into the incorrect site-packages directory to operate but this is less than ideal.
+
+For additional details and discussion see [conda issue 14053](https://github.com/conda/conda/issues/14053).
+
+Although this is motivated by the free-threading configuration of CPython, the same underlying issue occurs in many non-CPython implementations.
+This includes both PyPy and GraalPy which use different paths for site-packages than CPython.
+The conda packages in the conda-forge channel for both of these use symlinks to work around tools which link files into the incorrect site-packages directory.
+
+## Security
+
+Because this proposal can change the location where files are installed it is worth considering what, if any, security implementation this change entails.
+
+A primary security concern is that by allowing the `python` package to specify where files will be installed, a malicious package could write or overwrite files with malicious content.
+
+If files can only be installed into the environment where the package is installed, the damage from an attack is limited since a malicious package can already install files anywhere in the environment.
+
+For example a malicious `python` package could specify an invalid site-packages path, in which case `noarch: python` packages installed in the environment would not work.
+This is inconvenient but not a security concern and should be addressed by removing or disabling the package in question.
+
+The malicious package could also specify a site-packages path that would cause files from `noarch: python` packages to populate or replace key files in the environment.
+But this type of attack is already possible, in fact the `python` package itself could include these files.
+
+Of more concern is the possibility of a package writing files outside of the environment.
+This could replace critical system files with malicious content.
+Because of this risk, paths which are outside of the environment are invalid as `python_site_packages_path` entries.
+Tools must check that the field is valid and error out when it is invalid.
+
+## Backwards Compatibility
+
+This change maintains backwards compatibility.
+All existing `python` packages do not include this field.
+The behavior when this field is not specified matches what is currently done by conda, mamba and pixi.
+The only behavior change occurs when this field is included.
+
+Because existing releases of conda, mamba and pixie do not support this field, `python` packages should continue to provide work around to allow files to be installed into an incorrect site-packages directory until such time as this proposal has been implemented and available for a sufficient amount of time.
+
+## Other sections
+
+A number of alternatives were discussed and considered to address this problem. These include:
+
+* Using symlinks or a `sitecustomize.py` file to allow linking into the incorrect site-packages directory.
+  This is a work-around for the problem, not a fix.
+* Querying the interpreter to determine the site-packages directory.
+  This was rejected as it requires running the interpreter as part of the install step and requires that the `python` package be installed and operational before `noarch: python` packages can be linked.
+* Using the build string of the `python` or `python_abi` package to determine the site-packages location.
+  This is an indirect method of obtaining information that can be more effectively specified explicitly.
+* Storing this information is another metadata file, such as "about.json" or "paths.json".
+  The downside of this approach is that the data cannot be hotfixed and the path for the site-packages directory would not be known until the `python` package is downloaded and unpacked, potentially limiting the order in which packages can be linked.
+
+For more alternatives and discussion see [conda issue 14053](https://github.com/conda/conda/issues/14053).
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).


isuruf (approved):

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+## Abstract
+
+We propose adding a new optional field in repodata.json that `python` packages can use to specify the destination path when installing `noarch: python` packages.
+
+## Background
+
+In the conda ecosystem the package named `python` provides, either directly or via a dependency, an interpreter for the Python programming language.
+This interpreter is often CPython but can also be alternative implementations such as PyPy or GraalPy.
+
+Tools like conda, mamba and pixie support installing `noarch: python` packages into conda environments.
```
baszalmstra, line 24:
```suggestion
Tools like conda, mamba and pixi support installing `noarch: python` packages into conda environments.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+An optional field `python_site_packages_path` may be included in a conda package’s info/index.json file as a string and in the corresponding repodata.json entry for the package.
+When defined, this field specifies the path of the Python interpreter’s site-packages directory relative to the root of the environment.
+If a `python` package that includes this field is installed into an environment, all `noarch: python` packages that are installed into the environment will have the files in their "site-packages" directory linked into the path specified by this field.
+
+A value of "null" can be used to indicate that this field is not specified.
+It is acceptable to not include the field in repodata.json in this case.
+
+If this field is missing or specified as "null" a default site-packages path will be used.
+This path is "Lib/site-packages" if the environment targets a Windows platform and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
+This matches the current behavior of conda, mamba and pixie.
```
baszalmstra, line 45:
```suggestion
This matches the current behavior of conda, mamba and pixi.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+Tools must check that the field is valid and error out when it is invalid.
+
+## Backwards Compatibility
+
+This change maintains backwards compatibility.
+All existing `python` packages do not include this field.
+The behavior when this field is not specified matches what is currently done by conda, mamba and pixi.
+The only behavior change occurs when this field is included.
+
+Because existing releases of conda, mamba and pixie do not support this field, `python` packages should continue to provide work around to allow files to be installed into an incorrect site-packages directory until such time as this proposal has been implemented and available for a sufficient amount of time.
```
baszalmstra, line 143:
```suggestion
Because existing releases of conda, mamba and pixi do not support this field, `python` packages should continue to provide work around to allow files to be installed into an incorrect site-packages directory until such time as this proposal has been implemented and available for a sufficient amount of time.
```

baszalmstra (approved):

cep-999.md
jjhelmus: Sorry about that. Fixed.

dholth:
https://github.com/conda/ceps/pull/86/files is working in the same space, to allow ABI3 packages

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+      "license_family": "PSF",
+      "timestamp": 1722610680768,
+      "track_features": "free-threading",
+      "md5": "c09289eb86239e1221533457d861f1a3",
+      "name": "python",
+      "size": 16332720,
+      "subdir": osx-arm64,
+      "version": "3.13.0rc1",
+      "sha256": "fa0ae22c13450fe6c30c754ee5efbd7fe7e7533b878d7be96e74de56211d19df",
+      "python_site_packages_path": "lib/python3.13t/site-packages"
```
dholth, line 95:
```suggestion
      "python_site_packages": "lib/python3.13t/site-packages"
```

dholth (commented):
I tend to be in favor of putting the site-packages path in a separate file in python's info/, for example in link.json or paths.json where it would be possible to increment the version, and make it non-installable by incompatible condas.

cep-999.md
jjhelmus: The field specified in the CEP is `python_site_packages_path` with `_path`

dholth: Yes, suggestion is to rename without _path

jjhelmus: I'm open to either name. Does anyone have a strong opinion one way or another?

jjhelmus:
I have put together a draft PR for conda-build that implements support for specifying this field in the build section of a recipe, see conda/conda-build#5502.

With this I created sample packages that specify this metadata field for testing with downstream tools. They are available on anaconda.org in the `jjhelmus/label/sp_path` channel. 

There are three noarch packages with the following paths specified
* python 3.99.99 : `lib/python3.99t/site-packages` - A "free-threading" like path
* python 3.99.90 : `sample_path/from_the_file/index_json` - Something that stands out
* python 3.99.50 : `../../../../../invalid_path` : Should be rejected by tooling as being invalid

jjhelmus:
conda/conda#14256 contains an rough draft of the change needed for conda to make use of this field.

isuruf:
Should we fix the name for python in the shebang of entry-points in this CEP or a separate one? Currently there's a `python3.13t->python3.13` symlink

jezdez:
> Should we fix the name for python in the shebang of entry-points in this CEP or a separate one? Currently there's a `python3.13t->python3.13` symlink

I'd suggest keeping it here TBH, since a whole CEP for a simple change like this may be overkill.

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Background
+
+In the conda ecosystem the package named `python` provides, either directly or via a dependency, an interpreter for the Python programming language.
+This interpreter is often CPython but can also be alternative implementations such as PyPy or GraalPy.
+
+Tools like conda, mamba and pixi support installing `noarch: python` packages into conda environments.
+Whereas the content of other conda packages are linked into the target environment based upon their path within the package, files within the "site-packages" directory of `noarch: python` packages are linked into the site-packages directory of the target environment.
+
+In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is "Lib/site-packages" on Windows and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```
jezdez, line 27:
```suggestion
In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is `Lib/site-packages" on Windows and `lib/pythonX.Y/site-packages` on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+To avoid confusion this document will use a stylized `python` to indicate a conda package with that name.
+Other uses of “Python” will be followed by a qualifier, like "programming language" or "interpreter".
+Specific implementations of the Python programming language will be referred to by name, such as CPython or PyPy.
+
+## Abstract
+
+We propose adding a new optional field in repodata.json that `python` packages can use to specify the destination path when installing `noarch: python` packages.
```
jezdez, line 17:
```suggestion
We propose adding a new optional field in `repodata.json` that `python` packages can use to specify the destination path when installing `noarch: python` packages.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+In the conda ecosystem the package named `python` provides, either directly or via a dependency, an interpreter for the Python programming language.
+This interpreter is often CPython but can also be alternative implementations such as PyPy or GraalPy.
+
+Tools like conda, mamba and pixi support installing `noarch: python` packages into conda environments.
+Whereas the content of other conda packages are linked into the target environment based upon their path within the package, files within the "site-packages" directory of `noarch: python` packages are linked into the site-packages directory of the target environment.
+
+In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is "Lib/site-packages" on Windows and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
+These paths are the defaults for CPython but are not the correct paths for alternative implementation of the Python programming language or for some configurations of CPython.
+For example PyPy uses "lib/pypyX.Y/site-packages" on POSIX systems.
```
jezdez, line 29:
```suggestion
For example PyPy uses `lib/pypyX.Y/site-packages` on POSIX systems.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+In these cases kludgy workarounds, [often a symlink](https://github.com/conda-forge/pypy3.6-feedstock/blob/ac4201e80432469971626d277718edf80b22d6ef/recipe/build.sh#L102), are used in the `python` package to support the installation of `noarch: python` packages.
+In the next section a new field will be proposed that allows `python` package to explicitly declare the location of the site-packages directory to avoid the need for these work-arounds.
+
+## Specification
+
+An optional field `python_site_packages_path` may be included in a conda package’s info/index.json file as a string and in the corresponding repodata.json entry for the package.
```
jezdez, line 36:
```suggestion
An optional field `python_site_packages_path` may be included in a conda package’s `info/index.json` file as a string and in the corresponding `repodata.json` entry for the package.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+When defined, this field specifies the path of the Python interpreter’s site-packages directory relative to the root of the environment.
+If a `python` package that includes this field is installed into an environment, all `noarch: python` packages that are installed into the environment will have the files in their "site-packages" directory linked into the path specified by this field.
+
+A value of "null" can be used to indicate that this field is not specified.
+It is acceptable to not include the field in repodata.json in this case.
+
+If this field is missing or specified as "null" a default site-packages path will be used.
```
jezdez, line 43:
```suggestion
If this field is missing or specified as `null` a default site-packages path will be used.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+This path is "Lib/site-packages" if the environment targets a Windows platform and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```
jezdez, line 44:
```suggestion
This path is `Lib/site-packages` if the environment targets a Windows platform and `lib/pythonX.Y/site-packages` on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+``` Python
+def is_valid(target_prefix: str, python_site_packages_path: str) -> bool:
+    target_prefix = os.path.realpath(target_prefix)
+    full_path = os.path.realpath(os.path.join(target_prefix, python_site_packages_path))
+    test_prefix = os.path.commonpath((target_prefix, full_path))
+    return test_prefix == target_prefix
+```
+
+When a package which includes this optional field is indexed the `python_site_packages_path` field will be included in the repodata entry for the package.
+For example the repodata entry for a python-3.13.0rc1 package using the free-threading build configuration might look as follows:
```
jezdez, line 66:
```suggestion
For example the repodata entry for a `python-3.13.0rc1` package using the free-threading build configuration might look as follows:
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+      "name": "python",
+      "size": 16332720,
+      "subdir": osx-arm64,
+      "version": "3.13.0rc1",
+      "sha256": "fa0ae22c13450fe6c30c754ee5efbd7fe7e7533b878d7be96e74de56211d19df",
+      "python_site_packages_path": "lib/python3.13t/site-packages"
+    },
+```
+
+This field will also be included in repodata derivatives, like current_repodata.json or sharded repodata when appropriate.
```
jezdez, line 99:
```suggestion
This field will also be included in repodata derivatives, like `current_repodata.json` or sharded repodata when appropriate.
```

jezdez, line 99: `current_repodata.json` is less important ever since the libmamba solver has been rolled out, do you think it would be a dealbreaker if we'd not support this in that file or am I jumping the gun on this one?

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+Because this field is present in repodata.json it can be hot-fixed to correct mistakes or omissions.
```
jezdez, line 101:
```suggestion
Because this field is present in `repodata.json` it can be hot-fixed to correct mistakes or omissions.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+This allows existing `python` packages to retroactively specify the locations of their site-packages directory.
+
+## Motivation
+
+This proposal was motivated by the free-threading configuration of CPython 3.13 which uses a different site-packages location on POSIX systems.
+Specifically, the free-threading build uses "lib/python3.13t/site-packages".
```
jezdez, line 107:
```suggestion
Specifically, the free-threading build uses `lib/python3.13t/site-packages`.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+A number of alternatives were discussed and considered to address this problem. These include:
+
+* Using symlinks or a `sitecustomize.py` file to allow linking into the incorrect site-packages directory.
+  This is a work-around for the problem, not a fix.
+* Querying the interpreter to determine the site-packages directory.
+  This was rejected as it requires running the interpreter as part of the install step and requires that the `python` package be installed and operational before `noarch: python` packages can be linked.
+* Using the build string of the `python` or `python_abi` package to determine the site-packages location.
+  This is an indirect method of obtaining information that can be more effectively specified explicitly.
+* Storing this information in another metadata file, such as "about.json" or "paths.json".
```
jezdez, line 155:
```suggestion
* Storing this information in another metadata file, such as `about.json` or `paths.json`.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+In the next section a new field will be proposed that allows `python` package to explicitly declare the location of the site-packages directory to avoid the need for these work-arounds.
+
+## Specification
+
+An optional field `python_site_packages_path` may be included in a conda package’s info/index.json file as a string and in the corresponding repodata.json entry for the package.
+When defined, this field specifies the path of the Python interpreter’s site-packages directory relative to the root of the environment.
+If a `python` package that includes this field is installed into an environment, all `noarch: python` packages that are installed into the environment will have the files in their "site-packages" directory linked into the path specified by this field.
+
+A value of "null" can be used to indicate that this field is not specified.
+It is acceptable to not include the field in repodata.json in this case.
```
jezdez, line 41:
```suggestion
It is acceptable to not include the field in `repodata.json` in this case.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+If this field is missing or specified as "null" a default site-packages path will be used.
+This path is "Lib/site-packages" if the environment targets a Windows platform and "lib/pythonX.Y/site-packages" on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
+This matches the current behavior of conda, mamba and pixi.
+
+This field should only be included in the metadata for `python` packages but tools must not fail if packages with other names include this field, rather the entry should be ignored.
+
+To avoid unnecessarily increasing the size of repodata.json, it is highly recommended that this field only be included in `python` packages where it is required, that is where the site-packages directory is not the default value.
```
jezdez, line 49:
```suggestion
To avoid unnecessarily increasing the size of `repodata.json`, it is highly recommended that this field only be included in `python` packages where it is required, that is where the site-packages directory is not the default value.
```

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+If a package specifies a path which violates this requirement the package must not be installed and an appropriate error should be shown to the user.
+
+A possible way to check the validity of the `python_site_packages_path` field is with this function:
+
+``` Python
+def is_valid(target_prefix: str, python_site_packages_path: str) -> bool:
+    target_prefix = os.path.realpath(target_prefix)
+    full_path = os.path.realpath(os.path.join(target_prefix, python_site_packages_path))
+    test_prefix = os.path.commonpath((target_prefix, full_path))
+    return test_prefix == target_prefix
```
jezdez, line 62:
```suggestion
from pathlib import Path

def is_valid(target_prefix: str, python_site_packages_path: str) -> bool:
    target_prefix = Path(target_prefix).resolve()
    full_path = (target_prefix / python_site_packages_path).resolve()
    return target_prefix in full_path.parents
```

jezdez (commented):
Left a few comments, mostly just tweaks for readability

isuruf:
A bit of context about the python executable name.

On Unix, CPython (upstream and conda) has python3.13 for freethreading and default builds. For freethreading python3.13 is a symlink to python 3.13t. On Unix, PyPy does the same

On Windows, CPython upstream has python.exe for default build and python3.13t.exe for freethreading builds as there's only one installer. On conda-forge, python3.13.exe and python.exe is a copy of python3.13t.exe for freethreading builds.
On PyPy, they have python3.10.exe copy for python.exe

For noarch: packages, the entry-points right now uses python3.13 for unix and python.exe for windows. Do we want to keep using these or have a way for specifying the executable name? Note that it's straightforward to change the name in Unix as it's about changing the shebang, but it's hard on windows due to the need to create an executable.

I'm okay with keeping the current status quo and requiring `python<major>.<minor>` on Unix and `python.exe` on Windows for all python flavours.

jjhelmus:
At least on macOS, Python virtual environment created using `venv` from the standard library provide executables named `python`, `python3` and `python3.x` with the second two being symlinks to the first. A `t` executable is not provided for freethreading environments which is surprising to me.  

By default the path of the executable used to run pip is used when creating console entry-points in these environments. The `pip`, `pip3` and `pip3.x` command all use the path to `python` for the executable so this path will be used unless pip is started using `python3{.X} -m pip`.

Based on this I would be in favor of conda using `python` for the shebang in entry-points.

jjhelmus:
> I tend to be in favor of putting the site-packages path in a separate file in python's info/, for example in link.json or paths.json where it would be possible to increment the version, and make it non-installable by incompatible condas.

The benefit of including the site-packages path in repodata is the ability to correct it with hotfixing after the package is built. This can be used to add this field to existing packages, like the PyPy and GraalPy packages in conda-forge, or correct mistakes in future packages. If functionality is not needed then the field could be moved to link.json or paths.json. 

Another benefit of including this is repodata is discoverability. Repodata can be examined without downloading and extracting the package itself.

Incrementing the version would prevent the installation with incompatible version of condas but the user experience is not great. Specifically the environment would solve and the user would not be aware of a problem until the packages are downloaded, extracted and staged for linking.

cep-999.md
jjhelmus: It is not a dealbreaker to not include the field in `current_repodata.json`.  The field should be included in a `python` shard of repodata if/when that becomes a reality. I don't see including this as a big lift in either case as the field is tied to the package entry from which I expect both to be derived.

dholth: Individual entries in current_repodata and sharded repodata will be the same as the ones in monolithic repodata.

mariusvniekerk (commented):
So what should the behavior be if more than one package in installed with this field set?

I assume failure of some kind?

baszalmstra:
Another advantage of storing this information in the repodata is that that information is always readily available during installation, even for already installed records (even if the cache is cleared).

baszalmstra:
> So what should the behavior be if more than one package in installed with this field set?

Good question! I assumed it would be ignored except for the package that provides python. In rattler we currently assume that is the package called “python”. I wonder how conda does this.

jjhelmus:
> So what should the behavior be if more than one package in installed with this field set?
> 
> I assume failure of some kind?

This field is only allowed on the conda package named `python`, if packages with other names include this field it is ignored as mentioned in the CEP. There can only be one `python` package installed in an environment so there will never be a case where there is more that one package installed with this field. 

Another way of thinking about this change is that instead of using the package with name `python` to determine the major and minor version that are used in the site-packages path, the `python` package can also use this field to declare the path.  The sample [conda PR](https://github.com/conda/conda/pull/14256) shows how this could be implemented.

jezdez:
In coordination with @jjhelmus, I post this request for a vote:

@conda/steering-council

To vote, please use the following form to vote. If you have questions concerning the proposal, you may also leave a comment or code review.

**This vote will end in one week on 2024-10-15, End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@kkraus14 (Keith Kraus)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@beckermr (Matthew R. Becker)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@hind-m (Hind Montassif)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@trallard (Tania Allard)
- [ ]  yes
- [ ]  no
- [ ]  abstain

jjhelmus:
> When conda doesn't understand this field, does that Python package work?

Yes, the current GraalPy, PyPy and free-threading CPython `python` packages in conda-forge include work arounds like symlinks or `sitecustomize.py` files that work conda linking `noarch: python` packages into the incorrect path. The CEP mentions these work around the recommends that they be maintained for some time:

> Because existing releases of conda, mamba and pixi do not support this field, `python` packages should continue to provide work around to allow files to be installed into an incorrect site-packages directory until such time as this proposal has been implemented and available for a sufficient amount of time.

cep-999.md
jjhelmus: This style of check would require resolving `target_prefix` as well as it needs to take into account symbolic links and various re-directs on Windows file systems. 
The function as written in the CEP closely matches the filter used by by tarfile module in the CPython standard library for validating paths, https://github.com/python/cpython/blob/v3.13.0/Lib/tarfile.py#L816-L818

cep-999.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Background
+
+In the conda ecosystem the package named `python` provides, either directly or via a dependency, an interpreter for the Python programming language.
+This interpreter is often CPython but can also be alternative implementations such as PyPy or GraalPy.
+
+Tools like conda, mamba and pixi support installing `noarch: python` packages into conda environments.
+Whereas the content of other conda packages are linked into the target environment based upon their path within the package, files within the "site-packages" directory of `noarch: python` packages are linked into the site-packages directory of the target environment.
+
+In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is `Lib/site-packages" on Windows and `lib/pythonX.Y/site-packages` on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```
jaimergp, line 27:
```suggestion
In conda 24.7.1 and many other versions the location of the site-packages directory within the environment is `Lib/site-packages` on Windows and `lib/pythonX.Y/site-packages` on other platforms where X and Y and the major and minor versions of the `python` package installed in the environment.
```

wolfv:
I just had one shower-thought: what if we would store a map for this, and could that help with `noarch: ruby` / `noarch: php` etc. packages?

It's been a long standing issue that we cannot support other languages with noarch packages as easily as Python for which each package manager needs to add some specific code. 

If we had a `noarch_map: { ... }` field of some sort that would map directories, such as:

```json
noarch_map: {
   "site-packages": "lib/python3.10/site-packages"
}
```

And for Ruby it could read something like:

```json
noarch_map: {
   "gems": "lib/ruby5.3/gemfiles/gems"
}
```

This would still leave open the question on pre-compiling `py` / `rb` / ... files but it would solve one of the noarch-ruby problems and would not introduce more python special handling. 

The `noarch_map` could be consulted for any `noarch` package (generic or not) that contains a folder of the sort.

I think this might be worth it to revisit if we can afford it.

I was trying to find some old references / issues, could only find one so far.

- https://github.com/conda/conda/issues/702#issuecomment-53069244

isuruf:
I don't think we should introduce any other noarch types. `noarch: generic` is enough for most use-cases and we should just use a version independent path for other languages. For python, it's historic debt.

mcg1969:
But is that possible?

Certainly, one interpretation of noarch-Python is that it's historic debt. But eliminating it would require patching the default path search behavior of python; and there is version-dependent byte compilation as well.

In contrast, for R we are fortunate that the search path is platform and version-independent, but the byte compilation issue remains, which means that currently we have to rebuild many noarch packages with each major R release.

I think it would be good to  enumerate the additional languages being considered and the potential obstacles to the use of noarch-generic. If there are none, great, but at least for me I'm speaking from ignorance of these alternatives.

wolfv:
@isuruf we wouldn't have to introduce any new noarch types. Just move the files in the right places and the noarch-map would take care of it :)
I am just advocating to produce less technical debt for python specifics.

isuruf:
> Certainly, one interpretation of noarch-Python is that it's historic debt. But eliminating it would require patching the default path search behavior of python; and there is version-dependent byte compilation as well.

Right, it would require a `.pth` file to change the search behaviour. `noarch: python` should be used only for byte compilation, the moving of files shouldn't be necessary IMO, but unrelated to this CEP.

@wolfv, yes, but it's adding another layer of unnecessary behaviour when all we need to do is to make sure the packages end up in a version-independent path in the first place. You are producing more general technical debt in order to produce less technical debt for python specifics.

baszalmstra:
@jezdez What is the status of this CEP? Did it pass or are there still discussions to be had?

jezdez:
> @jezdez What is the status of this CEP? Did it pass or are there still discussions to be had?

I'll update the PR in the coming week, currently out of office.

jjhelmus:
> If we had a `noarch_map: { ... }` field of some sort that would map directories, such as:

Why make this a nested field when any individual package should only ever define one entry. A `python` package can declare the location of the site-packages directory, it would not declare a R or Ruby location.  In addition ["Flat is better than nested"](https://peps.python.org/pep-0020/).

wolfv:
One could imagine multiple entries, such as:

```
noarch_map: {
  "r_bin": "lib/r/5.0/bin",
  "r_lib": "lib/r/lib",
}
```

and noarch packages that would have `r_bin` and `r_lib` at the root would get mapped.

But yeah, idk, maybe not useful enough :)

wolfv:
Awesome, that's great! :)

jezdez:

Total voters: 17 (valid: 13 = 76.47%)

Yes votes (13 / 100.00%):
- @xhochy (Uwe Korn)
- @chenghlee (Cheng H. Lee)
- @ocefpaf (Filipe Fernandes)
- @msarahan (Michael Sarahan)
- @jezdez (Jannis Leidel)
- @wolfv (Wolf Vollprecht)
- @baszalmstra (Bas Zalmstra)
- @beckermr (Matthew R. Becker)
- @hind-m (Hind Montassif)
- @trallard (Tania Allard)

No votes (0 / 0.00%)):

Abstain votes (0 / 0.00%):

Not voted (4):
- @mbargull (Marcel Bargull)
- @jakirkham (John Kirkham)
- @kkraus14 (Keith Kraus)

Invalid votes (0):

Thus we reached quorum and enough YES votes to mark this as accepted. 🎉

baszalmstra:
This has been implemented in [rattler](https://github.com/conda/rattler/pull/909) (and thus in the next pixi release) and [rattler-build](https://github.com/prefix-dev/rattler-build/pull/1161) !

PR #91, opened 2024-09-20
chore: add first version of channel registry cep

wolfv:


cep-channel-registry.md
```diff
@@ -0,0 +1,104 @@
...
+
+The registry will be implemented as a JSON file in a public GitHub repository. It will include the following information for each channel:
+
+- List of mirrors
+- Cryptographic trust root for The Update Framework (TUF)
+- License information
+- Description
+- Channel dependencies, that link a given channel to other channels
+- Homepage and GitHub URL, Logo, etc.
+- Wether the channel is commercial, and where the Terms of Service are located
```
jezdez, line 49: Could you add an example for that as well?

cep-channel-registry.md
```diff
@@ -0,0 +1,104 @@
...
+        "https://mirrors.tuna.tsinghua.edu.cn/anaconda/cloud/conda-forge"
+      ],
+      "trust_root": "sha256:a1b2c3d4e5f6...",
+      "license": "BSD-3-Clause",
+      "description": "Community-led collection of recipes, build infrastructure and distributions for the conda package manager.",
+      "dependencies": [],
+      "homepage": "https://conda-forge.org",
+      "github_url": "https://github.com/conda-forge",
+      "logo_url": "https://conda-forge.org/assets/img/logo.png",
+      "commercial": false,
```
jezdez, line 70: "commercial" might not be enough to elaborate the complexity of terms of services, I'd suggest to instead add a `legal` field, that would be an optional field to enable to describe what it is. Also, a `terms_url` to point to the specific terms that are currently valid?

cep-channel-registry.md
dholth, line 49:
```suggestion
- Whether the channel is commercial, and where the Terms of Service are located
```

cep-channel-registry.md
SylvainCorlay: I like the idea of `legal` being a dictionary with some information, although there should probably still be a boolean that indicates whether running `conda install` could make you owe money to the people operating that channel.

PR #99, opened 2024-11-19
Add CEP for frozen environments

jaimergp:


kenodegard:
Something like this plugin (which uses a `$CONDA_PREFIX/.protected` file): https://github.com/conda-incubator/conda-protect

baszalmstra:
I wonder if what you had in mind for pixi itself. I assume it should add the frozen file to the environments it creates to prevent other tools to modify it. But should it also respect environments that have this file and not modify them?

With regard to the base environment, when will the frozen file be added? Will a user, after adding it, always have to pass an override flag?

jaimergp:
> Something like this plugin (which uses a $CONDA_PREFIX/.protected file): [conda-incubator/conda-protect](https://github.com/conda-incubator/conda-protect)

Oh yes, I need to add this to the CEP references. It's essentially what conda-protect does, but with some inspiration for EXTERNALLY-MANAGED. I envision that conda-protect could be updated to honor this `conda-meta/frozen` file.

> I assume it should add the frozen file to the environments it creates to prevent other tools to modify it. 

Correct.

> But should it also respect environments that have this file and not modify them?

Yes. Modifications should only be allowed with some manual override. We are trying to prevent users from shooting themselves in the foot, mostly by accident. But AFAIK, pixi does not really modify foreign environments, does it?

kenodegard:
> > I assume it should add the frozen file to the environments it creates to prevent other tools to modify it.
>
> Correct.

Hm should we add a `tool` field to the JSON spec to indicating which tool froze the env? A `timestamp` could also be useful?

jaimergp:
> Hm should we add a tool field to the JSON spec to indicating which tool froze the env? A timestamp could also be useful?

I was wondering about it but the equivalent PEP was kept deliberately simple. If the tool that froze the environment was to "sign it", then it can do so in the text message. The timestamp can be obtained from the modification time on disk too. I have nothing against adding the keys, but I want to make sure we are not falling into premature optimization with extra keys.

jaimergp:
Any other comments @conda/steering-council? I'd like to start a vote in 14 days.

cep-00??.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Specification
+
+- The marker file path MUST be `$CONDA_PREFIX/conda-meta/frozen`. This is case-sensitive.
+- The marker file can be empty. It can optionally contain a JSON document with the schema described with a single key `message`. The value MUST be a non-empty string.
+- Tools MUST respect the presence of `frozen` in the environment and error out with:
+    - A message chosen by the tool, if `frozen` is empty.
+    - The message included in the `message` key, if `frozen` is not empty.
+    - A help message explaining how to override the check, if available.
+- Tools SHOULD offer a way to override the presence of `frozen`, but never by default. For example, `--override-frozen-env`.
```
beckermr, line 37: You should include MUST for the "never by default" clause to reduce ambiguity.

jaimergp:
```suggestion
- Tools SHOULD offer a way to override the presence of `frozen` (e.g. `--override-frozen-env`). However, tools MUST NOT enable the overrides by default.
```

Something like this?

jaimergp:
Ok, I'm back to this CEP. @conda/steering-council, planning to submit a vote in 7 days. Let me know if there are any further comments, thank you!

marcelotrevisani (approved):

msarahan (approved):

jjhelmus (approved):

jaimergp:
I don't think we need to exclude it. Each conda tool can include the text they need in the `frozen` file. 

For example, `conda` could mention "If you are only updating openssl, then it is ok to run --ignore-frozen-env" for its `base` lock, but other clients like `micromamba` that are statically linked shouldn't care too much about it.

There are other UX concerns for `conda` and the plugin system, but IMO these should be covered by extra porcelain UI like `conda self-update` and `conda plugins` subcommands. Not the scope of this CEP, anyway.

xhochy:
Voted. Seems like my initial vote wasn't persisted by GitHub :(

kkraus14 (approved):

jaimergp:

* Total possible voters: 16. 
* Valid votes: 13 / 16 = 81.25%. 

Yes votes (13 / 100.00%):

1. @xhochy
1. @chenghlee
1. @ocefpaf
1. @marcelotrevisani
1. @msarahan
1. @mbargull
1. @jezdez
1. @wolfv
1. @jaimergp
1. @baszalmstra
1. @beckermr
1. @hind-m
1. @trallard

No votes (0 / 0.00%).

Abstain votes (0 / 0.00%).

Not voted (3):

1. @cj-wright
1. @mariusvniekerk
1. @jakirkham

Invalid votes (1):

1. @kkraus14

Note: Keith Kraus was incorrectly added to the voting list, but he went emeritus three months ago. The vote is thus not counted as valid. Sorry for the noise!

13/16 votes = 81.25% meets the minimum quorum. 
13/13 YES votes meets the approval threshold (60%).

PR #102, opened 2024-11-27
CEP 41: The staging output in v1 recipes / rattler-build

wolfv:
This CEP defines the `staging` output. It is a special output type that does not turn into a package. Instead, other outputs can `inherit` from it and use the "dirty" work dir to continue building a package (e.g. for different python bindings), or split the files from the cache into multiple sub-packages (headers, library, man pages ...).

One can try the cache with `--experimental` in `rattler-build`. The documentation can be found here: https://prefix-dev.github.io/rattler-build/latest/multiple_output_cache/


cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Abstract
+
+This CEP aims to define the cache output for v1 multi-output recipes.
+
+## Background
+
+Sometimes it is very useful to build some code once, and then split it into multiple build artifacts (such as shared library, header files, etc.). For this reason, `conda-build` has a special, implicit top-level build.
+
+There are many downsides to the behavior of `conda-build`: it's very implicit, hard to understand and hard to debug (for example, if an output is defined with the same name as the top-level recipe, this output will get the same requirements attached as the top-level).
```
h-vetinari, line 21: Not just that, tests under such an output will be silently skipped, which regularly trips people up: https://github.com/conda/conda-build/issues/4172

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+When computing variants and used variables, rattler-build looks at the union of a given output and the cache. That means, even if an output does not define any requirements, the cache would still add a variant for the `c_compiler`.
+
+When rattler-build executes the recipe, it will start by building the cache output that is appropriate for the current variant. This is computed by looking at all "used-variables" for the cache output and computing a "hash" for the cache. The build itself is executed in the same way as any other build.
+
+The variant keys that are injected at build time is the subset used by the cache output.
+
+When the cache build is done, the newly created files are moved outside of the `host-prefix`. Post-processing is not performed on the files beyond memoizing what files contain the `$PREFIX` (which is later replaced in binaries and text files with the actual build-prefix with a simple 1-to-1 replacement as the prefix-length are guaranteed to be equal).
+
+The cache _only_ restores files that were added to the prefix (conda-build also restored source files). Sources are re-created from scratch and no intermediate state is kept.
```
h-vetinari, line 60: That's actually a pretty serious constraint? Not being able to use a hot cache would kill certain recipes which need to run something on top of the `cache:` build - a classic example is building a library in the global stage, and the python bindings in the outputs (where the python bindings might otherwise rebuild the lib if not already present).

Another recent [example](https://github.com/conda-forge/llvmdev-feedstock/pull/297) would not work (though we decided against doing that an unrelated reason, i.e. that installing clobbered files is not guaranteed to follow the dependency relationship of outputs).

I have several other feedstocks that are installing from the intermediate build in all of the outputs, and it's unmaintainable to move these to simple `files:` directives under the current [circumstances](https://github.com/conda/conda-build/issues/5455).

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+<details> 
+  <summary>**Workaround for keeping intermediate build state**</summary>
+In order to keep state from the build around in the outputs, one could exercise the build in the $PREFIX and ignore the files in the outputs, or the build folder coule be placed in the `/tmp` folder:
+
+```bash
+cmake -B /tmp/mybuild -S .
+cmake --build /tmp/mybuild
+```
```
h-vetinari, line 69: So `/tmp` is special-cased in that it's not touched by any of the other build-logic?

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+Since this is not extremely convenient, we could discuss adding a "scratchpad" outside the `src` dir to rattler-build that will be kept between outputs and removed at the end of the whole build.
+</details>
```
h-vetinari, line 72: To double-check, this would save `$SRC_DIR` to the scratchpad in between operations, and restore it for the installation of any output that needs it?

I would go as far as saying that `cache:` implies this behaviour. I don't want to cache stuff in `$PREFIX` (then I could just install it and call that an output), I want to cache it _without_ installation.

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+  - package:
+      name: foo-devel
+    requirements:
+      run:
+        - ${{ pin_subpackage("libfoo") }}
+        - ${{ pin_subpackage("foo-headers") }}
+```
+
+Special care must be taken for `run-exports`. For example, the `${{ compiler('c') }}` package used to build the cache is going to have run-exports that need to be present at runtime for the package. To compute run-exports for the outputs, we use the union of the requirements - so virtually, the host & build dependencies of the cache are injected into the outputs and will attach their run exports to each output.
```
h-vetinari, line 98: I'm in favour of this, just noting that this diverges from conda-build, which currently does not inject REs from the global build-step to the outputs.

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+
+```bash
+cmake -B /tmp/mybuild -S .
+cmake --build /tmp/mybuild
+```
+
+Since this is not extremely convenient, we could discuss adding a "scratchpad" outside the `src` dir to rattler-build that will be kept between outputs and removed at the end of the whole build.
+</details>
+
+The new files can then be used in the outputs with the `build.files` key:
```
h-vetinari, line 74: How do I denote a dependence of a given output on the `cache:` phase? For example, we had a bunch of discussion in pytorch over https://github.com/conda-forge/pytorch-cpu-feedstock/commit/981a04d378bda35a2f2852cd0a448f0bcc9d56e6 (later reverted https://github.com/conda-forge/pytorch-cpu-feedstock/commit/54fa422e7cacbe544a2e7cf1384555aa9af13044), because it is undocumented under what conditions a build will "fan" out (e.g. build multiple python versions on top of the same library; rather than generate a job per python version)

In other words, we need to have some clarity whether an output like
```yaml
cache:
  requirements:
    build:
      - ${{ compiler('c') }}
    host:
      - # no python here
 
outputs:
  - package:
      name: foo
    requirements:
      host:
        - # depends on cache implicitly?
        - python
```

together with a CBC of 
```yaml
python:
  - 3.9
  - 3.10
  - 3.11
  - 3.12
```
will actually "fold" the python dimension into `foo` or not.

In conda-build, all outputs implicitly depend on the global build stage; perhaps we should make this explicit? `build.files` already does this partway (for outputs that only use `files:`). On the other hand, it might be more intuitive to keep the implication that `cache:` affects everything else - see discussion about `skip:`).

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+      - ninja
+    host:
+      - libzlib
+    # run: ...
+    # run_constraints: ...
+    # ignore_run_exports: ...
+  build:
+    # note: in the current implementation, all other keys for build
+    #       are also valid, but unused
+    script: build.sh
```
h-vetinari, line 44: One thing that's been a permanent annoyance with the way things have grown in conda-build is that a "global" `build.skip: ...` does not mean the entire build will get cancelled, because - in theory - the various outputs could do their own thing that doesn't rely on the package-level build. So in order to actually skip a given combination, it needs to be skipped _on all outputs_ (which quickly gets annoying in large multi-output recipes, especially for quick debugging).

If we go with the "all outputs implicitly depend on `cache:`" approach, then I think it would be good to specify that a `skip: true` on the `cache:` level completely skips the build, without having to repeat it per output.

h-vetinari (commented):
Thank you for writing this up, excited to see this moving forward! 🙏 😊

cep-cache-output.md
wolfv: So my vision for this was to have the build folder of cmake / meson etc. in a common place and we could add a new build time variable that would be constant (e.g. `$SCRATCH`).

Then your cmake invocation could look like: 

```
cd $SCRATCH
cmake $SRC_DIR -DCMAKE_INSTALL_PREFIX=$PREFIX
ninja install
```

The $SCRATCH dir could be on `/tmp` or in the work dir, but just not part of the `$SRC_DIR`, to keep it pristine.

I tried this in [this repo](https://github.com/wolfv/rattler-build-cache-test) and ran into some issue: we do create independent build folders per output that have different names. And CMake detects that and then complains, when invoked with differnet options, that the source dir doesn't match anymore:

```txt
 │ │ + cmake $SRC_DIR -GNinja -DCMAKE_INSTALL_PREFIX=$PREFIX -DBUILD_PYTHON_BINDINGS=ON
 │ │ CMake Error: The source "$SRC_DIR/CMakeLists.txt" does not match the source "/Users/wolfv/Programs/calculator/output/bld/rattler-build_libcalculator_1732782319/work/CMakeLists.txt" used to generate cache.  Re-run cmake with a different source directory.
```

We have a few options:

- Keep the same path for the "work dir" the for all outputs (this will prevent concurrent builds in the future). This does not need to imply that we restore the (dirty) work dir from the cache.
- We could also restore the "dirty" work dir from the cache. For that, we would need to keep a pristine copy of the work dir at cache time, because every output might modify or add files to the work dir. In this case, we would have to prevent every output to restore the global sources, and instead only restore the dirty work dir which already has the global sources. We would also have to keep the path to the workdir the same.

We could still have parallel builds or actions if the outputs are merely splitting the existing cache files (we could find if they have a build script or not).

cep-cache-output.md
wolfv: Hmm, I don't think the cache should have anything to do with skipping. 

The top-level can skip fine (which is intersected with the outputs).

Ie. would skip everything on win:

```
build:
  skip: [win]

outputs:
  - ... 
```

wolfv:
@h-vetinari I played around with also caching the "source" in https://github.com/prefix-dev/rattler-build/pull/1226 and the example in https://github.com/wolfv/rattler-build-cache-test.

I now think that the `cache` should have its own `source` section. Then everything else will become quite straight-forward. 

The example would then look like:

```yaml
recipe:
  name: calculator
  version: 1.0.0

cache:
  source:
    path: ../

  requirements:
    build:
      - ${{ compiler('cxx') }}
      - cmake
      - ninja
  build:
    script:
      # make sure that `alternative_name.md` is not present
      - test ! -f ./alternative_name.md
      - mkdir build
      - cd build
      - cmake $SRC_DIR -GNinja ${CMAKE_ARGS}
      - ninja install

outputs:
  # this first output will include all files installed during the cache build
  - package:
      name: libcalculator

    requirements:
      run_exports:
        - ${{ pin_subpackage('libcalculator') }}

  - package:
      name: py-calculator
    source:
      - path: ../README.md
        file_name: alternative_name.md

    requirements:
      build:
        - ${{ compiler('cxx') }}
        - cmake
        - ninja
      host:
        - pybind11
        - python
        - libcalculator

    build:
      script:
        # assert that the README.md file is present
        - test -f ./alternative_name.md
        - cd build
        - cmake $SRC_DIR -GNinja ${CMAKE_ARGS} -DBUILD_PYTHON_BINDINGS=ON
        - ninja install
```

jaimergp:
That's so close to the idea of `cache` being just a type of `output`! 🥳 

```yaml
outputs:
  - cache:
      name: optional, only needed if there are several caches
  - package:
      name: my-output
    build:
      needs-cache: bool # or cache.name
```

wolfv:
Yeah it could be done, but I think we can let it wait for v2 of the recipe format! :)

cep-cache-output.md
h-vetinari: All that sounds reasonable to me. I think I'd prefer option 2, but I don't understand though why 

> we would have to prevent every output to restore the global sources, and instead only restore the dirty work dir which already has the global sources.

is necessary. Some output may genuinely need to reconfigure something that needs the sources, even though they want to use a hot cache of the dirty work dir.

h-vetinari:
> I now think that the `cache` should have its own `source` section. Then everything else will become quite straight-forward.

That sounds fine in principle. Could you explain a bit more what the thinking behind that is though? If there's already a `source:` on recipe-level, would the `cache:` have to repeat something?

> That's so close to the idea of `cache` being just a type of `output`! 🥳

That would sound even better! More powerful and yet less divergence. (no opinion on whether this is v1 or v2 stuff)

However, this I don't like
```yaml
outputs:
  # this first output will include all files installed during the cache build
  - package:
      name: libcalculator
```
I think this should never be automatic, and always need an explicit opt-in. Either the output says `files:` (in which case it could pick up parts of what `cache:` installed), or it actually does the `ninja install` itself (rather than during `cache:`). This would also be a better match for what the example is already doing for `py-calculator`.

carterbox:
@carterbox, please review

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+<details> 
+  <summary>script build.sh default discussion</summary>
+We had some debate wether the cache output should _also_ default to `build.sh` or should not have any default value for the `script`. This is still undecided.
```
carterbox, line 49: In a multi-output recipe, do the outputs default to `build.sh`?

cep-cache-output.md
carterbox: This is unexpected. cache only caches files that are installed to $PREFIX? If you want to cache build artifacts, then you need to store them in $SCRATCH?

What about projects that do build-time source-code generation in-place? For example, cython projects which convert .pyx files into .c files. Or projects like magma which at build time duplicate the entire source from real float implementation into double and complex types (they don't use C++ templates). These generated sources will be need to be regenerated if the outputs need to reuse them (e.g. if you are trying to reuse the cache to amortize work across builds for multiple python versions)?

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+      name: foo-headers
+
+    build:
+      files:
+        - include/**
+  - package:
+      name: libfoo
+    build:
+      files:
+        - lib/**
```
carterbox, line 88: Are you going to support negative globs in `outputs.build.files` selection? For example:

```yaml
- package:
    name: foo-dev
  build:
    files:
      include:
        - lib/**
      exclude:
        - lib/**.so.*
- package:
    name: libfoo${{ soname }}
  build:
    files:
      include:
        - lib/**.so.*
```
Would be used to separate runtime libraries from development artifacts because only versioned shared objects are needed at runtime.

More recently, I've also been working on recipes that separate optional plugins into their own output. These are also typically installed to lib/

cep-cache-output.md
carterbox: I am not in favor of this because it is not an explicit behavior and not every output will need the run_exports from the cache. For example, if your devel package doesn't have any binaries, it doesn't need to depend on any compiler package exports. If you split your binaries into multiple outputs, each output may only depend on a part of what was used in `cache.requirements.host`.

I suspect that this design decision (implicit inheritance of run exports) will lead to more overdepending warnings (which are ignored by most maintainers). In contrast, if we retain the current behavior which is that each output must explicitly enumerate their dependencies, there will be overlinking errors (which are errors that stop the build) when dependencies are missing.

cep-cache-output.md
wolfv: Yes, that is currently the case. We're thinking of deprecating this though and require users to be explicit about build scripts everywhere. What would you prefer?

cep-cache-output.md
wolfv: The downside is that the outputs would then have to setup a build / host environment for nothing (just to add run exports).

What would you think of a more complex algorithm where we only attach the run exports to the lowest level output?

E.g. `cache -> pkg -> pkg-devel` where `pkg-devel -> pin_subpackage(pkg, exact=True)`?

I see the problem with the overdepending warning.

cep-cache-output.md
```diff
@@ -0,0 +1,103 @@
...
+  requirements:
+    build:
+      - ${{ compiler('c') }}
+      - cmake
+      - ninja
+    host:
+      - libzlib
+    # run: ...
+    # run_constraints: ...
+    # ignore_run_exports: ...
```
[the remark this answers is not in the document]
wolfv: You are right. `ignore_run_exports` makes sense though.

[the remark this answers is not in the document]
carterbox: > ignore_run_exports makes sense though.

Why? Could you please explain why you want run_exports from the cache build to all of the outputs?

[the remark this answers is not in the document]
wolfv: I elaborated on other options further below. We need to attach run exports from the cache to at least 1 output that uses the cache, otherwise there will be bugs.

cep-cache-output.md
h-vetinari: > The downside is that the outputs would then have to setup a build / host environment for nothing (just to add run exports).

Yeah, and this gets easily forgotten. I understand the issue with over-imbuing outputs with dependencies, but in this case I think the safer default is to inherit the run-exports, unless the output already has its own build/host environment (because then it's safe to assume that this should determine the actual REs for that output).

Part of my thinking on that is that, in many cases, superfluous REs are pretty benign, e.g. it's basically irrelevant whether a metapackage that does `pin_subpackage(pkg)` has its own REs, when `pkg` already has those run-exports anyway.

So while I generally dislike recipes having to _opt out_ of implicit things, I think in this case it's the safer/saner default.

> What would you think of a more complex algorithm where we only attach the run exports to the lowest level output?
> 
> E.g. `cache -> pkg -> pkg-devel` where `pkg-devel -> pin_subpackage(pkg, exact=True)`?

Can you explain the algorithm in a sentence? I think we need to have understandable rules. So far I think "inherit REs unless output has its own build/host env" is relatively easy to explain (aside from checking the technical boxes from my POV), and a more complicated algorithm would need to provide a pretty big improvement in comparison to be worth it IMO.

carterbox: The background section of this proposal specifically calls out implicit behavior as a "downside" of the top-level package. In my experience, confusion about multi-output packages comes from unexpected behavior when the top-level shares a name as one of the output packages so they aren't behaving independently. This is why I maintain that there should be no run_exports across package boundaries from the cache package to output package because creating more interaction between output packages and the top-level package will increase confusion.

Adding a non-trivial run_export tracking algorithm from the cache package to the output packages will not further the design goal of making top-level package cache easier to understand and use. It will only make recipes less verbose. It will be adding complexity and adding implicit behaviors which this proposal claims to dislike. If the goal of this proposal is to make the recipe format more convenient or less verbose, then the background section should claim that instead of claiming that the problem is reader comprehension and implicit behavior.

In my opinion, it's easier for users to understand if all outputs inherit none of the run_exports from the cache package. The easiest way to explain the cache package would be that the cache package is a binary redistribution that you are unpacking for each of the output packages. Renumerating host/build dependencies is verbose, but it's easy to conceptualize as "If I was building only the things in this specific output, what would I need?"

I think I've made all my points about this topic, so I will stop trying to argue in favor of it.

cep-cache-output.md
wolfv: This was fixed - the dirty source dir is now also cached :)

cep-cache-output.md
wolfv: This was fixed by storing the (dirty) source dir and restoring it.

cep-cache-output.md
wolfv: I see your points. I don't want this discussion to become too heated - would you be interested in chatting on this point (and perhaps others) over video next week?

We have very precise reporting in rattler-build for the run exports - in the log you would see "run dependencies" and then `foobar (RE from host-cache)` which would indicate that this run export came from the host section of the cache.

I think run-exports themselves are, to some degree, "hidden implicit" stuff that we are trying to surface in rattler-build by highlighting them in the logs and indicating where they came from.

carterbox: This is a CEP, so the appropriate meeting is the Conda Community Sync? I can attend this week. Feb 12 @ 11am Chicago.

wolfv: Yes!

seberg: From the side-lines since I was just bitten by this wanting to reduce duplication while splitting the recipe.  (Trying to cargo cult from a recipe that was very conscious about the split requirements to get the right dependencies.)

I can understand that implicitly adding it to all dependencies build on the cache is also odd even if the failure mode is more graceful.

Maybe that doesn't fit the pattern at all, but how about *forcing* users to be explicit about it?  So the user *must* either indicate which named outputs they want to export the caches dependencies to or indicate that they want to export it to none

You could go as far as allowing `ignore_run_exports` that are specific for a named output here (to keep them paired with the dependencies):
```
- export_run_exports:
  split_name1
  split_name2
- export_run_exports:
  split_name3
  - ignore_run_exports:
    # maybe even allow ignoring of the ones that come from cache here?
```
(I doubt this is valid yaml, but you get the point)

wolfv: That is not a bad idea! We could have an `cache.attach_run_exports_to: [ ... ]` field and we can require it to have at least one entry (at least by linting).

wolfv: @carterbox what are your feelings about this solution?

wolfv:
We just had a discussion at the build tools meeting with new ideas for the name "cache".

One idea was "prepare" / "prepare-step" / or something along these lines.
I was also thinking of "pre-build-step", or "pre_build" or something like this.

If anyone has another idea, would love to hear it. Cache is a bit too overloaded :)

wolfv:
These are ideas from my trusty assistant:

prepare
preprocess
common_build
shared_step
foundation
base_build
intermediate
scaffold
staging
preflight
common_prep
shared_build
core_build

carterbox:
What about something that includes the word "outputs" or "multioutput" so it's more obvious that you would only use this section when creating a multi-output recipe. For example:

```yaml
recipe:
  name: ...
  version: ...

outputs_prepare: ...

outputs:
  - package: ...
  - package: ...
```

"prepare_outputs", "pre_outputs", "multioutput_cache", "outputs_prebuild", "outputs_scaffold", "outputs_staging", ...

jaimergp:
Yea, I like where @carterbox is going. One more: `before_outputs`.

h-vetinari:
How about `shared_build` or perhaps `outputs_shared_build`...? Actually, if we want to emphasise the association with outputs, we could go one step further, and put it under `outputs:` directly:

```yaml
recipe:
  name: ...
  version: ...

outputs:
  - shared:  # or common, shared_build, ...
  - package: ...
  - package: ...
```

wolfv:
I think moving it under outputs is a nice idea. Main difficulty I see is that people might expect to be able to create more than 1 of those which is not possible right now (maybe later we can allow naming them and having multiple, though). But we can error on that, too ...

carterbox:
The cache is not itself an output though, so we would need to choose a name which indicates as such and doesn't end with the word "build". Otherwise, we would have an outputs.build.build section.

```yaml
recipe:
  name: ...
  version: ...

outputs:
  - definitely_not_an_output_but_used_to_make_outputs:
      source: ...
      requirements:
        build: ...
        host: ...
      build: ...
  - package: ...
  - package: ...
```
😆 But seriously... something like setup, staging, scaffold, precursor, bootstrap, mock.

wolfv:
Hi @minrk and sorry for the longer radio silence on this issue. I am excited that you and @isuruf separately reached out to push on this. It has been high on our TODO list for a while now and we definitely want to get to it. 

We had some thoughts in the meantime about making an explicit "inherit" key.

That would make a recipe look something like this:

```yaml
outputs:
  - cache:
      name: foo-cache   # required, string, follows rules of `PackageName`

    source:
      - url: https://foo.bar/source.tar.bz
        sha256: ...

    requirements:
      build:
        - ${{ compiler('c') }}
        - cmake
        - ninja
      host:
        - libzlib
        - libfoo
      # the `run` and `run_constraints` sections are not allowed here
      ignore_run_exports:
        by_name:
          - libfoo

    build:
      # only the script key is allowed here
      script: build_cache.sh

  - package:
      name: foo-headers
    
    # long form of newly added `inherit` key
    inherit:
      from: foo-cache
      run_exports: false

    build:
      files:
        - include/

  - package:
      name: foo
    
    # short form, inherits run exports by default
    inherit: foo-cache

    # will bundle everything else except files already present in `foo-headers`
    requirements:
      host:
        - ${{ pin_subpackage("foo-headers", exact=True) }}
```

In this case, inheritance would be optional, and inheriting from the `run_exports` could be controlled precisely as well. We could potentially also allow a cache output to inherit from another cache output. And multiple inheritance is also something that could be potentially added (e.g. `inherit: [{ from: foobar }, { from: bizbar }]`).

Unfortunately this would be a bit of a bigger refactor in rattler-build and we haven't really "dared" to do this, yet. 

This is a longer writeup of the idea: https://gist.github.com/wolfv/1a160129eb7ba7ccaab51fdf25d7e89b

minrk:
inheriting requirements seems fine, to me. When explicitly inheriting from the cache, it makes sense to apply the exact pin. When re-stating the same requirement, it makes sense to re-solve without regard for the cache, so I don't think the single joint solve or prioritized solve is something we should do, there are too many ways it could lead to problems, I think, and too complex for a potentially multi-implementation spec.

How would an output specify that it inherits one or more dependencies (from host and/or build)? Would a jinja function be appropriate `${{ pin_cache_dependency('name') }}`?

I think it makes sense to sketch a recipe where some outputs:

1. inherit all cache requirements
2. inherit a subset of requirements, but want to ensure it's the same version
3. use a different version or variant of a requirement in the cache (e.g. mutually exclusive mpi or blas provider)

isuruf:
IMO inheriting requirements is a bad idea in general because it will result in cryptic solve errors. 1 and 3 of @wolfv's suggestions will lead to solve errors. I don't understand 2 though.

minrk:
I also like the `inherit` proposal, if that's where consensus is pointing.

carterbox:
I think your example is incorrect. It's missing a dependency from the python module on the C++ library.
```
outputs:
  - cache: foobar-compiles
    requirements:
      build:
        - ${{ compiler('cxx') }}
        - cmake
        - ninja
      host:
        - libpng
  - package:
      name: libfoobar
    inherit:
      from: foobar-compiles
    requirements:
      run:
        - somelib
  - package:
      name: py-foobar
    inherit:
      from: foobar-compiles
    requirements:
      build:
        - ${{ compiler('cxx') }}
        - cmake
      host:
        - libfoobar  # this was missing
        - libpng     # only needed if py-foobar links directly
        - python
        - pip
      run:
        - libfoobar  # this was missing; run_exports don't work within a recipe
        - python
        - somelib
```
And if that's the case, then wouldn't the addition of libfoobar to the host environment of py-foobar constraint the host environment of py-foobar to be something compatible? Or you point isn't about output compatibility, but instead about brevity in syntax?

wolfv:
Sorry, yes, you're right. Although I think in this example you might need libpng in both the cache and output host to make a new cmake invocation work fine. That's the duplication we could avoid (also for compiler, cmake, etc)

h-vetinari:
> Thanks to this refactor, we now have the idea of `staging` outputs and inheritance implemented 🎉

Nice, congratulation! 🥳 

Can you provide some examples here of how the syntax looks now? I assume you didn't incorporate the `inherit_exports:` [idea](https://github.com/conda/ceps/pull/102#issuecomment-3684307924) above; if so, could you please comment on it?

msarahan (approved):
This is a big improvement over conda-build's implicit design. Thank you for thinking through it.

cep-XXXX.md
```diff
@@ -0,0 +1,163 @@
+# CEP XXXX - The staging output in v1 recipes / rattler-build
+
+<table>
+<tr><td> Title </td><td> The staging output in v1 recipes / rattler-build </td>
+<tr><td> Status </td><td> In Discussion </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &ltw.vollprecht@gmail.com&gt; </td></tr>
+<tr><td> Created </td><td> Nov 27, 2024</td></tr>
+<tr><td> Updated </td><td> </td></tr>
```
```suggestion
<tr><td> Updated </td><td> Mar 26, 2026 </td></tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,163 @@
...
+<tr><td> Discussion </td><td>  </td></tr>
```
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/102 </td></tr>
```

cep-0041.md
```diff
@@ -0,0 +1,163 @@
+# CEP XXXX - The staging output in v1 recipes / rattler-build
+
+<table>
+<tr><td> Title </td><td> The staging output in v1 recipes / rattler-build </td>
+<tr><td> Status </td><td> In Discussion </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &ltw.vollprecht@gmail.com&gt; </td></tr>
+<tr><td> Created </td><td> Nov 27, 2024</td></tr>
+<tr><td> Updated </td><td> </td></tr>
+<tr><td> Discussion </td><td>  </td></tr>
+<tr><td> Implementation </td><td> rattler-build </td></tr>
```
jaimergp, line 10: It would be nice if you could add some permalinks (hash-pinned) here to some of the main modules in rattler-build.

cep-XXXX.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Abstract
+
+This CEP aims to define the staging output for v1 multi-output recipes.
+
+## Background
+
+Sometimes it is very useful to build some code once, and then split it into multiple build artifacts (such as shared library, header files, etc.). For this reason, `conda-build` has a special, implicit top-level build.
+
+There are many downsides to the behavior of `conda-build`: it's very implicit, hard to understand and hard to debug (for example, if an output is defined with the same name as the top-level recipe, this output will get the same requirements attached as the top-level).
```
```suggestion
There are many downsides to the behavior of `conda-build`: it's too implicit, hard to understand, and hard to debug. For example, if an output is defined with the same name as the top-level recipe, this output will get the same requirements attached as the top-level.
```

cep-0041.md
```diff
@@ -0,0 +1,163 @@
...
+## Specification
+
+A recipe can have zero or more staging outputs.
+
+A staging output looks as follows:
+
+```yaml
+outputs:
+  - staging:
+      name: foo-cache   # required, string, follows rules of `PackageName`
```
jaimergp, line 86: Which rules are these? Encoded by a particular CEP?

cep-XXXX.md
```diff
@@ -0,0 +1,163 @@
...
+    inherit: foo-cache
+
+    # will bundle everything else except files already present in `foo-headers`
+    requirements:
+      host:
+        - ${{ pin_subpackage("foo-headers", exact=True) }}
+```
+
+> [!WARNING]
+> When using `outputs` we are going to remove the implicit `build.script` pointing to `script.sh`. Going forward, the script name / content has to be set explicitly.
```
jaimergp: This doesn't sound like it belongs in the specification of the CEP. Perhaps in further work, if anywhere. I'd just remove it.

cep-XXXX.md
```diff
@@ -0,0 +1,163 @@
...
+
+There are many downsides to the behavior of `conda-build`: it's very implicit, hard to understand and hard to debug (for example, if an output is defined with the same name as the top-level recipe, this output will get the same requirements attached as the top-level).
+
+For the v1 spec we are attempting to formalize the workings of the "top-level" build. For this, we introduce a new `staging` output, that has the same values as a regular output, but does not produce a package artifact. Instead, we keep changes from the `staging` output in a temporary location on the filesystem and restore from this checkpoint when building other outputs that _inherit_ from this `staging` cache.
+
+## Specification
+
+A recipe can have zero or more staging outputs.
+
+A staging output looks as follows:
```
jaimergp: This needs to be encoded in a more accurate way. It's not clear what "looks as" means. I recommend you use RFC language (MUST, MAY, SHOULD, etc) for the required keys. You may build on existing CEPs to simplify language, but the specification should be as specific (sorry to be redundant) as possible.

e.g.:

> As per CEP XXX, `outputs` MUST be a list of dictionaries, each representing an artifact. We extend that schema with an alternative dictionary type, which MUST comply to this schema:
> 
> - `staging` MUST be a required key, whose value follows the same schema as `package` in CEP XX`.
> - Everything else follows the schema described in CEP X, Y and Z.
>
> The existing `package` outputs MAY replace the `build` key with a `inherit` key, whose value MUST comply to this schema:
> ...

jaimergp (dismissed):
The ideas are sound and the proposal itself is solid, but the text right now does not conform to the quality standards we have set for the other CEPs. I know this was started before the recent wave of standardization, but I think we should strive to keep our own bar high. I suggest you rework the text by following the proposed CEP 0 template and fulfil the specification details with [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119). If you need help, I am happy to give a hand with the restructuring!

jaimergp:
Hey @wolfv, just a heads up @mgorny will open a PR against your branch with the necessary changes 🙏

cep-0041.md
jaimergp: cc @wolfv

jaimergp:
Let's fix the pre-commit lint and we are good to start the vote. Thanks @mgorny!

cep-0041.md
mgorny: FYI, I've changed this to say "same as `package.name`".

baszalmstra:
Dear @conda/steering-council,

The vote for this CEP has started. It will be open for two weeks, until **April 16nd, 2026, 23:59 Anywhere on Earth**. This time period has been chosen to make it eligible for time-out rules. As an Enhancement Proposal vote, it requires 60% affirmative votes to pass.

- Uwe Korn (@xhochy)
  - [x] yes
  - [ ] no
  - [ ] abstain
  - [x] yes
  - [ ] no
  - [ ] abstain
- Cheng H. Lee (@chenghlee)
  - [ ] yes
  - [ ] no
  - [x] abstain
- Filipe Fernandes (@ocefpaf)
  - [x] yes
  - [ ] no
  - [ ] abstain
  - [x] yes
  - [ ] no
  - [ ] abstain
- Michael Sarahan (@msarahan)
  - [x] yes
  - [ ] no
  - [ ] abstain
- Marcel Bargull (@mbargull)
  - [ ] yes
  - [ ] no
  - [ ] abstain
- John Kirkham (@jakirkham)
  - [ ] yes
  - [ ] no
  - [ ] abstain
- Jannis Leidel (@jezdez)
  - [ ] yes
  - [ ] no
  - [x] abstain
- Wolf Vollprecht (@wolfv)
  - [x] yes
  - [ ] no
  - [ ] abstain
  - [x] yes
  - [ ] no
  - [ ] abstain
- Bas Zalmstra (@baszalmstra)
  - [x] yes
  - [ ] no
  - [ ] abstain
- Hind Montassif (@Hind-M)
  - [x] yes
  - [ ] no
  - [ ] abstain
- Pavel Zwerschke (@pavelzw)
  - [x] yes
  - [ ] no
  - [ ] abstain

baszalmstra:
# Vote results

🗳️ The votes are in! Here is the tally for this Steering Council vote. Thanks to everyone who participated.

🎉 **Result: PASSED**

## Summary

| Metric | Value | Threshold | Status |
|---|---|---|---|
| Quorum (participation) | 12/14 (85.71%) | ≥ 9 voters | ✅ |
| Approval (yes / yes+no) | 10/10 (100.00%) | ≥ 60% | ✅ |

_Abstentions count toward quorum but are excluded from the approval fraction._

## Tally

- ✅ **Yes:** 10 (100.00% of decisive)
- ❌ **No:** 0 (0.00% of decisive)
- 🤷 **Abstain:** 2
- 💤 **Not voted:** 2
- ⚠️ **Invalid:** 0

### ✅ Yes

- Uwe Korn (@xhochy)
- Filipe Fernandes (@ocefpaf)
- Michael Sarahan (@msarahan)
- Wolf Vollprecht (@wolfv)
- Bas Zalmstra (@baszalmstra)
- Hind Montassif (@Hind-M)
- Pavel Zwerschke (@pavelzw)

### ❌ No

_None_

### 🤷 Abstain

- Cheng H. Lee (@chenghlee)
- Jannis Leidel (@jezdez)

### 💤 Not voted

- Marcel Bargull (@mbargull)
- John Kirkham (@jakirkham)

### ⚠️ Invalid

_None_

jaimergp:
Awesome! I'll work on the minting shortly.

Issue #59, opened 2023-07-15
Supporting microarchitecture-specific builds


chenghlee:
In various cases, users would like to take advantage of various CPU-specific features or instructions for improved performance, e.g., AVX\* or AES/crytography.

xref issues:
- conda-forge [archspec-enabled packages](https://github.com/conda-forge/conda-forge.github.io/issues/1261)

ltalirz:
Other relevant links:
- x86_x64_vx in archspec: https://github.com/archspec/archspec-json/blob/2d97655b19366ba8e836c1190a3807b57888f57f/cpu/microarchitectures.json#L243

isuruf:
With `__archspec` you can create a meta-package like `x86_64_feature_level` that depend on specific `__archspec` values and downstreams will depend on `x86_64_feature_level`.

I think a CEP is needed for how to handle virtual packages in a conda-build sense. Currently, virtual packages are one of the main things from a build machine that affect compilation.

For eg: let's say A was built with AVX512f and B needs A built with AVX512f. Then `B` can be compiled only on a build machine with `AVX512f`. We can workaround by setting `CONDA_OVERRIDE_ARCHSPEC`, but then, `BUILD_PREFIX` also gets that architecture and if there is a build tool relying on AVX512f, that build tool cannot be run.
Therefore `CONDA_OVERRIDE_ARCHSPEC` should affect only `host` and not `run`. (Even `host` is problematic when considering `python` etc). This is also an issue with cross-compilation.

ltalirz:
Two questions/thoughts:

1) What is the minimum microarchitecture of your current x86_64 CI build hardware (not the one you choose to build for)? 

2) Would it be acceptable to do cross-microarchitecture-compiles, and let the feedstock maintainers who want this feature worry about making sure that they continue to compile any build tools with the lowest-common-denominator microarchitecture? It would unfortunately mean that one cannot run the tests for these builds, but it may still be better than people who want this feature running off into different directions.

isuruf:
1. Ivybridge

2. Python blurs the line with build tools. For eg: `numpy` needs to be importable to get `numpy.get_include()` which means numpy needs to be built with the lowest-common-denominator architecture.

h-vetinari:
Very interested in this for arrow, [faiss](https://github.com/conda-forge/faiss-split-feedstock/issues/23), [pillow](https://github.com/conda-forge/pillow-feedstock/issues/103), onednn, etc.

[...]

PR #103, opened 2024-12-17
Write CEP about virtual packages

jaimergp:
Motivated by this request: https://github.com/conda/conda/issues/14443.

### DIFF
--- a/cep-????.md
+++ b/cep-????.md
@@ -1,0 +1,100 @@
+# CEP ???? - Virtual packages
+
+<table>
+<tr><td> Title </td><td> Virtual packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Dec 17, 2024</td></tr>
+<tr><td> Updated </td><td> Dec 17, 2024</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> Several </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+  "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+  described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+This CEP standardizes which virtual packages MUST be offered by conda solvers.
+
+## Specification
+
+A virtual package is defined as a package record with three fields: name, version and build string.
+The name MUST start with double underscore. The version and build string MUST follow the same semantics as in regular package records.
+
+### List of virtual packages
+
+In alphabetical order, every conda client MUST support the following virtual packages:
+
+- `__archspec`
+- `__cuda`
+- `__glibc`
+- `__linux`
+- `__osx`
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. The build string SHOULD reflect the detected CPU microarchitecture. If it cannot be detected, the build string SHOULD be `0`.
+
+The build string MUST be overridable with the `CONDA_OVERRIDE_ARCHSPEC` environment variable, if set to a non-empty value.
+
+#### `__cuda`
+
+This virtual package MUST be present when the system exhibits GPU drivers compatible with the CUDA runtimes. When available, the version value MUST be set to the oldest CUDA version supported by the detected drivers (i.e. the formatted value of `libcuda.cuDriverGetVersion()`), constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value.
+
+#### `__glibc`
+
+This virtual package MUST be present when the native platform is `linux-*`. Its version value MUST be set to the system `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_GLIBC` environment variable, if set to a non-empty value.
+
+If the `libc` version could not be estimated (e.g. the tool is not running on Linux), the tool SHOULD provide a default value (e.g. `2.17`) and inform the user of that choice and its possible overrides; e.g. via `CONDA_OVERRIDE_GLIBC`, a CLI flag or a configuration file. The environment variable MUST be ignored when the target platform is not `linux~-*`.
+
+#### `__linux`
+
+This virtual package MUST be present when the target platform is `linux-*`. Its version value MUST be set to the Linux kernel version, constrained to two to four numeric components formatted as `{major}.{minor}.{micro}.{patch}`. If the version cannot be estimated (e.g. because the native platform is not Linux), the fallback value MUST be set to `0`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_LINUX` environment variable, if set to a non-empty value that matches the regex `"\d+\.\d+(\.\d+)?(\.\d+)?"`. The environment variable MUST be ignored when the target platform is not `linux-*`.
+
+#### `__osx`
+
+This virtual package MUST be present when the target platform is `osx-*`. Its version value MUST be set to the first two numeric components of macOS version formatted as `{major}[.{minor}]`. If the version cannot be estimated (e.g. because the native platform is not macOS), the fallback value MUST be set to `0`. The build string MUST be `0`. 
+
+The version MUST be overridable with the `CONDA_OVERRIDE_OSX` environment variable. If this environment variable is set to the empty string `""`, then the `__osx` virtual package MUST NOT be present. The environment variable MUST be ignored when the target platform is not `osx-*`.
+
+#### `__unix`
+
+This virtual package MUST be present when the target platform is `linux-*`, `osx-*` or `freebsd-*`. The version and build string fields MUST be set to `0`.
+
+The version or build string fields MUST NOT be overriden by the `CONDA_OVERRIDE_UNIX` environment variable. However, if this environment variable is set to a non-empty value, the `__unix` virtual package MUST be present if otherwise if would not have been. The environment variable MUST be ignored when the target platform is not `linux-*`, `osx-*` or `freebsd-*`.
+
+#### `__win`
+
+This virtual package MUST be present when the target platform is `win-*`. The version MUST be set to the first three numeric components of the version reported by CMD's `ver` command, formatted as `{major}.{minor}.{build}`. If the version cannot be estimated (e.g. because the target platform does not match the native platform), the fallback value MUST be set to `0`. The build string MUST be `0`. 
+
+The version MUST be overridable with the `CONDA_OVERRIDE_WIN` environment variable. If this environment variable is set to the empty string `""`, then the `__win` virtual package MUST NOT be present. The environment variable MUST be ignored when the target platform is not `win-*`.
+
+## Motivation
+
+Virtual packages are used to expose details of the system configuration to a conda client. They are commonly used as dependencies in regular packages to constrain on which systems they can be installed. Some examples include:
+
+* On Linux, the minimum `libc` version that must be available in the system via the `__glibc` virtual package.
+* The oldest macOS version compatible with the package via the `__osx` virtual package.
+* Whether a `noarch` package should be constrained to a single operating system via the `__linux`, `__osx` or `__win` virtual packages (often with no version).
+* The minimum CPU microarchitecture level that the binaries require via the `__archspec` virtual package.
+* The lowest CUDA version the GPU driver is compatible with via `__cuda`.
+
+## References
+
+* [Virtual packages implementation in `conda/conda` 24.11.1](https://github.com/conda/conda/tree/24.11.1/conda/plugins/virtual_packages)
+* [ENH: make `__win` version usable for package metadata (conda/conda#14443)](https://github.com/conda/conda/issues/14443)
+* [Drop `CONDA_OVERRIDE_WIN` environment variable (mamba-org/mamba#2815)](https://github.com/mamba-org/mamba/pull/2815)
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).


cep-????.md
```diff
@@ -0,0 +1,127 @@
...
+- `__linux`
+- `__osx`
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. The build string SHOULD reflect the detected CPU microarchitecture. If it cannot be detected, the build string SHOULD be `0`.
+
+The build string MUST be overridable with the `CONDA_OVERRIDE_ARCHSPEC` environment variable, if set to a non-empty value.
```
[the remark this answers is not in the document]
jaimergp: That's how I started but as you will see, there are so many small details that are different across overrides: whether version or build is overridden, whether a certain format is expected, whether the empty string has a special meaning... I thought it was going to be a quick CEP and ha! 😂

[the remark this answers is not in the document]
jakirkham: Maybe we can just call out the name and say generally how this is used with the exception that in some cases it is handled differently (like `__unix` or historically `__win`)

cep-????.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### `__cuda`
+
+This virtual package MUST be present when the system exhibits GPU drivers compatible with the CUDA runtimes. When available, the version value MUST be set to the oldest CUDA version supported by the detected drivers (i.e. the formatted value of `libcuda.cuDriverGetVersion()`), constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value.
+
+#### `__glibc`
+
+This virtual package MUST be present when the native platform is `linux-*`. Its version value MUST be set to the system `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. The build string MUST be `0`.
```
jakirkham, line 52: In case we ever add `__musl`, etc.

```suggestion
This virtual package MUST be present when the native platform is `linux-*`. Its version value MUST be set to the system GNU `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. The build string MUST be `0`.
```

cep-????.md
```diff
@@ -0,0 +1,129 @@
...
+
+The `libc` version can be computed via:
```
jakirkham, line 54:
```suggestion
The GNU `libc` version can be computed via:
```

cep-????.md
```diff
@@ -0,0 +1,129 @@
...
+
+- Python's `os.confstr("CS_GNU_LIBC_VERSION")`
+- `getconf GNU_LIBC_VERSION`
+- `ldd --version` (in GLIBC distros)
+- System's `libc.so` (in MUSL distros, location not standardized)
```
jakirkham, line 59: Think MUSL should be its own thing. It is versioned differently. Plus users presumably want to distinguish GLIBC built packages from MUSL ones

cep-????.md
```diff
@@ -0,0 +1,129 @@
...
+- Python's `platform.mac_ver()[0]`
+- `sw_vers -productVersion`
+
+> If applicable, the `SYSTEM_VERSION_COMPAT` workaround MUST NOT be enabled; e.g. the version reported for Big Sur should be 11.x and not 10.16.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_OSX` environment variable. If this environment variable is set to the empty string `""`, then the `__osx` virtual package MUST NOT be present. The environment variable MUST be ignored when the target platform is not `osx-*`.
+
+#### `__unix`
+
+This virtual package MUST be present when the target platform is `linux-*`, `osx-*` or `freebsd-*`. The version and build string fields MUST be set to `0`.
```
jakirkham, line 92: In the future, would we want `__freebsd` or `__netbsd`?

cep-????.md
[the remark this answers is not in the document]
jaimergp: Added a paragraph before this section

cep-????.md
h-vetinari: With the `stdlib` infrastructure, we have the technical requirements in place to do something like:
```yaml
c_stdlib:               # [linux]
  - sysroot             # [linux]
  - musl                # [linux]
c_stdlib_version:       # [linux]
  - 2.17                # [linux]
  - 1.2                 # [linux]
zip_keys:               # [linux]
  - - c_stdlib          # [linux]
    - c_stdlib_version  # [linux]
```
The question whether we want to roll out a "conda-forge for musl" is of course a separate question.

For the purpose of this CEP, definitely `__glibc` needs to stay musl-free IMO. For future-proofing, we could add `__musl`, but no strong feelings.

jaimergp: This is the GLIBC version in MUSL distros, not the MUSL version itself. Happy to adjust the wording though. See [this block in the constructor SH templates](https://github.com/conda/constructor/blob/f6f8d62125665ceb72a6f0c3280a09be1fff8534/constructor/header.sh#L36-L60) for some more context.

cep-????.md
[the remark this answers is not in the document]
jakirkham: Looks great! Thanks Jaime 🙏

cep-????.md
jaimergp: I don't see why not! Maybe we should check which [platform prefixes](https://github.com/conda/conda/blob/d1584ce42d606899c0468cae1fdb2ed67bf3f4a7/conda/base/constants.py#L69-L89) we want to expose via virtual packages, and whether we want to also expose the architectures at a macro level (the 2nd part of a subdir).

jaimergp: (I'd slightly prefer to defer that to a different CEP, though)

cep-????.md
h-vetinari: Having glibc in musl is certainly not the [default](https://wiki.alpinelinux.org/wiki/Running_glibc_programs), I actually didn't know this is supported - TIL. In that case I agree it makes sense to report the glibc that's found.

So my point was that we could also build musl-native libs, but I guess that's much less urgent if alpine-users can just install a glibc compat (assuming this works with our glibc-flavoured packages).

cep-????.md
```diff
@@ -0,0 +1,131 @@
...
+- `__cuda`
+- `__glibc`
+- `__linux`
+- `__osx`
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. The build string SHOULD reflect the detected CPU microarchitecture. If it cannot be detected, the build string SHOULD be `0`.
```
jaimergp, line 42: I think we'll need to standardize the acceptable values for the build string here, but that's currently tied to the `archspec` Python package implementation. Or is that list based on something already standardized? 🤔 

Also, there will be more microarchitectures in the future... will we keep updating the list? Or, alternatively, should we just defer that to whatever frontend end-users end up using ("in the end" just to use the word "end" one more time)?

cep-????.md
jakirkham: Reading the original comment ( https://github.com/conda/constructor/issues/850#issuecomment-2343756454 ), think that user is referring to [`gcompat`]( https://github.com/Stantheman/gcompat?tab=readme-ov-file#introduction ):

> ...a library which provides glibc-compatible APIs for use on musl libc systems.

While that can be useful, think we should not encode that in this CEP

It may make sense to have a CEP extending this for MUSL and this could be one detail we include there, but would suggest we leave it out of this version and save it as future work

cep-????.md
jakirkham: We could list some as examples and leave the rest out

That said, maybe it is a good exercise to go through updating the CEP before updating tooling, etc.?

Would be interested to hear what others think on this question. It is a good one

cep-????.md
jaimergp: I think I need to reword these suggestions. They are not meant to tell anyone where to get their versions from, merely inform that's where some of the existing implementations have found them, as a suggestion...

jaimergp: > I think that user is referring to [gcompat](https://github.com/Stantheman/gcompat?tab=readme-ov-file&rgh-link-date=2024-12-17T23%3A01%3A03Z#introduction):

I'm not sure the original report was about `gcompat`, but about this alpine-focused [`glibc` version](https://github.com/sgerrand/alpine-pkg-glibc).

jakirkham: The challenge is if we include this line (even as an example) future readers will look at this line and say "Hey that is how they said we should be supporting MUSL." Even if that is not what we meant. So taking this line out is safer than keep it IMO

The rest of the examples above seem fine

h-vetinari: > @jakirkham: think that user is referring to [`gcompat`](https://github.com/Stantheman/gcompat?tab=readme-ov-file#introduction):

Look at the alpine wiki link I posted:
> [wiki] If you want to run [glibc](https://www.gnu.org/software/libc/) programs in Alpine Linux, there are a few ways of doing so.

gcompat is one of them, but there are others.

> @jakirkham: future readers will look at this line and say "Hey that is how they said we should be supporting MUSL."

I don't think that's realistic; it's also easy to preempt - just say "the version of glibc, if present (musl-based distros like alpine only provide this for compatibility; their main C library is musl itself)"

jakirkham: Sorry Axel think we had posted around the same time. Agree there can be other ways people try to handle this need

<hr>

To loop back to Jaime's original point on how to word this, what if we just say the following

```suggestion
- System's GNU `libc.so` (see specific disto for location)
```

Note: Some distros put `libc.so` in different places. So this is worth stating in its own right

cep-????.md
jaimergp: Added a new section with scope and future recommendations, thanks!

cep-????.md
jaimergp: I reworded these paragraphs a bit including the intent of the suggestions, let me know what you think!

cep-????.md
jaimergp: Expanded this quite a bit now.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+  described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+This CEP standardizes which virtual packages MUST be offered by conda solvers.
+
+## Specification
+
+A virtual package is defined as a package record with three fields: name, version and build string.
+The name MUST start with double underscore. The version and build string MUST follow the same semantics as in regular package records.
```
jakirkham, line 24:
```suggestion
The name MUST start with double underscore (`__`). The version and build string MUST follow the same semantics as in regular package records.
```

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
```
jakirkham, line 25: This seems to be a common feature. So let's refactor it out here and drop from the sections below.

```suggestion

The build string MAY be zero (`0`). Some exceptions apply. See below.

```

Also noticed some caveats. So used the same formula as with the environment variable `CONDA_OVERRIDE_*`

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+The version MUST be overridable with the `CONDA_OVERRIDE_GLIBC` environment variable, if set to a non-empty value.
+
+If the GNU `libc` version could not be estimated (e.g. the tool is not running on Linux), the tool SHOULD provide a default value (e.g. `2.17`) and inform the user of that choice and its possible overrides; e.g. via `CONDA_OVERRIDE_GLIBC`, a CLI flag or a configuration file. The environment variable MUST be ignored when the target platform is not `linux~-*`.
+
+> The GNU `libc` version can be computed via:
+>
+> - Python's `os.confstr("CS_GNU_LIBC_VERSION")`
+> - `getconf GNU_LIBC_VERSION`
+> - `ldd --version`. Note this only applies to GLIBC distros. In MUSL distros this will return the MUSL version instead. In these systems, you might need to locate the GNU `libc.so` library and call it directly.
```
jakirkham, line 85: On Alma 8 (using GLIBC) it shows the following

```bash
$ ldd --version
ldd (GNU libc) 2.28
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper.
```

On Alpine 3 (using MUSL) it shows the following

```bash
$ ldd --version
musl libc (aarch64)
Version 1.2.5
Dynamic Program Loader
Usage: /lib/ld-musl-aarch64.so.1 [options] [--] pathname
```

So the `libc` implementation is being reported in both cases.

Maybe we can just say something like

```suggestion
> - `ldd --version`. Please verify that it references GNU libc or GLIBC. For non-standard installs, using a GLIBC compatibility layer, this may require locating the implementation and directly querying.
```

Am still not sure we should be volunteering to support these non-standard installs in this spec though

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Motivation
+
+Virtual packages are used to expose details of the system configuration to a conda client. They are commonly used as dependencies in regular packages to constrain on which systems they can be installed. Some examples include:
+
+* On Linux, the minimum `libc` version that must be available in the system via the `__glibc` virtual package.
+* The oldest macOS version compatible with the package via the `__osx` virtual package.
+* Whether a `noarch` package should be constrained to a single operating system via the `__linux`, `__osx` or `__win` virtual packages (often with no version).
+* The minimum CPU microarchitecture level that the binaries require via the `__archspec` virtual package.
+* The lowest CUDA version the GPU driver is compatible with via `__cuda`.
```
jakirkham, line 139: Should we move this higher up? Or does it need to go here?

Think it is usually best to start with this high-level information before diving into the details. It helps the audience better understand why this matters before trying to understand the nitty gritty details

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Rationale
```
jakirkham, line 141: Maybe?

```suggestion
## Potential future work
```

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+This CEP attempts to describe and standardize the existing implementations of the currently available virtual packages. The following items are deliberately kept out of scope and recommended for discussion in future CEPs:
```
jakirkham, line 143: Trying to think of a way to streamline this and turn this into a bit of call to action for interested parties

How about...?

```suggestion
This CEP focuses on the standardization of existing virtual package implementations.

The following items are not considered here. Though would be open for discussion in future CEP work:
```

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Additional OSes, like `__freebsd` or `__netbsd`.
+- Architectures, like `__x86_64` or `__arm64`, or, more generally, [`__arch`](https://github.com/conda/conda/issues/13420).
```
jakirkham, line 146: Trying to distinguish this a bit from `__archspec` for clarity

```suggestion
- Coarse grain architecture information, like `__x86_64` or `__arm64`, or, more generally, [`__arch`](https://github.com/conda/conda/issues/13420).
```

jakirkham (commented):
Thanks Jaime! 🙏

Like the improvements and additions

Took another review pass with comments below

cep-????.md
h-vetinari: I'd be fine to rule out or defer musl-based systems, unless someone demonstrates that conda packages work by default with a given glibc-on-alpine shim.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. The build string SHOULD reflect the detected CPU microarchitecture when the target platform matches the native platform, 
+as provided by the generic values in the [`archspec/archspec-json` database](https://github.com/archspec/archspec-json/blob/v0.2.5/cpu/microarchitectures.json), plus some exceptions. For example,  `conda/conda` reports its generic values for Intel/AMD and `m*` names for Apple:
+
+- Generic: `x86_64_v1`, `x86_64_v2`, `x86_64_v3`, `x86_64_v4`, `arm`, `ppc`...
+- Apple: For M1, the value is `m1`. M2 and M3 are reported as `m2` and `m3`, respectively.
```
baszalmstra, line 46: Im not sure I understand this part. `x86_64_v1`, `m1` etc are also part of archspec and are also properly detected as such.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+| `*-aarch64` | `aarch64`  |
+| `*-arm64` | `arm64`  |
+| `*-ppc64` | `ppc64`  |
+| `*-ppc64le` | `ppc64le`  |
+| `*-riscv64` | `riscv64`  |
+| `*-s390x` | `s390x`  |
+| `zos-z` | `0` |
+| Any other value | `0` |
+
+The build string MUST be overridable with the `CONDA_OVERRIDE_ARCHSPEC` environment variable, if set to a non-empty value.
```
baszalmstra, line 65: You are using `MUST` here but use `MAY` in the top level section about environment variables.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### `__cuda`
+
+This virtual package MUST be present when the system exhibits GPU drivers compatible with the CUDA runtimes. When available, the version value MUST be set to the oldest CUDA version supported by the detected drivers (i.e. the formatted value of `libcuda.cuDriverGetVersion()`), constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
+
```
baszalmstra, line 70: `cuDriverGetVersion` returns the version of the installed driver, not the oldest CUDA version supported by the detected drivers. Due to backwards compatibility support newer drivers also support older versions.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value.
+
+#### `__glibc`
+
+This virtual package MUST be present when the native platform is `linux-*`. Its version value MUST be set to the system GNU `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_GLIBC` environment variable, if set to a non-empty value.
+
+If the GNU `libc` version could not be estimated (e.g. the tool is not running on Linux), the tool SHOULD provide a default value (e.g. `2.17`) and inform the user of that choice and its possible overrides; e.g. via `CONDA_OVERRIDE_GLIBC`, a CLI flag or a configuration file. The environment variable MUST be ignored when the target platform is not `linux~-*`.
```
baszalmstra, line 79: Should there be a difference between gnu libc not being present and the version not being parsable? I think conda should also be able to run on systems without glibc if packages dont require it.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+This virtual package MUST be present when the target platform is `osx-*`. Its version value MUST be set to the first two numeric components of macOS version formatted as `{major}[.{minor}]`. If the version cannot be estimated (e.g. because the native platform is not macOS), the fallback value MUST be set to `0`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_OSX` environment variable. If this environment variable is set to the empty string `""`, then the `__osx` virtual package MUST NOT be present. The environment variable MUST be ignored when the target platform is not `osx-*`.
+
+> The macOS version can be obtained via:
+> 
+> - Python's `platform.mac_ver()[0]`
+> -  `sw_vers -productVersion`
+>
+> If applicable, the `SYSTEM_VERSION_COMPAT` workaround MUST NOT be enabled; e.g. the version reported for Big Sur should be 11.x and not 10.16.
```
baszalmstra, line 110: I think this should go out of the quoted section. Its pretty important.

cep-????.md
```diff
@@ -0,0 +1,160 @@
...
+
+#### `__unix`
+
+This virtual package MUST be present when the target platform is `linux-*`, `osx-*` or `freebsd-*`. The version and build string fields MUST be set to `0`.
+
+The version or build string fields MUST NOT be overridden by the `CONDA_OVERRIDE_UNIX` environment variable. However, if this environment variable is set to a non-empty value, the `__unix` virtual package MUST be present if otherwise if would not have been. The environment variable MUST be ignored when the target platform is not `linux-*`, `osx-*` or `freebsd-*`.
+
+#### `__win`
+
+This virtual package MUST be present when the target platform is `win-*`. The version MUST be set to the first three numeric components of the Windows build version, formatted as `{major}.{minor}.{build}`. If the version cannot be estimated (e.g. because the target platform does not match the native platform), the fallback value MUST be set to `0`. The build string MUST be `0`.
```
baszalmstra, line 120: This is not currently the case for any installer  (conda, mamba, pixi) right? I think all of them report `__win=0=0` at the moment.

Im all for this change though!

Edit: mamba does seem to report the windows version properly:

```
> pixi exec mamba info
...
virtual packages : __win=10.0.26100=0

> pixi exec conda info
...
virtual packages : __win=0=0

> pixi info
...
Virtual packages: __win=0=0
```

cep-????.md
jaimergp: Maybe. What would be the expected behaviour? Something like this?

- Not found: `__glibc` does not show up.
- Not parsable: `__glibc=0` + warning/advice to override.

jaimergp: I reworded this section a bit now so let's take another look.

jaimergp:
I reworked some sections:

* `__glibc` only makes sense if native and target platforms match the same linux subdir, so now the text specifies the `MUST` clauses more strictly.
* It makes no sense to override `__unix` or forcing its presence in non unix platforms, so this is now specified.
* Removing (only some!) virtual packages by setting overrides to an empty string seems like a bad / useless idea. This will involve some fixes in `conda/conda` for sure, but we'll have a document now.
* General considerations were reworded too for clarity.
* Other suggestions from the code review.

cep-????.md
jaimergp: Exactly! This is what triggered the CEP: we will need to extend what mamba does to the full ecosystem soon due to some Windows deprecations in the boost library. See referenced issues in the text for more details.

cep-????.md
jaimergp: Moved. Thanks!

cep-????.md
jaimergp: I think this returns the CUDA version as seen by the driver. The [docs](https://docs.nvidia.com/cuda/archive/9.2/cuda-driver-api/group__CUDA__VERSION.html#group__CUDA__VERSION_1g8b7a10395392e049006e61bcdc8ebe71) are not super clear. There's also [`cudaRuntimeGetVersion`](https://docs.nvidia.com/cuda/archive/9.2/cuda-runtime-api/group__CUDART____VERSION.html#group__CUDART____VERSION_1g0e3952c7802fd730432180f1f4a6cdc6)At least this is what [`conda/conda` uses](https://github.com/conda/conda/blob/e74a2b9d8a74837afc3bcbef609fe4bd29572e16/conda/plugins/virtual_packages/cuda.py#L142).

@jakirkham could you help clarify this? Thanks!

cep-????.md
jaimergp: Yep, the general section covers all cases (and some virtual packages are not overridable), so we can only use `MAY` there. Then we go case-by-case specifying the `MUST` parts. Happy to reword for clarity if needed, though.

cep-????.md
jaimergp: Moved!

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. 
+
+The build string MUST reflect one of:
+
+- The detected CPU microarchitecture when the target platform matches the native platform, as specified in the [`archspec/archspec-json` database](https://github.com/archspec/archspec-json/blob/v0.2.5/cpu/microarchitectures.json). For Apple Silicon, the reported values MUST correspond to the `vendor == Apple` entries; e.g. `m1`, `m2`, etc. For any other vendors, the reported values MUST correspond to the `vendor == generic` entries; e.g. `x86_64_v1`, `x86_64_v2`, `x86_64_v3`, `x86_64_v4`, `arm`, `ppc`, etc.
```
isuruf, line 60: Why should it correspond to `vendor == generic`? It's not the current behaviour in conda nor is it desired.

jaimergp: It's so far what I had observed in my Windows cloud machine, but maybe that's because `archspec` could not identify the actual CPU? 

Reading the [used function](https://github.com/archspec/archspec/blob/2084d981cbda2fcf7dfd5f51d4eb56ccc507bce1/archspec/cpu/detect.py#L338), that [might be indeed the case](https://github.com/archspec/archspec/blob/2084d981cbda2fcf7dfd5f51d4eb56ccc507bce1/archspec/cpu/detect.py#L364-L365).

jaimergp: Can we just say "whatever `archspec.cpu.detect.host()` does"?

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+This virtual package MUST be present when the system exhibits GPU drivers compatible with the CUDA runtimes. When available, the version value MUST be set to the oldest CUDA version supported by the detected drivers (i.e. the formatted value of `libcuda.cuDriverGetVersion()`), constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value that can be parsed as a version string.
+
+#### `__glibc`
+
+This virtual package MUST NOT be present if the target platform is not `linux-*`.
+
+This virtual package MUST be present when the native and target platforms are both the same type of `linux-*` and GNU `libc` is installed in the system. The version value MUST be set to the system GNU `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. If the version cannot be estimated, the tool MUST set the version to a default value (e.g. `2.17`).
```
jjhelmus, line 91: Is it worth calling out an expected behavior on Linux systems where the `libc` is not GNU, for example those using [musl](https://musl.libc.org/).

jakirkham: There has been a lot of discussion on this point already above: https://github.com/conda/ceps/pull/103#discussion_r1889312955

Essentially we have decided so far to...

* Focus on existing virtual packages like `__glibc` (not ones yet to be added)
* Standard system configurations (so no `glibc` compat layer)

There is a "Potential future work" section below that states `__musl` could be pursued in a future CEP

jjhelmus: This is covered by the callout below:
> More `libc` implementations, like `__musl`.

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### `__osx`
+
+This virtual package MUST be present when the target platform is `osx-*`. Its version value MUST be set to the first two numeric components of macOS version formatted as `{major}.{minor}`. If applicable, the `SYSTEM_VERSION_COMPAT` environment variable workaround MUST NOT be enabled; e.g. the version reported for Big Sur should be 11.x and not 10.16. If the version cannot be estimated (e.g. because the native platform is not macOS), the fallback value MUST be set to `0`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_OSX` environment variable if set to a non-empty value that can be parsed as a version string. The environment variable MUST be ignored when the target platform is not `osx-*`.
+
+> The macOS version can be obtained via:
+> 
+> - Python's `platform.mac_ver()[0]`
```
jjhelmus, line 126: This will return the SYSTEM_VERSION_COMPAT version if the Python interpreter running the command was built against the 10.15 SDK or earlier.

See https://eclecticlight.co/2020/08/13/macos-version-numbering-isnt-so-simple/ for details.

jaimergp: AFAICR, if you start the Python interpreter with SYSTEM_VERSION_COMPAT=0 it returns the 11.x based version, right?

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+- `__archspec`
+- `__cuda`
+- `__glibc`
+- `__linux`
+- `__osx`
+- `__unix`
+- `__win`
+
+#### `__archspec`
```
chenghlee, line 54: Are we aware of any existing packages that actually depend on `__archspec`? If not, I would rather we not make this virtual package mandatory and pull out `__archspec` into its own CEP (or set of CEPs). Mostly because "what microarchitecture is this CPU?" is a question that can get complicated quickly; see, e.g., ARM big.LITTLE, Intel P/E cores, Intel Xeon 6, etc.

cep-????.md
jjhelmus: Yes, running setting the environment variable will return a >=11 version. 

My main concern here the "MUST NOT" language around the SYSTEM_VERSION_COMPAT workaround. I agree with the idea but this is not the case for the current version of conda (see conda/conda#13832). The example an this issue still reports a 10.16 version for `__osx`. Changing this to "SHOULD" would be reasonable, especially given that there are many releases of tools/packages that will report the compatible version that already exist.

cep-????.md
jjhelmus: The [microarch-level](https://github.com/conda-forge/microarch-level-feedstock) packages in conda-forge depend on `__archspec` to provide microarchitecture-level (e.g. x86-64-v2) meta-packages.

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+The version MUST be overridable with the `CONDA_OVERRIDE_OSX` environment variable if set to a non-empty value that can be parsed as a version string. The environment variable MUST be ignored when the target platform is not `osx-*`.
+
+> The macOS version can be obtained via:
+> 
+> - Python's `platform.mac_ver()[0]`
+> -  `SYSTEM_VERSION_COMPAT=0 sw_vers -productVersion`
+
+#### `__unix`
+
+This virtual package MUST be present when the target platform is `linux-*`, `osx-*` or `freebsd-*`. The version and build string fields MUST be set to `0`.
```
chenghlee, line 131: Rather than list the specific target platforms now and having to adjust this CEP if/when new platforms are adopted, I would suggest language like "when the target platform is sufficiently POSIX-y" and list the attributes necessary for that to be true (e.g., uses `/` for path delimiters, supports `fork(3)`, etc.)

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+> The GNU `libc` version can be computed via:
+>
+> - Python's `os.confstr("CS_GNU_LIBC_VERSION")`
+> - `getconf GNU_LIBC_VERSION`
+> - `ldd --version`. Please verify that it references GNU `libc` or GLIBC. For non-standard installs, using a GLIBC compatibility layer, this may require locating the implementation and directly querying.
+
+#### `__linux`
+
+This virtual package MUST be present when the target platform is `linux-*`. Its version value MUST be set to the Linux kernel version, constrained to the first two to four numeric components formatted as `{major}.{minor}.{micro}.{patch}`. If the version cannot be estimated (e.g. because the native platform is not Linux), the tool MUST set `version` to a fallback value of its choice. The build string MUST be `0`.
```
chenghlee, line 108: The Linux kernel upstream only defines `{major}.{minor}.{micro}`; anything beyond that (including the `{patch}` component you wrote) is part of the [distribution kernel's](https://www.kernel.org/releases.html#distribution-kernels) version string. I don't think conda tooling should expose those components since their semantics of those will vary from distribution to distributon; e.g., patch 42 on Fedora may differ from patch 42 on Ubuntu.

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value that can be parsed as a version string.
+
+#### `__glibc`
+
+This virtual package MUST NOT be present if the target platform is not `linux-*`.
+
+This virtual package MUST be present when the native and target platforms are both the same type of `linux-*` and GNU `libc` is installed in the system. The version value MUST be set to the system GNU `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. If the version cannot be estimated, the tool MUST set the version to a default value (e.g. `2.17`). 
+
+If the native platform does not match the target platform, the tool MAY export `__glibc` with its `version` field set to a default value (e.g. `2.17`) of its choice.
```
chenghlee, line 93: Should we provide some guidance here on how the default value should be selected? If we don't, I could see situations arising where two different conda install tools or two different versions of the same install tool provide different default values on the same system.

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### `__cuda`
+
+This virtual package MUST be present when the system exhibits GPU drivers compatible with the CUDA runtimes. When available, the version value MUST be set to the oldest CUDA version supported by the detected drivers (i.e. the formatted value of `libcuda.cuDriverGetVersion()`), constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
+
+The version MUST be overridable with the `CONDA_OVERRIDE_CUDA` environment variable, if set to a non-empty value that can be parsed as a version string.
+
+#### `__glibc`
+
+This virtual package MUST NOT be present if the target platform is not `linux-*`.
```
chenghlee, line 89: "...if the target platform's C standard library is not [the GNU C Library](https://www.gnu.org/software/libc/)". Not sure if we codify the assumption of all Linux systems inherently provide GNU `libc` or that only Linux uses GNU `libc`.

cep-????.md
chenghlee, line 54: Are we aware of any existing packages that actually depend on `__archspec`?

If not, I would rather we _not_ make this virtual package mandatory at this time, mostly because "what microarchitecture is this CPU?" is a question that can get complicated quickly; see, e.g., ARM big.LITTLE, Intel P/E cores, Intel Xeon 6, etc. To me, this virtual package is still (pseudo-)experimental, in the sense that we still need to work out how package maintainers should/want to use this package. IMO, `__archspec` should be its own CEP or set of CEPs (cf. conda/ceps#59).

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+* The lowest CUDA version the GPU driver is compatible with via `__cuda`.
+
+## Specification
+
+A virtual package is defined as a package record with three fields: name, version and build string.
+The name MUST start with double underscore (`__`). The version and build string MUST follow the same semantics as in regular package records. More specifically, the version field MUST follow the version string specifications, regardless its origin (computed from a system property, overridden by the user or configuration, or provided by default by the tool).
+
+Some general considerations: 
+
+- The version or build string of a virtual package MAY be overridden by the value of `CONDA_OVERRIDE_{NAME}` environment variable, with `{NAME}` being the uppercased name of the virtual package. Many exceptions apply so please observe the details in the section below.
```
chenghlee, line 38: Just to be clear: "...being the uppercased name of the virtual package, excluding the leading double underscore (`__`)."

cep-????.md
jaimergp: I think we could consider that a bug in conda and submit a fix for 25.1 (as we are doing for __win). I can survey public repodata for __osx usage in the wild if that helps inform this decision.

cep-????.md
jaimergp: * [`conda/conda` sets it to `2.5`](https://github.com/conda/conda/blob/e74a2b9d8a74837afc3bcbef609fe4bd29572e16/conda/common/_os/linux.py#L32) (yes, I know).
* [Pixi chose 2.28](https://github.com/prefix-dev/pixi/blob/bb2b584c0c1c7085401638d5be80fb9ebf73f9a8/crates/pixi_default_versions/src/lib.rs#L3-L6).
* [Mamba skips it](https://github.com/mamba-org/mamba/blob/6e6515ecfb43d76fbdd772ae5e1b0974d96a8daf/libmamba/src/core/virtual_packages.cpp#L314).

cep-????.md
jaimergp: See [conda/conda@`d94af02` (#14449)](https://github.com/conda/conda/pull/14449/commits/d94af02ba53f14ae3a6e1598c6de5751219c34e8)

cep-????.md
jaimergp: Hm, this contradicts [this comment in `conda/conda`](https://github.com/conda/conda/blob/e74a2b9d8a74837afc3bcbef609fe4bd29572e16/conda/plugins/virtual_packages/linux.py#L20-L29). I'm happy to trim to three fields, but we'll need a source.

cep-????.md
jaimergp: I agree with the sentiment, but I'm going to need some help or references to compile that list 😬

cep-????.md
jaimergp: Current conda implementation assumes the latter "only Linux uses GNU libc". See https://github.com/conda/conda/blob/e74a2b9d8a74837afc3bcbef609fe4bd29572e16/conda/plugins/virtual_packages/linux.py#L14-L16.

If that's incorrect, (1) we need to fix that in `conda/conda`, and (2) I will update this paragraph.

cep-????.md
jaimergp: https://github.com/conda/ceps/pull/103#discussion_r1894102373

cep-????.md
chenghlee: Ah ha! Since Linux 3.0, "[[mainline] kernels with 3-part versions are -stable kernels](https://www.kernel.org/doc/html/latest/process/howto.html#various-stable-trees-with-multiple-major-numbers)"; however, the [mainline 2.6 series](https://www.kernel.org/pub/linux/kernel/v2.6/) had [4-part version numbers](https://en.wikipedia.org/wiki/Linux_kernel_version_history#Releases_2.6.x.y). The existing "3- or 4-part" comment in `conda/conda` comes from the fact that 2.6 is the upstream for the CentOS/RHEL 6 kernel, which was actively supported by Anaconda and conda-forge at the time I wrote that patch.

In light of that, I suggest replacing:

> Its version value MUST be set to the Linux kernel version, constrained to the first two to four numeric components formatted as `{major}.{minor}.{micro}.{patch}`.

with something like:

> Its version value MUST be set to the upstream (AKA mainline) Linux kernel version, but it MUST _exclude_ any and all [distribution-specific components](https://www.kernel.org/releases.html#distribution-kernels) of the kernel version.

I think that captures the intent of my previous comment, which was to ensure:
1. we don't allow conda packages to depend on [Linux] distribution-specific kernels via the `__linux` virtual package; and
2. we don't have to deal with the various ways Linux distros version their kernels (which may or may not be compatible with how conda does version ordering)

(That said, my suggested language does allow for non-release/development/`-rcN` kernels, but I suspect that user base is relatively small and able to cope if something goes horribly wrong.)

cep-????.md
jakirkham: Asking offline to see if others have thoughts here

That said, here are some rough thoughts. Starting with some background

When the CUDA Toolkit is typically shipped (like in the installer), it ships with a full suite of things including the driver library. However in Conda, we ship pretty much everything else except the driver library. So with Conda, the user is on the hook for ensuring the driver library is installed, which would happen when they install the driver

So this raises a question. How does Conda decide what version of the driver is compatible with a particular CUDA Toolkit?

This is what `__cuda` is solving. It is giving us the driver library's associated CUDA Toolkit version. IOW if that driver library was shipped in an installer, this is that installer's CUDA Toolkit version

So now when we try to install the CUDA Toolkit libraries, we can answer the question, will the driver be able to handle these?

Originally the answer was the CUDA Toolkit libraries we install could be no newer than the CUDA Toolkit the driver came from. We had [checked this down to the major minor version]( https://github.com/conda-forge/cudatoolkit-feedstock/blob/0f01f6edfe22e77ac068daa46505711a1481272e/recipe/meta.yaml#L701-L704 ) (patch version was allowed to float). So if this driver check reported CUDA 10.1, we could only use CUDA Toolkit libraries from 10.1 or older. Eventually we relaxed this to major only for [11.2+]( https://github.com/conda-forge/cudatoolkit-feedstock/blob/0f01f6edfe22e77ac068daa46505711a1481272e/recipe/meta.yaml#L705 ) and [12]( https://github.com/conda-forge/cuda-version-feedstock/blob/dff44fb6818a6525498ca3a72cb78d0f043782fc/recipe/meta.yaml#L16 )

So really `__cuda`'s version is the upper bound on the libraries we can use

Sorry for the long winded answer. Though hopefully the additional details provide context. Perhaps this can be incorporated into the text above somehow (after some discussion and Q&A)

jakirkham: Maybe something like this captures the idea above. Though please feel free to adjust further. Happy to look again after

```suggestion
This virtual package MUST be present when the system exhibits NVIDIA GPU drivers compatible with the CUDA runtimes.
For systems without such support the virtual package MUST NOT be present.
When available, the version value MUST be set to the latest CUDA version supported by the detected drivers (i.e.
the formatted value of [`cuDriverGetVersion()`]( https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VERSION.html )),
constrained to the first two components (major and minor) and formatted as `{major}.{minor}`. The build string MUST be `0`.
```

jaimergp: Thanks John. With 12+, how is the upper bound determined? It's not about major.minor anymore, right? e.g. how do I know which `y` to use in `<=12.y`? Or are we expected to use `<13`?.

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+  | `*-armv6l`      | `armv6l`                         |
+  | `*-armv7l`      | `armv7l`                         |
+  | `*-aarch64`     | `aarch64`                        |
+  | `*-arm64`       | `arm64`                          |
+  | `*-ppc64`       | `ppc64`                          |
+  | `*-ppc64le`     | `ppc64le`                        |
+  | `*-riscv64`     | `riscv64`                        |
+  | `*-s390x`       | `s390x`                          |
+  | `zos-z`         | `0`                              |
+  | Any other value | `0`                              |
```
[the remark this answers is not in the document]
jaimergp: What's the platform/archspec tuple for `wasm32`?

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`. 
+
+The build string MUST reflect one of:
+
+- If the target platform matches the native platform, the best fitting CPU microarchitecture in the [`archspec/archspec-json` database](https://github.com/archspec/archspec-json/blob/v0.2.5/cpu/microarchitectures.json). The reference CPU detection implementation is [`archspec.cpu.detect.host()`](https://github.com/archspec/archspec/blob/v0.2.5/archspec/cpu/detect.py#L338).
+
+- The target platform architecture (second component of the platform string), mapped as:
```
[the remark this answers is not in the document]
jaimergp: I'm happy to extend the table as needed, so far it's a description of the observed behaviours in `conda`, mostly. aka "2nd part of the subdir, except for -32 and -64"

cep-????.md
```diff
@@ -0,0 +1,171 @@
...
+- `__cuda`
+- `__glibc`
+- `__linux`
+- `__osx`
+- `__unix`
+- `__win`
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`.
```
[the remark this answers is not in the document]
jaimergp: Because `conda` always reports it and I was being descriptive. Happy to change this but we need to agree on why one behavior is better to guarantee changes in existing clients.

cep-????.md
jaimergp: Added [6a9f82e](https://github.com/conda/ceps/pull/103/commits/6a9f82e19411a66272ec457981d26387c58538b5)

cep-????.md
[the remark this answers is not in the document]
baszalmstra: we use

* emscripten-wasm32
* wasi-wasm32

https://github.com/conda/rattler/blob/2c4946907dd166197bd2e26cf1f0378a9008e8b6/crates/rattler_conda_types/src/platform.rs#L35

cep-XXXX.md
```diff
@@ -0,0 +1,184 @@
...
+## Specification
+
+A virtual package is defined as a package record with three fields: name, version and build string.
+These fields MUST comply with [CEP 26](./cep-0026.md). More specifically, the version field MUST follow the version string specifications as discussed in [CEP PR #132](https://github.com/conda/ceps/pull/132), regardless its origin (computed from a system property, overridden by the user or configuration, or provided by default by the tool).
+
+Some general considerations:
+
+* The version or build string of a virtual package MAY be overridden by the value of `CONDA_OVERRIDE_{NAME}` environment variable, with `{NAME}` being the uppercased name of the virtual package (excluding the leading underscores). Many exceptions apply so please observe the details in the section below.
+* The build string MAY be zero (`0`). Some exceptions apply. See below.
+* When the tool used a fallback default value instead of a computed one, it SHOULD also inform the user of that choice and its possible override options (e.g. `CONDA_OVERRIDE_{NAME}` variables, CLI flags, configuration file, etc).
```
baszalmstra: > When the tool used a fallback default value instead of a computed on

Do you have an example of this? Do you mean that the tool should warn if it failed to determine a version/buildstring for a virtual package?

cep-XXXX.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### `__archspec`
+
+This virtual package MUST be always present, with the version set to `1`.
+
+The build string MUST reflect one of:
+
+* If the target platform matches the native platform, the best fitting CPU microarchitecture in the [`archspec/archspec-json` database](https://github.com/archspec/archspec-json/blob/v0.2.5/cpu/microarchitectures.json). The reference CPU detection implementation is [`archspec.cpu.detect.host()`](https://github.com/archspec/archspec/blob/v0.2.5/archspec/cpu/detect.py#L338).
+
+* Otherwise, the target platform architecture (second component of the platform string), mapped as:
```
baszalmstra: I understand this is the current state of afairs but this is weird to me because many of the build strings in the table below are not present in the archspec database.

cep-XXXX.md
```diff
@@ -0,0 +1,184 @@
...
+
+If the `CONDA_OVERRIDE_GLIBC` environment variable if set to a non-empty value that complies to the version string specification and the target platform is `linux-*`, the tool MUST export `__glibc` with its version value set to the value of the environment variable.
+
+The build string MUST always be `0`.
+
+> The GNU `libc` version can be computed via:
+>
+> * Python's `os.confstr("CS_GNU_LIBC_VERSION")`
+> * `getconf GNU_LIBC_VERSION`
+> * `ldd --version`. Please verify that it references GNU `libc` or GLIBC. For non-standard installs, using a GLIBC compatibility layer, this may require locating the implementation and directly querying.
```
baszalmstra: Rattler also tries to dynamically load the `gnu_get_libc_version` symbol from `libc.so.6`.

jaimergp:
```suggestion
> * `ldd --version`. Please verify that it references GNU `libc` or GLIBC. For non-standard installs, using a GLIBC compatibility layer, this may require locating the implementation and directly querying.
> * Dynamically load the `gnu_get_libc_version` symbol from `libc.so.6`. 
```

cep-XXXX.md
```diff
@@ -0,0 +1,192 @@
...
+
+The build string MUST be `0`.
+
+If the `CONDA_OVERRIDE_CUDA` environment variable is set to a non-empty value that can be parsed as a version string, the `__cuda` virtual package MUST be exposed with that version.
+
+#### `__glibc`
+
+This virtual package MUST NOT be present if the target platform is not `linux-*`.
+
+This virtual package MUST be present when the native and target platforms are both the same type of `linux-*` and GNU `libc` is installed in the system. The version value MUST be set to the system GNU `libc` version, constrained to the first two components (major and minor) formatted as `{major}.{minor}`. If the version cannot be estimated, the tool MUST set the version to a default value (e.g. `2.17`).
```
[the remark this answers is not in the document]
jaimergp: Yep, "a default value" not "this default value". I will elaborate a bit more so it's not as ambiguous.

PR #110, opened 2025-02-04
CEP-0014: permit lists in `context`

mgorny:
### Description

Update the section describing the `context` block to permit lists of scalars in addition to scalars.  Add a note that the resulting Jinja variable will have the same type as the value.  For example, this can be used to declare lists of skipped tests, conveniently interspersing them with comments, and afterwards combining into a single string.

The change is backwards compatible and minimal, so it seems cleanest to update the specification in place rather than via another CEP.

See also: https://github.com/prefix-dev/rattler-build/pull/1402

CC @wolfv

### DIFF
--- a/cep-0014.md
+++ b/cep-0014.md
@@ -5,7 +5,7 @@
 <tr><td> Status </td><td> Accepted</td></tr>
 <tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
 <tr><td> Created </td><td> May 23, 2023</td></tr>
-<tr><td> Updated </td><td> Jan 22, 2024</td></tr>
+<tr><td> Updated </td><td> Feb 04, 2025</td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda-incubator/ceps/pull/56 </td></tr>
 <tr><td> Implementation </td><td>https://github.com/prefix-dev/rattler-build</td></tr>
 </table>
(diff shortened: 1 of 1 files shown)


cep-0014.md
```diff
@@ -68,7 +68,8 @@ To benefit from autocompletion, and other LSP features in editors, we can add a
 ## Context section
 
 The context section is a dictionary with key-value pairs that can be used for string interpolation.
-The right-hand side of the key-value pair is a scalar (bool, number or string).
+The right-hand side of the key-value pair is either a scalar (bool, number or string), or a list of scalars.
```
wolfv, line 71: Should we make sure that lists can have only a single type? E.g. `List[string]`, `List[bool]`, or `List[integer]`.

mgorny: Good question. On one hand, FWICS minijinja is fine with mixed-type sequences. On the other hand, I don't see any good reason to use such a thing — and it probably increases the risk of mistakes, such as missing quoting around `no`.

mgorny: Added a sentence to that effect, and will look into updating rattler-build patch soon.

isuruf:
Does this mean we are allowed to do something like

```yaml
context:
  build_reqs:
    - ${{ compiler('cxx') }}
    - make

recipe:
  name: test
  version: 0.1.0

outputs:
  - package:
      name: foo
    requirements:
      build: ${{ build_reqs }}

  - package:
      name: bar
    requirements:
      build: ${{ build_reqs }}
```

wolfv:
I think this syntax would not work, but a list of list is flattened and that would then work, yes.

mgorny:
> Although I am unsure whether modifying existing CEPs in this way is "allowed". Let's ask on the conda CEP Zulip channel ...

I see the only reply was basically "yes", so should we (you?) start the vote?

wolfv:
@mgorny we decided we can just "accept" this. Maybe you can take a look at my latest change to cep-0014 and add a little changelog thing as well? Then we can merge! Thank you & sorry for the long wait :)

Issue #77, opened 2024-05-23
Adding `host_exports` (like `run_exports` except between `build` & `host` only)


jakirkham:
`run_exports` has been immensely helpful. Just to summarize it `run_exports`:

* Allows packages with knowledge of what they need to express it for easy downstream consumption
  * Packages that are part of the build can add dependency to `run`
    * `weak` only happens when the dependency is added to `host` (not `build`)
    * `strong` happens when a dependency is added to `build` or `host` 
  * Dependencies added can express ABI compatibility

However there are still some cases that are not covered. In particular there are cases where a dependency added to `build` needs to add itself to `host`, but there is no need to add it to `run` (in fact this would often be better to avoid). IOW `run_exports/strong` does not meet this need. Some examples include:

* Header-only dependencies of a package
  * https://github.com/conda/conda-build/issues/3796#issuecomment-549431589
* Tighter alignment in `build` & `host` of compiler dependencies (like OpenMP)
  * https://github.com/conda-forge/openmp-feedstock/issues/126
* Compiler runtime libraries that are statically linked (like `cudart`)
  * https://github.com/conda-forge/staged-recipes/pull/21723#issuecomment-1405444862

Solving these cases would involve a `host_exports` that allows adding other dependencies to `host`. Could imagine to use cases for a package providing `host_exports`

* `host_exports/weak` (default)
  * Add dependencies to `host` if providing package is added to `host`
  * Solves the header-only dependency case above for a library
* `host_exports/strong` (opt-in)
  * Add dependencies to `host` if providing package is in `build` or `host`
  * Handles the compiler dependency alignment problem (with OpenMP)
  * Handles statically linked compiler libraries (like `cudart`)
  * Of most use for compiler use cases

jakirkham:
cc @xhochy (for awareness)

jakirkham:
@jaimergp @wolfv would be interested to hear your perspectives on this idea? 🙂

jaimergp:
I'm generally supportive of the idea provided that it's useful for a wide range of cases. A quick look made me see it as a "couple of edge cases". For example, it's not clear to me what "Add dependencies to host if providing package is added to host" means; if the package is already in host, why do we need to export anything? Is it to constrain it further? I guess we can use  `pin_compatible` but we want to avoid that "annoyance" if the package in build already knows how tell host, right?

jakirkham:
Thanks Jaime! 🙏

Am happy with limiting it to the case where a package is added to `requirements/build`

[...]

PR #111, opened 2025-02-05
CEP XXXX: Conditional dependencies, extras and flags

wolfv:
## Checklist for submitter

- [X] I am submitting tree new CEPs:
  - [Conditional dependencies](https://github.com/wolfv/ceps/blob/repodata-next/cep-XXX1.md)
  - [Optional dependency groups](https://github.com/wolfv/ceps/blob/repodata-next/cep-XXX2.md)
  - [Flags](https://github.com/wolfv/ceps/blob/repodata-next/cep-XXX3.md)

This defines new repodata with two additional fields and a modfied matchspec syntax:

- extra (optional) dependency sets that can be activated by the user or other packages
- conditional dependencies to be activated on a specific platform, or when other package versions are also pulled in (e.g. require `six; if python <3.8`)
- flags to make it easier to select variants confidently (instead of globbing the build string)

Some initial discussion happened over the past days on:

- Zulip:
  - ["Flags in conda packages", conda-forge](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.22flags.22.20in.20conda.20packages)
  - ["Flags", conda](https://conda.zulipchat.com/#narrow/channel/457620-ceps/topic/Flags)
  - ["Conditional dependencies syntax in v2 environments & recipes", conda](https://conda.zulipchat.com/#narrow/channel/464880-conda-maintainers/topic/Conditional.20dependencies.20syntax.20in.20v2.20environments.20.26.20recipes/with/509031850)
- [PR](https://github.com/conda/rattler/pull/1040)
- [HackMD](https://hackmd.io/z4MMDhlkS6WTzyTyaRkjNQ). 

TODO:

- [ ] specify that we need to understand full-blown matchspecs in the repodata.v2 (e.g. with bracket syntax), because we want other packages to be able to request extras or flags).
- [ ] Decide whether flag negation should be `~foo`, `!foo`, or `-foo`.
- [ ] Decide if we need to add a default flags entry to the repodata record, or another way to mark the default variant (can also still use down-priorization via track records)
- [ ] Investigate ordering of flags and whether we should have some weighing e.g. for numerical flags (higher == better)? Or wether this should be decided by another order integer that the users can compute.


cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+# with their own set of flags
+flags: ["gpu:cuda", "blas:mkl", "archspec:4", "release"]
+```
+
+Flags can then be matched using the following little syntax:
+
+- `release`: only packages with `release` flag are used
+- `~release`: disallow packages with `release` flag
+- `?release`: if release flag available, filter on it, otherwise use any other
+- `gpu:*`: any flag starting with `gpu:` will be matched
```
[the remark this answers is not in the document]
wolfv: Would an exact match not work fine?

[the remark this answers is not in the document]
wolfv: Would an exact match not work fine?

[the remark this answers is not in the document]
mariusvniekerk: It would, we should just state how to do an exact match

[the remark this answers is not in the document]
wolfv: You coudl just ask for `flags = ["blas:mkl"]` for an exact match :)

cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+version: 1.0.0
+depends:
+ - python >=3.8
+ - pywin32; if __win
+ - six; if python <3.8
+```
+
+The proposed syntax is to extend the `MatchSpec` syntax by appending `; if <CONDITION>` after the current MatchSpec.
+
+We would like to also allow for AND and OR with the following syntax:
```
jaimergp, line 73: You probably need `NOT` and parentheses for precedence overrides, right?

cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+```yaml
+name: sqlalchemy
+version: 1.0.0
+depends:
+ - python >=3.8
+ - pywin32; if __win
+ - six; if python <3.8
+```
+
+The proposed syntax is to extend the `MatchSpec` syntax by appending `; if <CONDITION>` after the current MatchSpec.
```
jaimergp, line 71: It would be nice if we didn't need the `;` because somehow `conda` will [happily ignore the `if` parts](https://github.com/conda/conda/blob/ffecefa61867489f764880366fb5ccef2143675c/conda/models/match_spec.py#L619-L623) while parsing `MatchSpec`s.

```python
>>> from conda.models.match_spec import MatchSpec as M
>>> M("python 3.8 * if python")
MatchSpec("python==3.8[build=*]")
>>> M("python 3.8 'if' if  __win")  # quote 'if' to force parse it as a build string
MatchSpec("python==3.8='if'")
```

jaimergp: It will also ignore parenthesised blocks:

```python
>>> M("python 3.8 * (__win)")
MatchSpec("python==3.8[build=*]")
>>> M("python 3.8 (__win)")
MatchSpec("python==3.8")
>>> M("python 3.8 (__win and __osx)")
MatchSpec("python==3.8")
>>> M("python 3.8 (if __win and __osx)")
MatchSpec("python==3.8")
```

jaimergp: `libmamba` also ignores parentheses:

```python
>>> from libmambapy.specs import MatchSpec as LibmambaMatchSpec
>>>print(LibmambaMatchSpec.parse("python 3.8 * (__win and __osx)"))
python==3.8
>>> print(MatchSpec.parse("python 3.8 * (if __win and __osx)"))
python==3.8
```

jaimergp: My suggestion would be to design the syntax like this:

```shell
name [version [build]] ('if' condition)
```

The `if` literal could be omitted if folks feel it's clearer that way.

cep-repodata-next.md
AntoinePrv: In regular MatchSepc, we have `,` and `|` used for versions for `and` and `or`. I think we should keep thing similar, even if it means supporting `and` and `or` in versions.

cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+  postgres:
+    - postgres >=3.5
+    - pyxpgres >=8
+```
+
+When a user, or a dependency, selects an extra through a MatchSpec, the extra and it's dependencies are "activated". This is conceptually the same as having three packages with "exact" dependencies from the "extra" to the base package: `sqlalchemy`, `sqlalchemy-sqlite` and `sqlalchemy-postgres` – which is the workaround currently employed by a number of packages on conda-forge.
+
+## Conditional dependencies
+
+Conditional dependencies are activated when the condition is true. The most straight-forward conditions are `__unix`, `__win` and other platform specifiers. However, we would also like to support matchspecs in conditions such as `python >=3`.
```
AntoinePrv, line 58: I very much like this idea. An implementation note: while `if __unix` is reasonably easy to implement because it is "static", `if python <3.8` is conceptually much harder as it is not something that can be decided ahead of solving. I requires to be able to adapt the dependencies of a package as partial candidates are investigated during solve.

cep-repodata-next.md
AntoinePrv: Another idea (not sure if a good one) could be: rather than make MatchSpecs more complex than they already are, build on the idea of selectors from the new recipe format and allow `depends` to contain objects like so:
```yaml
depends:
 - python >=3.8
 - if: python <3.8
    then: six
```
In the recipe format, this is expressed via a variation on the selector to avoid being process at build time (`if(run):` for instance).

This disallow basically disallow conditionals outside of recipes (or format with selector), but makes a more consistent narrative around conditionals.

cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+
+It's very natural to build different variants for a given package in the conda ecosystem with different properties: blas implementation, gpu / cuda version, and other variables make up the variant matrix for certain recipes.
+
+However, it is not easy to specify which variant a user really wants in conda today. Most of the time, some string-matching on the build string is used to select one of the options, such as `pytorch 2.5.* *cuda`.
+
+There are other workarounds by using `mutex` packages and constraining them such as `blas_impl * mkl` which could be used to select only packages that also depend on the MKL build.
+
+However, it would be nice if we could have a flexible, powerful and simple syntax to enable or disable "flags" on packages in order to select a variant.
+
+A RepodataRecord should get a new field "flags" that is a list of strings, such as:
```
AntoinePrv, line 92: How do `flags` and `extra` mix and overlap? Wouldn't conditional dependencies on a flag be enough to generate the `extra` category?

cep-repodata-next.md
baszalmstra: The idea was indeed also posted on zulip. One issue is that it then becomes harder for cli tools to adopt.

cep-repodata-next.md
baszalmstra: If find this hard to distinguish when parsing the version. E.g. 

```
python <3.8|>3.9 | numpy >=2.0
python <3.8|>3.9 or numpy >=2.0
```

I find the version with `or` easier to read than the one with pipes.

cep-repodata-next.md
baszalmstra: I propose we use the `when` syntax without a semicolon to force an error on older versions of conda that dont support it and the distinguish between the already established `if` syntax in recipe v1. We can use the same keyword in a more expanded form as the "build spec"

```
foobar when python >=3.8
```

E.g. in a recipe:

```yaml
- if: unix
  then: foobar
  when: python >=3.8
  
# OR
  
- when: python >=3.8
  then: foobar
 
# OR 
  
- "foobar when python >=3.8"
```

cep-repodata-next.md
baszalmstra: Flags are part of a variant. So there is no variation of flags for a single variant. E.g. flags could be used to say a particular variant is using cuda and another is used to target cpu. Extras are a way to select additional dependencies for a particular variant. If a variant also adds a CLI tool it provides the extra "cli". Only if that extra is requested by another package are particular dependencies also requested.

In technical terms extras can indeed be implemented as conditional dependencies. E.g. for a package `my_package` we could express it as `typer when my_package[cli]`. If there is a package that depends on `my_package[cli]` `typer` would also be required.

cep-repodata-next.md
baszalmstra: This is the PR that implements this for resolvo: https://github.com/prefix-dev/resolvo/pull/136

cep-repodata-next.md
```diff
@@ -0,0 +1,119 @@
...
+extras:
+  sqlite:
+    - sqlite >=1.5
+    - py-sqlite-adapter 1.0
+  postgres:
+    - postgres >=3.5
+    - pyxpgres >=8
+```
+
+When a user, or a dependency, selects an extra through a MatchSpec, the extra and it's dependencies are "activated". This is conceptually the same as having three packages with "exact" dependencies from the "extra" to the base package: `sqlalchemy`, `sqlalchemy-sqlite` and `sqlalchemy-postgres` – which is the workaround currently employed by a number of packages on conda-forge.
```
baszalmstra, line 54: I think we should clarify what happens if an extra is requested for a package but the selected variant doesnt provide that extra. E.g. what happens if I depend on a `foobar[extras=["doesntexist"]]`

baszalmstra: Maybe I missed it but its also not defined how to depend on an extra?

wolfv: Yes, we discussed this today some more and there are a few options:

- An extra implicitly always exists (as empty) and so we would always choose the latest "base" package (whether it has the extra or not). We would then warn if it does not have the `extra`. I believe this is the behavior that `pip` implements, and makes it easier to be forward compatible even if other dependencies might ask for an extra that does not exist anymore.
- We could strictly require that the extra is there and otherwise fail (or select an older version of the package that has the extra).
- We could forward propagate extras (e.g. the first time we see an extra, we implicitly add it to all later versions, even if empty) which would prevent the solver from choosing a version before the extra existed, but might choose a later one after the extra was removed).

I think we should decide on a behavior, most likely choosing the one from pip. We should also read up on the idea of adding a "default" extra (default dependencies that can be deselected with `no-default-extra` or something along the lines which I recently saw being proposed somewhere in PyPI world as part of wheel-next I think.

wolfv:
This is what a recipe looks like - we have a branch where the tests are passing.

```
package:
  name: conditional
  version: 0.1.0

build:
  flags:
    - cuda
    - openmp:mpi

requirements:
  run:
    - rattler-build
    - bzip2 *; if rattler-build >=0.44
    - zstd *; if rattler-build <0.44

  extras:
    science:
      - numpy
      - matplotlib
    web:
      - flask
    cli:
      - click

tests:
  - script:
    - echo "Hello, world!"
    requirements:
      run:
        - rattler-build >=0.44

  - script:
    - echo "This is a test for Rattler < 0.44"
    requirements:
      run:
        - rattler-build <0.44

  - script:
    - echo "This is a test for Rattler < 0.44"
    requirements:
      run:
        - conditional [extras=[science, web]]
```

wolfv:
Another comment to remind ourselves, to simplify the initial implementation:

- we will not allow conditions in run_constraints (for now)
- We will not allow extra groups to contain anything other than run dependencies (no run constraints or run exports)

baszalmstra:
We just merged

- https://github.com/prefix-dev/resolvo/pull/136

which adds support for all features mentioned in this CEP. We are working on an preliminary version of this CEP in rattler based on this.

baszalmstra:
> Let's split this CEP up between the various proposals here, each have sufficiently complex to implement constraints.

The reason we combined these is they are require changes to the repodata format. We aimed to bundle these together in a single change. 

But I do understand the desire to split the different features. Do you have a suggestion how we can avoid having to bump the repodata version in each CEP?

wolfv:
I think I would be pro-splitting if we would see lots of engagement on the topics presented here. But so far it is not an unwieldy discussion so I don't exactly see the point of splitting :)

h-vetinari:
> "This CEP is co-submitted with X, Y, Z. All of them propose to bump to repodata version v2, the implementation of which MUST happen simultaneously when all CEPs have been reviewed and accepted or rejected".

One of that CEPs could be about this proposed [design](https://github.com/conda/ceps/issues/77#issuecomment-3186740038) for a better run-export infrastructure, which will likely (probably? certainly?) require repodata changes. So v2 would be the moment to pull that off, because I'm not assuming we'll be doing this very often.

wolfv:
When implementing support for conditional dependencies, we noted that using `if foobar * and or bizbar` might be confusing if `and` is used as a _build string_.

I think it might be sensible to change `and` and `or` to use `&&` and `||` instead (C/C++ syntax) which are characters that are not allowed anywhere in a matchspec.

Thoughts? Looks less Pythonic for sure.

We could also allow `and` and `or` as "inputs" and change to `&&` and `||` in the indexer.

Or we disallow build strings like exact match `and`/`or` and require the user to type `foobar[build_string="and"]` instead.

jaimergp (changes requested):
The new features proposed in this CEP are really exciting and I want to see this happen, so count with my support! Before I delve into the technical details, I'd like to mention a few editorial changes that would be required (and I'm happy to assist in fixing those). Right now, it feels that this is more of a "pre-CEP" writeup. Namely:

- The template is not being followed. None of the proposed sections are used. It's particular important to have: Abstract, Rationale, Specification, Rejected Ideas, and References. The Copyright notice is mandatory.
- There are a number of related PRs on-flight that can help with the definitions and context. For example, the `repodata` schema can be simply referred to as "the repodata schema as defined in CEP PR #135", and then marked as part of the "Requires" frontmatter metadata.
- There are some missing or outdated references to implementation PRs (e.g. https://github.com/conda/rattler/pull/1040 was closed in favor of https://github.com/conda/rattler/pull/1550; https://github.com/conda/rattler/pull/1816 is missing).
- As mentioned elsewhere, this PR should be better split in three parts; one per new field. All of them can "require" each other so they are all part of the v3 repodata update.
- I recommend resorting to MUST, SHOULD and MAY in the Specification section to clarify what's allowed and what's not.

Now onto the actual contents:

- The repodata version 2 is taken. This needs to be version 3.
- The conditionals syntax needs to be better specified. Right now it's unclear what happens with space-separated matchspecs, especially if they happen to have a build string `and`. I recommend constraining this so we don't end up with nested conditionals and other complications (unless intended!).
- The strings allowed in "flags" are not well described. Maximum length, allowed characters and other specificities would be welcome.
- The Backwards Compatibility section should cover what happens with the v1 repodata when a package carries dependencies with the new features. Would the two repodatas get out of sync? Are we expecting some packages to be only available in the new repodata? Are both repodata versions supposed to carry the same content? What's the expected behavior here?
  - See [this `conda-whl-channel` review](https://github.com/conda/conda/issues/15380#issuecomment-3539125427) for some tricks that may be useful here.
- Can packages depend on other packages with their extras? If so, how is this specified? Recipe formats have only allowed `name [version [build]]` syntax so far. Is this the moment where we require repodata v3 to _only_ use the canonical form pre-standardized in https://github.com/conda/ceps/pull/82?

There's maybe some more nuance to this, but this is what I got after reading it again today! Thanks for listening 🤓 And I reiterate: I'm happy to help write this up in the format that we need!

baszalmstra:
How do we deal with quotes in quotes if we want to express: `foo[if="bar[build=\"*gpu\"]"]`. Escape them like the above? Is that already something that is defined in other CEPs?

jaimergp:
> How do we deal with quotes in quotes if we want to express: `foo[if="bar[build=\"*gpu\"]"]`. Escape them like the above? Is that already something that is defined in other CEPs?

I think we support both single and double quotes, so they can be mixed, but also escaped. That should be part of https://github.com/conda/ceps/pull/82.

cep-XXX1.md
```diff
@@ -0,0 +1,132 @@
...
+requirements = [
+  "typing-extensions; python_version < '3.9'",
+  "pywin32; sys_platform == 'win32'"
+  "requests",
+]
+```
+
+Can only be converted into conda packages by making a potentially `noarch: python` package (because it's pure Python) a non-noarch package distributed as as many as 25 platforms (five supported Python versions, times five platforms). This way, the conditional dependency is resolved (in a way) at build time by rendering the dependencies into the granular artifacts. For example, the `typing-extensions` dependency would only be present in the variants built for Python 3.8, and we would only find `pywin32` in the packages published to the `win-*` subdirs.
+
+The adoption of conditional dependencies in conda would allow packagers to express the same complexity in a single `noarch: python` package, because these conditions would be evaluated at "solve time". Providing these capabilities requires new `MatchSpec` syntax and a bump in the repodata version to allow the new syntax in the `depends` field.
```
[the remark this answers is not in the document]
jaimergp: I kind of like the idea of not breaking the schema completely, but base updates on additive fields. In that case, why don't we let go of the `"packages.whatever": {key: package record}` idea and devise a schema under a new key that is better future-proofed? For example a new top-level key that is `v3` or similar: its value can be whatever we need it to be.

What we need to nail down is how to transition from version to version without leaving users behind, and without having partial availability. Some sort of "migrated_at"  timestamp so the client errors can be informative?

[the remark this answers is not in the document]
baszalmstra: The problem is that that doesnt really allow structural changes to json; you already have to parse the depends to know how to parse the depends.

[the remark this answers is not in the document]
wolfv: There is actually a `__conda` virtual package IIRC - which obviously doesn't work / translate well to other clients. Changing that for a "conda-spec" virtual package would make sense IMO.

However, as Bas pointed out, I don't think that would work for this transition. Good material for another CEP though! :)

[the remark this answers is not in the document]
jaimergp: There you go: https://github.com/conda/ceps/pull/146

cep-XXX3.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Rationale
+
+...
+
+## Specification
+
+### Repodata record syntax
+
+The `info/index.json` file of each conda artifact MUST support a new field, `flags`, the value of which MUST be a list of non-empty strings matching the regex `^[a-z0-9\-_:+\.]+$`. Subsequently, the `schema_version` value MUST be bumped to `5`.
```
baszalmstra: What is `+` used for?

wolfv:
We would like to start the RFC period for this CEP - until April 16th when we want to start the vote! @conda/steering-council

cep-XXX2.md
```diff
@@ -0,0 +1,103 @@
...
+{
+  // ...,
+  "packages.conda": {
+    // ...,
+    "example-1.0-0.conda": {
+      // ...,
+      "depends": [
+        "main-dependency"
+      ]
+      "extras": {
```
baszalmstra: I rattler we now call this `extra-depends`, e.g. if a certain extra is enabled these depends are added. The reason is that I can imagine that we could also have constraints being activated through extras. WDYT? @danyeaw @wolfv

cep-XXX3.md
```diff
@@ -0,0 +1,91 @@
...
+
+The problem with this approach is that it doesn't scale well for more than one "flag" per build string. What if a package needs to distinguish among more than one feature? That is, not just GPU vs CPU, but also BLAS backend, MPI or licensing? The glob strings are not expressive enough for a single spec, so several ones need to be supplied (e.g. `package=*=*gpu*`, `package=*=*mkl*`, `package=*=*mpich*` and `package=*=*nogpl*`), resulting in complicated lookahead regexes that balloon in computational complexity.
+
+The answer to this problem is to provide a specific field that is designed to provide such selection capabilities without the expressivity and complexity problems observed above: `flags`.
```
baszalmstra, line 31: How should records that only differ by flags be sorted? That is currently ambiguous. I think it would make sense to add a `variant_order` field to the record. This would be used to distinguish between variants after comparing the versions and build numbers. So first sort by version, then by build-number and then by variant-order, and then by complex trickery with dependency comparisons. That allows us to get rid of weird ordering that has to be done by artificially incrementing the build number.

cep-XXX2.md
dholth: It makes sense to repeat the word "depends" in the second dict. In Python's pyproject.toml it's called https://packaging.python.org/en/latest/specifications/pyproject-toml/#dependencies-optional-dependencies

cep-XXX3.md
dholth: In wheel the client orders each variant that it supports to rank them.

wolfv: @dholth yes, but how?

wolfv: I am fine with adding a variant order field, but it doesn't really change the situation from today (where we already have this ambiguity in variants - just that we don't have flags to match on).

baszalmstra: I indeed didnt specify how the ordering would come to, but my idea would be that in a recipe we could define which flags we prefer over others. Since the builder is most often responsible for expanding the build matrix, it could then compute the variant order.

@wolfv Didn't you have a prototype somewhere that computed this ordering?

wolfv: yep, I had a simple website, just like yours :D Need to find that project again. I can try to re-add this. We should probably also add the `foo:bar` grouping syntax already that we can later expand further with additional matchers. 

@dholth do you have opinions whether we should have strictly only exact string matching, or allow globbing / regexes as well?

dholth: In general I like exact matching because otherwise users imagine that each flag has its own internal syntax but I don't have specific advice for this spec.

cep-XXX2.md
danyeaw: Hey @baszalmstra, I'm good with us calling it this as well. However, isn't this currently called `extra_depends` in Rattler, not `extra-depends`?

cep-XXX1.md
```diff
@@ -0,0 +1,134 @@
...
+
+This CEP extends [CEP 29](./cep-0029.md) with a new keyword, `when`.
+
+### Syntax
+
+A conditional dependency is defined as a `MatchSpec` string that features a `when` keyword, the value of which MUST be a string that encodes a logical expression of one or more `MatchSpec` queries. The logical expression follows a Python-like syntax: `MatchSpec` strings MAY be joined with operators `and` (logical AND) and `or` (logical OR), and grouped within parentheses `()` for precedence overrides.
+
+The inner `MatchSpec` queries inside `when` MUST be expressed in their square brackets syntax, with the exception of simple `name` and `version` queries that MAY be expressed as `{name}{operator}{version}` (no space separators). These inner `MatchSpec` queries MUST NOT feature their own `when` field.
+
+If necessary, the `when` value MUST be quoted to avoid parsing ambiguities.
```
chenghlee: I would drop the "If necessary, " part and just say "The `when` value MUST be quoted to avoid parsing ambiguities."

cep-XXX1.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Specification
+
+This CEP extends [CEP 29](./cep-0029.md) with a new keyword, `when`.
+
+### Syntax
+
+A conditional dependency is defined as a `MatchSpec` string that features a `when` keyword, the value of which MUST be a string that encodes a logical expression of one or more `MatchSpec` queries. The logical expression follows a Python-like syntax: `MatchSpec` strings MAY be joined with operators `and` (logical AND) and `or` (logical OR), and grouped within parentheses `()` for precedence overrides.
+
+The inner `MatchSpec` queries inside `when` MUST be expressed in their square brackets syntax, with the exception of simple `name` and `version` queries that MAY be expressed as `{name}{operator}{version}` (no space separators). These inner `MatchSpec` queries MUST NOT feature their own `when` field.
```
chenghlee, line 58: We may need to be more restrictive on what types of `MatchSpec` clauses can appear in `when`, especially since square bracket clauses as described in CEP 29 are relatively unconstrained and could lead to highly ambiguous parse trees. E.g., you could have square bracket clauses that themselves use quotes, parentheses, `and`, `or`, etc.

For the purposes of this CEP, it might be easier if we just start by allowing only `{name}{operator}{version}` clauses and expand the set of allowable expressions in follow up CEPs as the need arises.

cep-XXX2.md
baszalmstra: Yep, you are totally right, snake_case is used in repodata.json

cep-XXX1.md
baszalmstra: We have a parser that works. From what perspective do you want to restrict this more? We tried to avoid ambiguity by restricting to `{name}{operator}{version}` already.

You probably have a few ideas of what ambiguous cases look like? Maybe you can share those?

cep-XXX1.md
baszalmstra:
```suggestion
The `when` value MUST be quoted to avoid parsing ambiguities.
```

cep-XXX1.md
```diff
@@ -0,0 +1,134 @@
...
+<tr><td> Requires </td><td> https://github.com/conda/ceps/pull/146 </td></tr>
+</table>
+
+## Abstract
+
+This CEP proposes the introduction of conditional dependencies in the conda ecosystem. The work comprises two changes: the addition of a new keyword argument, `when`, to the `MatchSpec` syntax, and its evaluation at solve time. The value of the `when` argument is always a logical expression of one or more MatchSpec queries targeting the solver context.
+
+## Motivation
+
+Conditional dependencies are often requested in the conda ecosystem. After all, an equivalent feature exist in Python packaging in the form of "environment markers" (introduced in [PEP 508](https://peps.python.org/pep-0508/)), so it's a reasonable expectation. There are even traces of planned support of it in `conda.models.match_spec` (see commit [`ab33436`](https://github.com/conda/conda/pull/7606/changes/ab33436591cd2f8ce9c792a50a412d1d02c75504)) in the context of ["pip interoperability"](https://github.com/conda/conda/issues/7053).
```
lucascolley:
```suggestion
Conditional dependencies are often requested in the conda ecosystem. After all, an equivalent feature exists in Python packaging in the form of "environment markers" (introduced in [PEP 508](https://peps.python.org/pep-0508/)), so it's a reasonable expectation. There are even traces of planned support of it in `conda.models.match_spec` (see commit [`ab33436`](https://github.com/conda/conda/pull/7606/changes/ab33436591cd2f8ce9c792a50a412d1d02c75504)) in the context of ["pip interoperability"](https://github.com/conda/conda/issues/7053).
```

cep-XXX3.md
```diff
@@ -0,0 +1,122 @@
...
+
+Similarly, the `build.number` field is currently overloaded with "variant order". To remediate this, we introduce a new integer field specifically to order variants of the same package version - the variant order field. Like the build number, it is a single integer.
```
lucascolley:
```suggestion
Similarly, the `build.number` field is currently overloaded with "variant order". To remediate this, we introduce a new integer field specifically to order variants of the same package version - the `variant_priority` field. Like the build number, it is a single integer.
```

cep-XXX3.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Specification
+
+### Repodata record syntax
+
+The `info/index.json` file of each conda artifact MUST support two new fields, `flags` and `variant_priority`.
+
+The value of the `flags` field MUST be a list of non-empty strings matching the regex `^[a-z0-9_]+(:[a-z0-9_]+)?$`. We allow a _single_ `:` for `key:value` semantics.
+
+The value of the `variant_priority` field must be a positive or negative integer.
```
lucascolley: Can it be `0`?

cep-XXX3.md
```diff
@@ -0,0 +1,122 @@
...
+
+Subsequently, the `schema_version` value MUST be bumped to `3`.
+
+In recipes, these values MUST be supported in the `build` section of each output (i.e. sibling to `number` and `track_features`).
+
+In recipes, it MUST be represented as a list of strings under the `build.flags = [str]` key for each package output.
+For the variant order `build.variant_priority: int`.
```
lucascolley:
```suggestion
In recipes, these values MUST be supported in the `build` section (i.e. sibling to `number` and `track_features`),
represented under the `build.flags: [str]` and `build.variant_priority: int` keys respectively,
for each package output.
```

cep-XXX3.md
```diff
@@ -0,0 +1,122 @@
...
+### Package solver sort order changes
+
+Package sorting MUST take into account the `variant_priority` field. If absent, it defaults to `0`. Higher values are preferred over lower values.
+
+Sorting MUST use the following criteria, in order of decreasing precedence:
+
+1. Tracked features (fewer is preferred)
+2. Version (higher is preferred)
+3. Build number (higher is preferred)
+4. Variant priority (higher is preferred)
```
lucascolley: Feel free to ignore if I have missed discussion, but this surprises me? Suppose we started shipping ASAN and TSAN variants of NumPy alongside the regular builds, with a lower variant priority. If someone requests the `numpy` matchspec, and the ASAN variant has a higher build number (e.g. due to a bugfix in the recipe), will the solver prioritise the ASAN variant? That seems wrong?

lucascolley: In other words it seems to me variant priority should take precedence over build number

cep-XXX2.md
```diff
@@ -0,0 +1,103 @@
...
+<tr><td> Created </td><td> Feb 5, 2025</td></tr>
+<tr><td> Updated </td><td> Mar 9, 2026</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/111 </td></tr>
+<tr><td> Implementation </td><td> TBD </td></tr>
+<tr><td> Requires </td><td> https://github.com/conda/ceps/pull/146 </td></tr>
+</table>
+
+## Abstract
+
+This CEP proposes the introduction of optional dependency groups in conda packaging. Optional dependency groups (or "extras") allow packagers to specify additional requirements behind a named set. This is implemented by adding a new repodata record field, `extras`, as a dictionary that maps group names to a list of dependencies. Extras can be selected via its corresponding `MatchSpec` keyword matching one or more group names.
```
lucascolley:
```suggestion
This CEP proposes the introduction of optional dependency groups in conda packaging. Optional dependency groups (or "extras") allow packagers to specify additional requirements behind a named set. This is implemented by adding a new repodata record field, `extras`, as a dictionary that maps group names to a list of dependencies. Extras can be selected via the corresponding `MatchSpec` keyword matching one or more group names.
```

cep-XXX2.md
```diff
@@ -0,0 +1,103 @@
...
+
+- The `{package}-with-{group-name}` pattern can be found in projects like `ray` or `arrow`.
+- The `{package}` + `{package}-base` pattern was introduced for `matplotlib`.
+- Some projects choose to add all extras in the base package to avoid the complexity of having to add more outputs to their recipe.
+
+Instead of overloading the repodata documents with more package names (which causes additional solver strain and works against optimization efforts like sharded repodata), authors propose to extend the repodata records with the additional dependency metadata, thus reducing unnecessary duplication.
+
+## Rationale
+
+The chosen keyword in the and `MatchSpecs` is  `extras` given its popularity in Python packaging.
```
lucascolley: In the what? Repodata record? And nit: `MatchSpec`s or just `MatchSpec`

cep-XXX3.md
baszalmstra: Yes

cep-XXX3.md
baszalmstra: The idea is that a builder like rattler build usually builds all variants at once. Different variants can have different ordering. Changing the recipe results in a new build of all variants.

Are you saying that in the tsan, asan case these are build with a different cadance?

cep-XXX1.md
```diff
@@ -0,0 +1,134 @@
...
+```shell
+$ conda create -d "python[when=__unix]"
+InvalidMatchSpec: Invalid spec 'python[when=python]': Invalid spec 'python[when=python]': Cannot match on field(s): {'when'}
+```
+
+When used as part of the runtime requirements of a newly built package, the conditional dependency must be included in the `depends` field of the resulting repodata record. Since this new syntax is backwards incompatible with older clients, the resulting `repodata.json` documents (or a derivative, like the sharded forms) MUST NOT include records of packages whose `info/index.json` features a value of `schema_version` equals to or greater than 3.
+
+## Rejected ideas
+
+- `...; if syntax` and other backwards compatible proposal that would be inadequate because then the condition would be ignored, hence creating invalid solutions.
```
lucascolley: Do we mean
```suggestion
- `...; if ...` syntax, and other backwards compatible proposals, are rejected because the condition would be ignored by solvers which do not implement this CEP, resulting in invalid solutions.
```

cep-XXX3.md
lucascolley: Ah, indeed my worry is resolved if the highest build number is always equal across variants. Is that something feasible to enforce, or are there situations where we wouldn't want that to be the case?

lucascolley: Also: maybe not realistic, but would this also introduce races where a lower-priority variant is uploaded to a channel before a higher-priority variant, and the matchspec is queried at the same time?

cep-XXX3.md
lucascolley: Maybe this sentence isn't true, then, depending on your definition of positive.

cep-XXX3.md
lucascolley: > Ah, indeed my worry is resolved if the highest build number is always equal across variants. Is that something feasible to enforce, or are there situations where we wouldn't want that to be the case?

If this is the case, though, why does it matter for build number to take precedence over variant priority? Wouldn't they be equivalent?

lucascolley: > Also: maybe not realistic, but would this also introduce races where a lower-priority variant is uploaded to a channel before a higher-priority variant, and the matchspec is queried at the same time?

This could definitely be a problem with manual `pixi upload`, right?

baszalmstra: Yeah this is my concern: 

> Also: maybe not realistic, but would this also introduce races where a lower-priority variant is uploaded to a channel before a higher-priority variant, and the matchspec is queried at the same time?

If a package originally has 3 variants, those builds get variant orders 0, 1, and 2. Later, one variant is removed and the recipe is rebuilt. If the solver compares variant order across all builds indiscriminately, an older build with a higher variant order can still outrank a newer build, even though the newer build reflects the current recipe.

When the recipe meaningfully changes, you bump the build number. Variant ordering should then only be used to rank variants within the same build number, not across different build numbers.

baszalmstra: Maybe a better explanation:

Variant order is only meaningful within a single build (i.e. for the same build number). If we compare variant order globally (before build number), it would need to be stable across builds, which it generally isn’t.

lucascolley: This still seems fragile to me. I meant the following race-condition:
- T0: user requests `foo`, gets default variant `foo-A-0`
- T1: `foo` is rebuilt for both variants `foo-A` and `foo-B`
- T2: `foo-B-1` is uploaded
- T3: user requests `foo`, gets non-default variant `foo-B-1` because the build number is higher than `foo-A-0`
- T4: `foo-A-1` is uploaded
- T5: user requests `foo`, gets default variant `foo-A-1`

The variant changing based on minor upload time discrepancies feels wrong to me. If the variant priority came first then that would be fixed:
- T3: user requests `foo`, gets default variant `foo-A-0` since it has higher priority than `foo-B-*`

---

> If a package originally has 3 variants, those builds get variant orders 0, 1, and 2. Later, one variant is removed and the recipe is rebuilt. If the solver compares variant order across all builds indiscriminately, an older build with a higher variant order can still outrank a newer build, even though the newer build reflects the current recipe.

Then you can just make sure to avoid dropping non-minimal variant priorities within the same version to resolve this? Sure, there is a lot of space for user error, but this seems like less of a problem than the race-condition, and could perhaps be resolved via linting?

lucascolley: > Variant order is only meaningful within a single build (i.e. for the same build number). If we compare variant order globally (before build number), it would need to be stable across builds, which it generally isn’t.

I guess my point is that it is useful for variant order to be meaningful more broadly than just a single build, and further, that is what users might/would expect.

baszalmstra: Mm I see your point. But I think the variant first order has the same problem, build-first gives temporary flips; version-order-first gives staleness. 

I think the same problem also already exists for how we currently deal with down prioritization. Or variants with the same build number but with different dependencies only. 

Im thinking if it would be possible to enforce that all variants are visible at once.. Looking at wheelnext, which I suppose suffers from the same issue, I can't find a text where this is explicitly mentioned.

cep-XXX3.md
Hind-M: Is there a reason negative values are allowed? (It seems that all the other integer criteria used for sorting are positive?)
Is it to support ordering when `variant_priority` is missing and defaults to 0?
If so, it might be helpful to mention this and maybe add more examples to clarify the use cases?

wolfv:
I just removed the variant_priority field because there are too many open questions about it, especially with regards to _where_ it should sit in the version ordering.

We can (continue) to workaround this by using the build number, or local version numbers.

We want to start voting on these 3 CEPs on Monday.

cep-XXX1.md
jaimergp: Taking one last look here, does `{name}{operator}{version}` include `python>=3,<4` forms or only `python>=3`?

cep-XXX2.md
jaimergp: Took a stab at clarifying this in [e0bfb2e](https://github.com/conda/ceps/pull/111/commits/e0bfb2e0358e104091651f19907c934b5fa9564b)

cep-XXX3.md
jaimergp: I guess we can just say "integers" then?

jaimergp: Looks like this section was removed altogether.

jaimergp:
Superseded by #164, #165, #166.

cep-XXX1.md
jezdez: Re-raised on https://github.com/conda/ceps/pull/164#discussion_r3146270057 since this CEP was split out and voting has been called.

jezdez: > We have a parser that works.

@baszalmstra Hey Bas, that seems uncalled for given Cheng's valid feedback about the CEP. It reads a little passive-aggressive to me, and I'd appreciate it if we followed up on this thread in #164. Given this circumstance of your comment, calling for a vote without resolving it seems a little unfortunate.

PR #112, opened 2025-02-18
[CEP 27] A simple Sigstore predicate

wolfv:
This is the initial CEP of what we are proposing to use as a predicate in the Conda ecosystem when signing packages.


beckermr (changes requested):
Given the discussion we've had, I think we need to be a bit more specific on what we mean by "validating" by the receiving server. Does the server reject packages if they do not match? What about mirroring? What happens if a server disappears or is taken down? 

The destination server information doesn't seem externally verifiable in this broader context.

baszalmstra:
> Given the discussion we've had, I think we need to be a bit more specific on what we mean by "validating" by the receiving server. Does the server reject packages if they do not match? What about mirroring? What happens if a server disappears or is taken down? 
> 
> The destination server information doesn't seem externally verifiable in this broader context. 

The attestations can be validated using the public sigstore fulcia server. This is the beauty of an sigstore attestation. As long as the public instance stays available you can always cryptographically verify that the signee signed for the predicate inside the attestation. This is what the server does, it validates the validity of the attestetion, and through trusted publishing it knows to trust the signer.

From a client perspective this means you just have to trust the signee to trust the predicate inside the attestation. Not the location where you get the attestation from. So if the signee is github we can attest that in a github workflow there was at least the intend to upload the subject (aka the package) to a specific channel. (Note that we dont dont actually know for certain that the package was actually uploaded! But we can assume the intend was there.)

For mirrors this allows anyone to verify that a package was uploaded to the base server at least. This is important because it allows validating that a package that was downloaded from for instance prefix.dev/conda-forge was indeed actually uploaded to anaconda.org and the prefix server didnt have a mallicious package.

beckermr:
If the signature on the attestation is from conda-forge or a github job controlled by conda-forge, then having the target channel in the predicates is duplicate information or simply not useful. From the point of view of mirroring, Seeing that conda-forge signed a package and said "I intended this to be uploaded to blah" is not any different than seeing conda-forge signed a package. The man in the middle attack on a secondary server is not secured by the predicate, but by the initial signature itself.

baszalmstra:
> Where are the attestations stored?

It doesnt really matter because attestations can be stored anywhere and be validated from anywhere. Eg the origin of the attestation doesnt matter for verification. For this CEP I dont think it matters where they are stored.

But on prefix.dev it is already possible to upload attestations. I think it makes sense to standardize  how you can download the attestations associated with a certain package to make it possible for clients to download and verify attestations of packages they install. I think you can expect a followup CEP in the near future.

beckermr:
I think we should make the validation of the predicate within the attestation optional since it is unclear what one does with it. Suppose I get a package with an attestation from prefix that is signed by conda-forge and has a target channel that is not anaconda.org. What do I do with that information? I have verified the signature so the package is actually from conda-forge, but apparently it was intended for upload to another spot. Did conda-forge change its infrastructure or did something else go wrong? There is no way to know a priori without consulting some other source by hand. Verification of the information in the predicates is not possible in general and so should not be required.

baszalmstra:
True! For reference, the pypi predicate also doesnt include any of this information. Its essentially empty. Its just a predicate to be able to sign the subject.

baszalmstra:
But perhaps that might also not be true for non condaforge channels where there isnt a infrastructure like conda forges availabe.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+The `subject` field is already defined in the in-toto specification and contains the name of the package and its digest.
+For conda packages a SHA256 hash MUST be used.
+The subject MUST be the full filename of the conda package that will be part of the repodata.json and under which it will appear on the server.
+
+The `predicateType` field is used to specify the schema of the predicate. The `predicate` field contains the actual predicate data.
+We propose to publish a schema to validate the `predicate` field. The schema will be available at `https://schemas.conda.org/predicate-v1.json`.
+
+The predicate MUST contain the `targetChannel` field, to indicate where the package is being uploaded to. This field MUST be validated by the receiving server. The channel MUST be in canonical form (full URL, no trailing slashes).
```
beckermr, line 61:
```suggestion
The predicate MUST contain the `targetChannel` field, to indicate where the package is being uploaded to. This field's value MAY be validated by the receiving server, but such validation is not possible in general and thus not required. The `targetChannel` MUST be in canonical form (full URL, no trailing slashes).
```

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Discussion
+
+This predicate adds basic verifiable facts about the package. It will tie the producer of the package to the target channel.
+This is similar to what PyPI has implemented with the [PyPI publish attestation](https://docs.pypi.org/attestations/publish/v1/). Since there is no single authoritative index in the Conda world, we add the `targetChannel` field to reach parity.
```
beckermr, line 66:
```suggestion
This predicate ties the producer of the package to the initial target channel for that package. The initial target channel information may become out of date as channels, their locations, names, etc. evolve. This implementation is similar to the [PyPI publish attestation](https://docs.pypi.org/attestations/publish/v1/), but includes the `targetChannel` since there is no single authoritative `conda` package index.
```

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+
+The `subject` field is already defined in the in-toto specification and contains the name of the package and its digest.
+For conda packages a SHA256 hash MUST be used.
+The subject MUST be the full filename of the conda package that will be part of the repodata.json and under which it will appear on the server.
+
+The `predicateType` field is used to specify the schema of the predicate. The `predicate` field contains the actual predicate data.
+We propose to publish a schema to validate the `predicate` field. The schema will be available at `https://schemas.conda.org/predicate-v1.json`.
+
+The predicate MUST contain the `targetChannel` field, to indicate where the package is being uploaded to. This field MUST be validated by the receiving server. The channel MUST be in canonical form (full URL, no trailing slashes).
+
```
beckermr, line 62:
```suggestion

On the server, the certificate MUST be tested against the Trusted Publisher used to upload the certificate to establish a chain of trust.

```

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+## Discussion
+
+This predicate adds basic verifiable facts about the package. It will tie the producer of the package to the target channel.
+This is similar to what PyPI has implemented with the [PyPI publish attestation](https://docs.pypi.org/attestations/publish/v1/). Since there is no single authoritative index in the Conda world, we add the `targetChannel` field to reach parity.
+
+On the server, the certificate should be tested against the Trusted Publisher used to upload the certificate to establish a chain of trust.
```
beckermr, line 68:
```suggestion
```

beckermr (changes requested):
I have made some suggestions top clarify the CEP.

cep-sigstore-predicate.md
baszalmstra: Is it defined what a trusted publisher is? Perhaps we could rephrase tonsomething like:, the server must validate that the signee is a trusted entity. 

What a trusted entity is, is left up to the implementation.

beckermr: I copied this out of the discussion part since it previously read

> On the server, the certificate should be tested against the Trusted Publisher used to upload the certificate to establish a chain of trust.

I assumed whoever wrote it knew what it meant and when they said "should" they actually meant MUST. 

@wolfv can you elaborate?

cep-sigstore-predicate.md
wolfv: @beckermr how would you feel about making it an optional field? I think if the field exists, it should be validated.

cep-sigstore-predicate.md
wolfv: I think we should change it to MUST to have the same guarantees as PyPI.

cep-sigstore-predicate.md
beckermr: I thought the predicate had to be non-empty in order to sign at all?

beckermr: Ah ok. Yes. I'm happy to have it be optional, but if it is present it must be validated. I'll change my review.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+<tr><td> Implementation </td><td>  </td></tr>
+</table>
+
+## Abstract
+
+We want to standardize attestations for the conda ecosystem.
+
+### Sigstore Attestations
+
+Sigstore attestations are cryptographic statements about software artifacts that provide:
```
chenghlee, line 19: Does "SigStore attestation" here mean the SigStore bundle itself or some other attestation payload defined elsewhere, wrapped as the SigStore verification material? To reduce confusion for someone reading those specs, we should use the same [language that SigStore does](https://docs.sigstore.dev/about/bundle/).

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+
+- Authenticity: Proof of who created/signed the artifact
+- Integrity: Verification that the artifact hasn't been tampered with
```
chenghlee, line 22: SigStore cannot directly make any integrity claims; integrity is tied to the sturcture and contents of the verification material, and those are defined outside the SigStore framework.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+- Transparency: Public record of signatures in a tamper-evident log
+
+### Key Components
+
+- Predicates: JSON documents containing metadata about the signing event, using the `in-toto` format
+- Signatures: Cryptographic proofs made using ephemeral keys
+- Rekor: A tamper-evident log that stores attestations
+- Fulcio: A certificate authority that issues short-lived certificates
```
chenghlee, line 30: Is the claim of this CEP to make `rekor` and `falcio` the only acceptable transparency log and certificate authority for the conda ecosystem?

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+    "predicateType": "https://schemas.conda.org/predicate-v1.json",
+    "predicate": {
+        // Canonical URL of the target channel
+        "targetChannel": "https://prefix.dev/conda-forge",
+    }
+}
+```
+
+The `subject` field is already defined in the in-toto specification and contains the name of the package and its digest.
+For conda packages a SHA256 hash MUST be used.
```
chenghlee, line 55: Suggest rewording to:
> For conda packages, the digest MUST include the package's SHA256.

The specification should be clear if only the SHA256 can be included, or if optionally, other package checksums can be included as well.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+In this document, we want to standardize the sigstore predicate for conda packages. The bundle format to be used for sigstore attestations is the `v0.3` bundle format.
+
+## Specification
+
+The in-toto predicate should contain the following fields:
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v0.1",
+    "subject": [{
```
chenghlee, line 41: Is the `subject` field constrained to one and only one package? Or are there circumstances in which multiple packages can be covered by a single attestation?

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+
+Sigstore attestations are cryptographic statements about software artifacts that provide:
+
+- Authenticity: Proof of who created/signed the artifact
+- Integrity: Verification that the artifact hasn't been tampered with
+- Transparency: Public record of signatures in a tamper-evident log
+
+### Key Components
+
+- Predicates: JSON documents containing metadata about the signing event, using the `in-toto` format
```
chenghlee, line 27: Strictly speaking, predicates are not part of the metadata about a SigStore signing event. "Predicates" may exist as part of the SigStore verification material, if said material is a DSSE envelope.

cep-sigstore-predicate.md
chenghlee, line 61: What does the predicate material attest to? Target channel makes it seem like what we want to attest to is some like PyPI Trusted Publishing, but that is not at all clear from the content of this proposed CEP.

If what we want to attest to are claims about the security of build process (a la SLSA), this predicate specification is definitely not sufficient.

chenghlee: Some other notes:

> ...[channel where the] package is being uploaded to...

We might want to clarify or refine that clause. I might be wrong, but a strict reading of that could mean that channel value for conda-forge should be "cf-staging", not "conda-forge". (Since packages are initially uploaded to "cf-staging" and then _copied_ over to "conda-forge".) 


> channel MUST be in canonical form

We need some reference defining what "canonical form" means. I'm not sure there's consensus of a "canonical" channel in the conda-verse, since packages are perhaps surprisingly mobile across channels.

cep-sigstore-predicate.md
chenghlee, line 68: It's not clear to me what which server is meant here, nor what chain of trust is being established. (Missing threat model in the CEP.)

chenghlee (changes requested):

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+## Specification
+
+The in-toto predicate should contain the following fields:
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v0.1",
+    "subject": [{
+        "name": "file-name-0.0.1-h123456_5.conda",
+        "digest": {"sha256": "..."}, ...
```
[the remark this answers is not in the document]
wolfv: It implies that there may be more than one subject. You can find the spec here https://in-toto.io/Statement/v0.1

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+    }],
+    // Schema URL
+    "predicateType": "https://schemas.conda.org/predicate-v1.json",
```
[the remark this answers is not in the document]
wolfv: What would be your proposal? `predicate-base.json`? or `simplePredicate-v1.json`?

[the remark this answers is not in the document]
chenghlee: `predicate-{thing we're attesting to}-v{version}.json`? IMO, the name of the attestation schema should reflect the threat models and associated security claims, which I'm not sure are clearly stated in this CEP.

As an example, PyPI [currently] supports [two different attestation predicate](https://docs.pypi.org/attestations/#supported-attestations), one named "publish" (making claims about who put a package on PyPI) and another named "provenance" (making claims about how the package on PyPI was built). We should follow that pattern.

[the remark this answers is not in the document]
chenghlee: Also, if we're going to follow [in-toto parsing rules](https://github.com/in-toto/attestation/blob/main/spec/v1/README.md#parsing-rules), it might actually be easier if the `predicateType` URI is not actually a file, since said rules require that the URI only expose the major version number. If we use files, we'll likely have to do some symlinking or other magic to make sure we can have schemas for v1.0, v1.1, etc.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,72 @@
...
+
+In this document, we want to standardize the sigstore predicate for conda packages. The bundle format to be used for sigstore attestations is the `v0.3` bundle format.
+
+## Specification
+
+The in-toto predicate should contain the following fields:
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v0.1",
```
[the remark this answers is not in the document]
Callek: This comment is still outstanding

cep-sigstore-predicate.md
Callek: Still an open question (per the current spec, `subject` is one-package only), but worth calling out that the comment is un-answered.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,303 @@
+# CEP - Standardizing a publish attestation for the conda ecosystem
+
+<table>
+<tr><td> Title </td><td> Standardizing a publish attestation for the conda ecosystem </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> Feb 18, 2025 </td></tr>
+<tr><td> Updated </td><td> Feb 18, 2025</td></tr>
```
Callek, line 8: nitpick: updated should be updated.

cep-sigstore-predicate.md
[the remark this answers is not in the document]
Callek: This is now `.../publish/v1` which feels like it resolves this comment

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,303 @@
...
+    compromise.
+
+  Another key misuse-resistance contribution within Sigstore is *machine
+  identities*. A machine identity behaves similarly to a human identity
+  (Alice or Bob), but identifies a machine instead of a human. For example,
+  `github.com/example/example/.github/workflows/release.yml@refs/tags/v1.2.3`
+  could be the machine identity of a GitHub Actions workflow that ran from
+  `release.yml` within `example/example` against the `v1.2.3` tag.
+
+## Motivation
```
Callek, line 88: The motivation still seems *mostly* catered toward "attestations in general" and makes very little mention that this is "attesting about the publication metadata".  I wonder if this could be slightly reworded to emphasize the actual goals of `publish/v1`

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,303 @@
...
+   - The `predicateType` field **MUST** be `https://schemas.conda.org/attestations/publish/v1`.
+   - The `subject[0].name` field **MUST** match the filename of the conda package.
+   - The `subject[0].digest` field **MUST** match the SHA256 hash of the conda
+     package.
+   - The `predicate.targetChannel` field **SHOULD** match the channel that
+     the package was retrieved from, if `predicate` is present. However, the
+     verifier **MAY** choose to allow a channel mismatch, e.g. if the known
+     context is a mirroring context (where the conda package was originally
+     published to a different channel, but is now being consumed from
+     a mirror).
```
Callek, line 212: Possibly not worth addressing in this version of the spec, however I imagine a scenario:

User creates a mirror that pulls in a few specific packages or channels together. User *wants* to validate publish attestations against all the packages as they use them (not at mirror time) -- how can a user identify that YES this package listed as published to `cf-staging`, and this other package to `somesecretorg` are valid, unless they blindly trust the set of `[null, cf-staging, somesecretorg]` to be valid for any package they use, regardless of where it truly came from.

cep-sigstore-predicate.md
Callek: The current CEP still seems to be missing a sense of "what threat will/may this protect against" - It would be great to flesh that out a bit, rather than simply the "what are we specifying in the attestation"

cep-sigstore-predicate.md
Callek: Obsolete Comment?

cep-sigstore-predicate.md
woodruffw: Thanks, I forgot to answer directly here: this CEP as-drafted proposed only a single package subject at a time, since allowing multiple subjects introduces (IMO) a degree of malleability that makes downstream usage harder.

To be precise: if an attestation is allowed to contain multiple subjects, a malicious party could do something like:

```
subjects:
  authentic-0.0.1-etc.conda
  inauthentic-0.0.1-etc.conda
```

...where the `authentic` package identity is one they're authorized (via the server) to publish for, while `inauthentic` is some other package they aren't authorized for. This is turn would make it easy for the server to accidentally accept "mixed" attestations and re-broadcast them as if they're "fully" verified.

This is mentioned under the `Verifying` section as well:

> Servers **SHOULD** perform the same verification process as clients,
with the qualification that the server's trust in the signing identity
is established latently via the server's publishing and upload mechanism.
For example, if the server supports [Trusted Publishing], then the package's
attestation should be verified against the set of Trusted Publisher identities
for that package.

cep-sigstore-predicate.md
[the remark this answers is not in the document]
wolfv: Fixed.

cep-sigstore-predicate.md
wolfv: I think this is outdated after the rewrite.

cep-sigstore-predicate.md
wolfv: I wouldn't want to generalize from what `conda-forge` does. We could very well special case how we validate packages for conda-forge based on the workarounds that are currently applied, but I don't think it's a _nice_ situation that conda-forge is in from which we should spec out how we want to deal with sigstore. 

conda-forge could decide to upload to a S3 bucket first, and then upload to the conda-forge channel, with another Github Action that also makes OIDC trusted publishing work, as an example.

As such, I don't think this needs any clarification.

And yes, for the canonical form, I propose to use the FULL URL for the channel without trailing slash, e.g. `https://anaconda.org/conda-forge`. This has nothing to do with packages being mobile, which is another concern that we could solve in a number of ways (not relevant here).

cep-sigstore-predicate.md
wolfv: Reworded to: 

>     - `subject[0].digest` **MUST** be a [`DigestSet`], and it **MUST** contain
      a single `sha256` entry with the SHA256 hash of the conda package.

cep-sigstore-predicate.md
wolfv: Any suggestions how to reword it?

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+* Ruby (RubyGems): [rubygems/release-gem]
+
+## Specification
+
+### Attestation format
+
+This CEP proposes the following attestation statement layout, using the
+[in-toto Statement schema]:
+
+- `predicateType` **MUST** be `https://schemas.conda.org/attestations/publish/v1`
```
jaimergp, line 129: Sorry if this is an obvious question, but what are the contents of https://schemas.conda.org/attestations/publish/v1? Or is it just a string identifying _this_ attestation format? i.e. is it empty?

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
+# CEP - Standardizing a publish attestation for the conda ecosystem
+
+<table>
+<tr><td> Title </td><td> Standardizing a publish attestation for the conda ecosystem </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
```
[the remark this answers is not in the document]
wolfv: Added!

wolfv:
Hello @conda/steering-council. Requesting a vote for this CEP.

To vote, please use the following form to vote. If you have questions concerning the proposal, you may also leave a comment or code review.

**This vote will end on 2025-07-02, End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

@xhochy (Uwe Korn)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@mbargull (Marcel Bargull)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@jezdez (Jannis Leidel)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [ ]  abstain

@baszalmstra (Bas Zalmstra)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@Hind-M (Hind Montassif)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@trallard (Tania Allard)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@pavelzw (Pavel Zwerschke)
- [ ]  yes
- [ ]  no
- [ ]  abstain

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+   the use of *ephemeral keys* for signing. Modifying the example above:
+
+   - Instead of maintaining a long-lived signing key, Alice generates an
+     *ephemeral key* and binds it to her *identity*
+     ("`alice@trustme.example.com`").
+
+     This binding is done via a certificate issued by [Fulcio], which verifies a
+     *proof of possession* (such as from [OpenID Connect]) from Alice for her
+     claimant identity. The certificate issued by Fulcio is, in turn auditable
+     via [RFC 6962] Certificate Transparency (CT) logs.
```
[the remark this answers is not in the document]
woodruffw: That's whatever identity Alice actually claims, e.g. `alice@example.com` or `https://github.com/alice` or `https://github.com/alice/example/.github/workflows/publish.yml` (the latter being a machine identity for a CI/CD workflow instead of a human identity).

I could probably clarify that by dropping "claimant," since in context these are the only kinds of relevant identities 🙂

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+in part or in full:
+
+* _Who_ (or _what_) published this package?
+* _What_ is the package's hash?
+* _Where_ was this package _published from_, and where _to_?
+* _When_ was this package published?
+
+However, this metadata is not currently **cryptographically verifiable**:
+the consuming party must either trust it as presented, or verify it manually
+against independent sources of truth (such as a project's release history).
```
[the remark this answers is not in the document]
woodruffw: Good catch -- this should probably be "authenticatable" or similar, since that's the relevant "better than integrity" property.

(The underlying argument here is that a hash is "verifiable" in the sense that you can cross-check it, but you don't have an a priori reason to believe that hash is authentic. With a signature you have that authenticity, modulo trust in a key/identity.)

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+This CEP recommends the following signing process:
+
+1. The signer (i.e. Alice or Alice's trusted machine identity) uses a
+   [Sigstore]-compatible client to generate an ephemeral keypair and bind it to
+   their identity via a public certificate.
+2. The signer generates an in-toto statement as described above, and
+   produces an attestation by signing that statement with their ephemeral
+   private key.
+3. The signer uploads their attestation to the Sigstore transparency log
+   as a [DSSE] envelope.
```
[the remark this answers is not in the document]
woodruffw: Yeah, in context this was intended to be the "public good" instance of Sigstore.

> what are the downsides of using a 3rd party service for trust?

In a traditional PKI, I think the biggest downside is unbounded trust extension: a third-party issuer can "go rogue" and do immense damage to the parties that depend on it. Sigstore mitigates that (successfully IMO) with multiple layers of cryptographic transparency (Fulcio's CT log, Rekor for signature transparency).

Apart from that, I think the single biggest downside/risks are stability and longevity: the assumption is that Sigstore (1) remains available, (2) has as-good-or-better uptime characteristics than its downstreams, and (3) doesn't disappear in the next few years. It's my personal opinion that Sigstore satisfies these, which is why I felt comfortable proposing it as a third-party service within PyPI. Inductively GitHub seems to feel the same way, since their public attestation features are built on top of Sigstore's public good instance.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+4. The signer produces a [Sigstore bundle] containing their certificate,
+   attestation, and transparency log inclusion proof.
+
+Each of these steps is performed transparently by a Sigstore client like
+[sigstore-python], except for step (2) as it concerns the specific
+layout of the signed-over statement.
```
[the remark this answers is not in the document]
woodruffw: Ah, this is poorly phrased -- I was trying to split between the things that sigstore-python can do _internally_ versus the things a user needs to feed into it. Everything in the above steps is done by _interacting_ with sigstore-python's APIs or CLI, it's just that (2) is where the user tells sigstore-python _what_ to sign 🙂

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+
+This CEP's design introduces an *unforgeability* property: a package
+that claims a piece of untrusted metadata can be verified by cross-checking
+that metadata against the package's attestation.
+
+An attacker cannot forge or spoof this check, modulo their ability
+to compromise the authentic signing identity itself. This is effectively
+a stronger unforgeability property than traditional signing, as
+the signature binds not only the package's contents, but also the package's
+provenance.
```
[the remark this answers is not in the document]
woodruffw: Yeah, I could probably remove this -- the original intent was to note that the scheme has the same existential unforgeability property (EUF) as its underlying digital signature scheme, but the mention of the domain binding confuses things significantly there.

(The argument that the attacker cannot forge the check is standard under EUF, which is assumed via Sigstore's choice of supported signature schemes. But the part about being "stronger" than traditional signing is a red herring that I'll remove.)

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+This CEP's design introduces a *transparency* property: attestations
+are not considered valid unless they are included in a publicly
+auditable, append-only transparency log. This log is made up of *entries*,
+which bind over the attestation itself (including the subject),
+the attestation's signature, as well as the attestation's signing identity.
+
+Because a valid transparency log inclusion proof is a requirement during
+verification, an attacker who compromises a signing identity *cannot*
+perform a targeted attack without making the attack itself publicly
+discoverable.
```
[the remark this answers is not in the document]
woodruffw: > Can you say something about how this added transparency influences the ability to defend against targeted attacks?

Yeah, can do -- the idea is that in the context of targeted attacks, the victimized party (i.e. the one the attacker is impersonating) can monitor for uses of their identity in real time, meaning that they can respond proactively to the compromise instead of waiting for the first report of compromise from a downstream. I'll fill that in.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+This CEP suggests that attestation verification be performed by
+both clients (i.e. package installers retrieving packages from
+attestation-aware channels) and servers (i.e. attestation-aware channels).
+
+Servers **SHOULD** perform the same verification process as clients,
+with the qualification that the server's trust in the signing identity
+is established latently via the server's publishing and upload mechanism.
+For example, if the server supports [Trusted Publishing], then the package's
+attestation should be verified against the set of Trusted Publisher identities
+for that package.
```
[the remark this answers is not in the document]
woodruffw: I see no reason to have it not be a `MUST` -- the argument for having it be a `SHOULD` is that it _could_ be useful to convey attestations via a server without being able to verify them directly. However we intentionally didn't want to encourage that on PyPI with PEP 740 and so the same likely doesn't apply here 🙂

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+provenance.
+
+### Transparency and auditability
+
+Traditional signing schemes provide integrity and authenticity
+modulo trust in a signing identity.
+
+This is a strong property, but not a perfect one: an attacker who *does* manage
+to compromise a signing identity can mount a *targeted* attack, wherein a
+the general public observes only legitimate artifacts and signatures while
```
Hind-M, line 270:
```suggestion
to compromise a signing identity can mount a *targeted* attack, wherein
the general public observes only legitimate artifacts and signatures while
```

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+<tr><td> Implementation </td><td>  </td></tr>
+</table>
+
+## Abstract
+
+This CEP proposes a standard attestation layout for the conda ecosystem.
+This attestation layout is based on the [in-toto] framework
+and will enable further integration with signing schemes like
+[Sigstore].
+
```
chenghlee, line 19: As a stylistic note, we should add something like this, since it looks like we're using "MUST" in the RFC2119-way:

> ## Specification
>
> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
>
> More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the implementation.

mbargull:
(Not a discussion to be held here since this proposal concerns package content attestation:
Other important, directly user-facing artifacts we distribute are the package indices, i.e., "repodata", which are also "mutable" (due to "repodata patches").
It'd make sense to think about some future-future work regarding of attestation of repodata and/or repodata patches (e.g., to answer "Where did that `depends: ["injected-dep-pkg"]` in the dependency graph come from?" questions) :wink: .)

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
+# CEP - Standardizing a publish attestation for the conda ecosystem
+
+<table>
+<tr><td> Title </td><td> Standardizing a publish attestation for the conda ecosystem </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;, William Woodruff &lt;william.woodruff@trailofbits.com&gt;</td></tr>
+<tr><td> Created </td><td> Feb 18, 2025 </td></tr>
+<tr><td> Updated </td><td> Jun 18, 2025</td></tr>
+<tr><td> Discussion </td><td>  </td></tr>
```
jaimergp, line 9:
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/112 </td></tr>
```

cep-sigstore-predicate.md
jaimergp: Let's go for the JSON schema as Jannis suggested, and let's open the relevant PR in https://github.com/conda/schemas to provide the file.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+      The predicate is identified by a **predicate type**,
+      which defines the predicate's expected schema.
+
+- [Sigstore] is a project that enables misuse-resistant software signing
+  and verification via short-lived certificates and a tamper-evident log.
+  Sigstore composes with attestation frameworks like in-toto to provide
+  transparency and misuse-resistance properties on top of the integrity
+  and authenticity properties of attestations.
+
+   One of Sigstore's major misuse-resistance contributions is
```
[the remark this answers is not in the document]
wolfv: Fixed.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+     context is a mirroring context (where the conda package was originally
+     published to a different channel, but is now being consumed from
+     a mirror).
+
+At the end of this process, the verifier is confident in the following facts:
+
+- The package was published by the signer (Alice or Alice's machine identity).
+  - If the publisher is a machine identity, this further establishes source
+    provenance via the machine identity's claims. See [Sigstore OID information]
+    for additional information on these claims.
```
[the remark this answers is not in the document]
facutuesca: I think "build provenance" might be more appropriate? The machine identity claims are about the build stage (which source revision was used to build, what environment was used to execute the build, what workflow defined the build, etc), rather than about the source.
At least following the definitions in https://slsa.dev/spec/draft/provenance, I think `build provenance` makes more sense.

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+    - Instead of establishing trust a long-lived key from Alice, Bob establishes
+      trust in Alice's identity.
+
+    - Bob can verify the attestation's signature against Alice's emphemeral key,
+      which in turn can be verified as authentically Alice's via the Fulcio-
+      issued certificate.
+
+    With this flow, neither Alice nor Bob needs to maintain long-lived signing
+    or verifying keyrings, in turn reducing the attacker surface for key
+    compromise.
```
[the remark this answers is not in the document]
facutuesca: fixed

cep-sigstore-predicate.md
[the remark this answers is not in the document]
facutuesca: fixed, changed to:
```
However, the authenticity of this metadata is not currently **cryptographically verifiable**:
```

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+An example of a compliant statement is provided below:
+
+```json
+{
+    "_type": "https://in-toto.io/Statement/v1",
+    "subject": [{
+        "name": "file-name-0.0.1-h123456_5.conda",
+        "digest": {"sha256": "01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b"},
+    }],
+    "predicateType": "https://schemas.conda.org/attestations/publish/v1",
```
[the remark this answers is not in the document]
facutuesca: fixed

cep-sigstore-predicate.md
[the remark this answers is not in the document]
facutuesca: fixed, rephrased to:
```
Each of these steps is performed transparently by a Sigstore client like
[sigstore-python], except for the generation of the in-toto statement in
step (2), which should be handled externally and then fed into the Sigstore
client for signing.
```

cep-sigstore-predicate.md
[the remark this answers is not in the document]
facutuesca: fixed, `SHOULD` -> `MUST`

facutuesca:
draft PR in `conda/schemas` opened here: https://github.com/conda/schemas/pull/76

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+   (i.e., Sigstore bundles containing attestations).
+
+    One potential distribution mechanism is to have attestation-aware
+    conda channels distribute each package's attestations alongside
+    the package, or via a similarly discoverable channel-side API.
+    Prior art for this type of distribution mechanism can be found in
+    the PyPI and RubyGems ecosystems, e.g. [PyPI's Integrity API].
+
+3. This CEP specifies an single initial in-toto predicate
+   (`https://schemas.conda.org/attestations/publish/v1`), which conveys
```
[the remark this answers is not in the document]
facutuesca: fixed

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+    - Within in-toto, an attestation's statement is composed of a
+      **subject** and a **predicate**. The subject is the resource
+      (or resources) being attested to, and the predicate is a
+      an arbitrary collection of metadata about the subject.
+      The predicate is identified by a **predicate type**,
+      which defines the predicate's expected schema.
+
+- [Sigstore] is a project that enables misuse-resistant software signing
+  and verification via short-lived certificates and a tamper-evident log.
+  Sigstore composes with attestation frameworks like in-toto to provide
```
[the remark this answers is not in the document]
facutuesca: fixed, changed to `Sigstore uses attestation frameworks like in-toto...`

cep-sigstore-predicate.md
```diff
@@ -0,0 +1,366 @@
...
+for provenance: the package itself describes its source repository,
+its maintainers, etc.
+
+This allows an attacker to forge a package's metadata such that a typical
+downstream unduly trusts the package. For example, a typical downstream
+may observe that the package lists a "trustworthy" GitHub repository
+or organization as its source, and therefore trust the package without
+confirming that its contents actually reflect those of the repository.
+This is compounded when the *distribution format* (such as a conda package
+or Python wheel) does not directly match the claimant repository's
```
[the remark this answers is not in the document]
facutuesca: fixed, reworded to:

```
This is compounded when the *distribution format* (such as a conda package
or Python wheel) does not directly match the source layout (such as the
contents of a GitHub repository), making direct comparison nontrivial.
```

cep-sigstore-predicate.md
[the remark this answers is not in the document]
facutuesca: fixed, removed the part about being stronger:
```
An attacker cannot forge or spoof this check, modulo their ability
to compromise the authentic signing identity itself.
```

cep-sigstore-predicate.md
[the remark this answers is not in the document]
facutuesca: Fixed, added more about how the transparency log helps detecting and responding to compromises:

```
Because a valid transparency log inclusion proof is a requirement during
verification, an attacker who compromises a signing identity *cannot*
perform a targeted attack without making the attack itself publicly
discoverable. This means the log can be monitored in real time for
uses of identities, allowing users to proactively detect and respond
to an identity being compromised, rather than having to wait for the
compromise being detected downstream and reported.
```

PR #113, opened 2025-03-10
CEP 31: Build provenance metadata

### LINKED ISSUE #62: Provenance metadata
Recently we started tracking provenance metadata in conda-forge builds ( https://github.com/conda-forge/conda-smithy/pull/1577#issuecomment-1801465866 )

This follows the same metadata Anaconda appears to use as well ( https://github.com/conda-forge/conda-smithy/pull/1577#issuecomment-1791603046 )

Wonder if we should promote this metadata and document it somewhere. Idk if it should be a CEP or a guideline or example in the doc

Opening here to discuss and track

jaimergp:
## Checklist for submitter

- [X] I am submitting a new CEP: [Build provenance metadata](https://github.com/jaimergp/ceps/blob/provenance/cep-XXXX.md).

- [x] The vote period has ended and the vote has passed the necessary quorum and approval thresholds.
- [x] A new CEP number has been minted. Usually, this is `${greatest-number-in-main} + 1`.
- [x] The `cep-XXXX.md` file has been renamed accordingly.
- [x] The `# CEP XXXX - ` header has been edited accordingly.
- [x] The CEP status in the table has been changed to approved.
- [x] The last modification date in the table has been updated accordingly.

- Closes #62.
- Standardises what conda-forge has been doing for a while, and I think `defaults` too? More details in https://github.com/conda-forge/conda-smithy/pull/1577.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,8 @@ for conda's implementation, all major changes should be submitted as
 | [0027](cep-0027.md) | Standardizing a publish attestation for the conda ecosystem                 |
 | [0028](cep-0028.md) | Customizable system DLL linkage checks for Windows                          |
 | [0029](cep-0029.md) | The `MatchSpec` query language                                              |
+| [0030](cep-0030.md) | Virtual packages                                                            |
+| [0031](cep-0031.md) | Build provenance metadata                                                   |
 
 ## References
 


cep-9999.md
```diff
@@ -0,0 +1,55 @@
...
+## Specification
+
+CI pipelines MUST instrument the build workflows to add the following metadata keys
+via `--extra-meta` or equivalent:
+
+- `sha`: Required. Commit hash of the feedstock being built.
+- `remote_url`: Required on CI. Repository URL of the feedstock being built.
+- `flow_run_id`: Optional. CI-specific identifier for the workflow run.
+
+For local workflows such as those specified by CFEP 03, `remote_url` MAY be omitted, but authors strongly recommend providing the adequate value manually if necessary.
```
JeanChristopheMorinPerso, line 44:
```suggestion
For local workflows such as those specified by [CFEP-03](https://github.com/conda-forge/cfep/blob/main/cfep-03.md), `remote_url` MAY be omitted, but authors strongly recommend providing the adequate value manually if necessary.
```
Also, if `remote_url` is omitted, should `sha` also be omitted?

cep-9999.md
```diff
@@ -0,0 +1,55 @@
...
+scope for this CEP and may be discussed separately.
+
+## Specification
+
+CI pipelines MUST instrument the build workflows to add the following metadata keys
+via `--extra-meta` or equivalent:
+
+- `sha`: Required. Commit hash of the feedstock being built.
+- `remote_url`: Required on CI. Repository URL of the feedstock being built.
+- `flow_run_id`: Optional. CI-specific identifier for the workflow run.
```
JeanChristopheMorinPerso, line 42: The name `flow_run_id` originates from Anaconda's current build system. I wish that we could use something more generic and meaningful. But I guess the boat has sailed already? Or do you foresee a future where CF would be willing to change this to something else (on our side at Anaconda, this is very easy to do).

JeanChristopheMorinPerso (commented):
I'm excited to see this being formalized!

cep-9999.md
jaimergp: Hm, I had assumed users interested in provenance are already using some type of version control, but maybe we can't force that either.

About the dash in CFEP 03, see https://github.com/conda/ceps/blob/971cb232407d2d8c5b23394bcae876725b2f12e7/.pre-commit-config.yaml#L35.

cep-9999.md
jaimergp: We can change it easily too, but then we will have to maintain two ways of accessing this metadata because the already stamped artifacts won't be rebuilt. I think `flow_run_id` is sufficiently generic. I always read it as "(work)flow run ID".

chenghlee:
Is this an intermediate step towards generating [SLSA provenance attestations](https://slsa.dev/spec/v1.0/provenance)? If so, should we just straight in that direction, or would trying to implement SLSA delay the gains we want to get now?

dbast:
thanks @jaimergp for the initiative here.

some background:
* this got enabled in conda-build in Jan 2022 via https://github.com/conda/conda-build/pull/4303
* this got enabled for defaults packages beginning of April 2022
* this got enabled in conda-forge in Nov 2023 via https://github.com/conda-forge/conda-smithy/pull/1577 

@chenghlee yes, it was meant as intermediate step to enable (multiple different) actual attestations via an attestation worker using the data to lookup things:
* process attestation: does the mentioned sha actually exist and was it part of a PR review process?
* automation attestation: the sha after merge to main triggered a CI build matching the flow_run_id and the sha256 of the package is also present in the build log -> proof that a package is automatically build without manual interventions.
* provenance attestation: if the previous attestations are successfully it becomes implicitly clear that the provenance data is correct.

though, if and how those attestations can be stored via e.g. Sigstore in context of SLSA is a different thing.

tl;dr enables attestations that would be hard otherwise without any provenance data.

cep-9999.md
dbast: please keep `sha` as mandatory ... some packages may not have the `remote_url` to not expose private repositories to the public, but the `sha` is this helpful for internal audits / attestations to check that e.g. the `sha` actually exists / was reviewed as part of an PR / triggered a CI run etc.

jaimergp: I am writing this to describe the current conventions, not to impose on how packages should be built. I think that should be decided by packaging organizations separately. I plan to submit a CFEP for conda-forge where we do "require these fields in CI pipelines, as recommended by CEP XYZ".

Someone using conda-build to share an artifact with their research lab internally may not need to care about whether the recipe is version controlled or what a git hash is.

jaimergp:
> Is this an intermediate step towards generating [SLSA provenance attestations](https://slsa.dev/spec/v1.0/provenance)? If so, should we just straight in that direction, or would trying to implement SLSA delay the gains we want to get now?

I'm not personally aiming for that, only wanted to standardise what otherwise was an undocumented convention. I think that SLSA provenance can be iterated on later, and this can just reflect the current state. This way we can refer at non-SLSA provenance like "CEP XYZ metadata".

cep-XXXX.md
```diff
@@ -0,0 +1,81 @@
...
+out of scope for this CEP and may be discussed separately.
+
+## Specification
+
+Build provenance metadata is optional. If necessary, the following metadata keys MAY be used to
+record the corresponding information:
+
+- `sha`: String. Full commit hash of the recipe repository being built.
+- `remote_url`: String. CVS URL of the recipe repository being built. HTTP(S) preferred.
+- `flow_run_id`: String. CI-specific identifier for the workflow run.
```
baszalmstra: Is there an example on how this value can be used? I dont think conda-metadata-app uses this right?

PR #114, opened 2025-03-10
Draft another CEP for PURL annotation

jaimergp:
Rehashing discussion in https://github.com/conda/ceps/pull/63, more focused on `about` rather than `index`.

### DIFF
--- a/cep-9999.md
+++ b/cep-9999.md
@@ -47,7 +47,6 @@ package:
 about:
   # ...
   purls:
-    - pkg:pypi/django
     - pkg:pypi/django@{{ version }}
 ```
 
--- a/cep-9999.md
+++ b/cep-9999.md
@@ -16,7 +16,7 @@ This CEP aims to standardize how the conda ecosystem can leverage free-form
 artifact metadata to annotate source and ecosystem provenance of the distributed
 artifacts. 
 
-We propose using the `extra` field in the recipe metadata to implement
+We propose using the `about` field in the recipe metadata to implement
 a `purl` field where one or more [Package URLs](https://github.com/package-url/purl-spec)
 (PURLs) can be specified.
 
--- a/cep-9999.md
+++ b/cep-9999.md
@@ -54,7 +54,7 @@ about:
 ## Specification
 
 The `about` section (top-level or output-specific) in the recipe file MAY
-contain a `purl` field that takes a string or a list of strings as specified
+contain a `purls` field that takes a list of strings as specified
 by the [PURL standard](https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst).
 
 If the build tool does not support this new field in `about`, the free-form `extra`
--- a/cep-9999.md
+++ b/cep-9999.md
@@ -1,0 +1,65 @@
+# CEP ???? - Source provenance metadata
+
+<table>
+<tr><td> Title </td><td> CEP ???? -  Source provenance metadata </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Mar 10, 2025</td></tr>
+<tr><td> Updated </td><td> Mar 10, 2025</td></tr>
+<tr><td> Discussion </td><td> N/A </td></tr>
+<tr><td> Implementation </td><td> N/A </td></tr>
+</table>
+
+## Abstract
+
+This CEP aims to standardize how the conda ecosystem can leverage free-form
+artifact metadata to annotate source and ecosystem provenance of the distributed
+artifacts. 
+
+We propose using the `extra` field in the recipe metadata to implement
+a `purl` field where one or more [Package URLs](https://github.com/package-url/purl-spec)
+(PURLs) can be specified.
+
+## Motivation
+
+Provenance metadata is useful to assess from which sources an artifact was built.
+conda recipes already specify the necessary sources in their `source` section, but this field
+is mostly concerned with source download and preparation for the build. In
+most cases, it will contain one or more URLs or source-controlled repositories. These
+URLs cannot be guaranteed to uniquely identify a project since they may change, and their
+syntax might evolve across releases (e.g. different versioning schemes).
+
+Instead we propose adding a new field that _uniquely_ identifies the project being built.
+The Package URL project is chosen for this endeavour. This is speciallyuseful
+for efforts like dependency mapping across ecosystems. See
+[PEP 725](https://peps.python.org/pep-0725/) for practical applications in the
+context of PyPI <> conda interoperability.
+
+Example for `django`:
+
+```yaml
+# meta.yaml
+{% set version = "1.10.1" %}
+package: 
+  name: django
+  version: {{ version }}
+# ...
+about:
+  # ...
+  purls:
+    - pkg:pypi/django
+    - pkg:pypi/django@{{ version }}
+```
+
+## Specification
+
+The `about` section (top-level or output-specific) in the recipe file MAY
+contain a `purl` field that takes a string or a list of strings as specified
+by the [PURL standard](https://github.com/package-url/purl-spec/blob/main/PURL-SPECIFICATION.rst).
+
+If the build tool does not support this new field in `about`, the free-form `extra`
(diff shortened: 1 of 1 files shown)


cep-9999.md
```diff
@@ -0,0 +1,64 @@
...
+<tr><td> Implementation </td><td> N/A </td></tr>
+</table>
+
+## Abstract
+
+This CEP aims to standardize how the conda ecosystem can leverage free-form
+artifact metadata to annotate source and ecosystem provenance of the distributed
+artifacts. 
+
+We propose using the `extra` field in the recipe metadata to implement
```
JeanChristopheMorinPerso, line 19:
```suggestion
We propose using the `about` field in the recipe metadata to implement
```
`about` instead of `extra`?

cep-9999.md
```diff
@@ -0,0 +1,64 @@
...
+# meta.yaml
+{% set version = "1.10.1" %}
+package: 
+  name: django
+  version: {{ version }}
+# ...
+about:
+  # ...
+  purls:
+    - pkg:pypi/django@{{ version }}
```
JeanChristopheMorinPerso, line 50: Should the CEP specification mandate that PURLs have a version or do we allow versionless PURLs? I'd also be curious to know what's your opinion on the cases where we'll need to use the `generic` type.

jaimergp: I initially thought of using versionless PURLs, but the examples in https://github.com/conda/ceps/pull/63#issuecomment-2430639657 made me consider adding the version because well, it has to be there anyway. That said, I don't think the version field should be required.

About generic, yes and no. Some packages won't have a canonical location or ecosystem (pretty common in C/C++), so the most convenient way is to annotate them as `generic`. [PEP 725 (draft)](https://peps.python.org/pep-0725/) goes into more detail about this. Check also https://github.com/jaimergp/external-metadata-mappings, where I am dabbing with the concept of cross-ecosystem mappings. There's a [Streamlit app](https://external-metadata-mappings.streamlit.app/) too. I do see the value in things like `generic/libtiff`, as well as PEP-725-proposed `virtual:interface/*` and `virtual:compiler/*`. Hopefully this can be [integrated in PURL like `pkg:abstract/*`](https://github.com/package-url/purl-spec/issues/222) though.

cep-9999.md
```diff
@@ -0,0 +1,64 @@
...
```
chenghlee, line 50: Should we consider making the `purls` a mapping, rather than a simple list? This would make the field more easily interpretable and would give us more flexibility to further extend it.

E.g.,
```
purls:
  upstream:
    - pkg:pypi/foo
  vendors:
    - pkg:github/bar
    - pkg:github/baz
```

cep-9999.md
JeanChristopheMorinPerso: >I initially thought of using versionless PURLs, but the examples in https://github.com/conda/ceps/pull/63#issuecomment-2430639657 made me consider adding the version because well, it has to be there anyway.

I think it depends on how you want to use it. If you just want a standard way to map package names, then I guess versionless is fine. But if you want to map to a specific version, then you need the upstream version.

>About generic, yes and no. Some packages won't have a canonical location or ecosystem (pretty common in C/C++), so the most convenient way is to annotate them as `generic`. [PEP 725 (draft)](https://peps.python.org/pep-0725/) goes into more detail about this. Check also https://github.com/jaimergp/external-metadata-mappings, where I am dabbing with the concept of cross-ecosystem mappings. There's a [Streamlit app](https://external-metadata-mappings.streamlit.app/) too. I do see the value in things like `generic/libtiff`, as well as PEP-725-proposed `virtual:interface/*` and `virtual:compiler/*`. Hopefully this can be https://github.com/package-url/purl-spec/issues/222 though.

I have been following your work in https://github.com/jaimergp/external-metadata-mappings and the other repos, and that's great!

I'm asking about generics because we already started using PURLs to map our package names to upstream packages but decided to wait before adding any kind of support for generics. It gets messy pretty quickly. But more importantly, you need to know what you want to use the PURL for. Is it to map to an exact upstream location or is it to map something to something else and you don't really care about the exact source?

I don't know if we should define this in the CEP though.

cep-9999.md
jaimergp: A free-form mapping or a mapping with well defined keys? If the latter, which ones? I don't want to end up in SBOM-like territory.

cep-9999.md
jaimergp: > I'm asking about generics because we already started using PURLs to map our package names to upstream packages but decided to wait before adding any kind of support for generics. It gets messy pretty quickly.

Do you have some examples of the messy parts? Really interested to see other use cases.

JeanChristopheMorinPerso: I have to say that we didn't look very deeply into generics. But I feel like where it gets tricky is that the spec says that `download_url` and `checksum` can be provided. But the thing is that the `download_url` could differ based on the subdir or other things. There are also potential cases where do very different things could be named the same way. Since `generic` doesn't have a concept of namespace, I'm not too sure how we would handle these (though it's hypothetical mostly).

I might also overthinking it (which is very possible).

beckermr:
The design of repodata patches, while it has warts, might help us here. We could designate a specially named conda package that holds the authoritative set of PURLS for that channel. A given channel could upload that package if they choose to do so and that would be the source of truth.

cep-9999.md
```diff
@@ -0,0 +1,64 @@
...
+
+Provenance metadata is useful to assess from which sources an artifact was built.
+conda recipes already specify the necessary sources in their `source` section, but this field
+is mostly concerned with source download and preparation for the build. In
+most cases, it will contain one or more URLs or source-controlled repositories. These
+URLs cannot be guaranteed to uniquely identify a project since they may change, and their
+syntax might evolve across releases (e.g. different versioning schemes).
+
+Instead we propose adding a new field that _uniquely_ identifies the project being built.
+The Package URL project is chosen for this endeavour. This is speciallyuseful
```
mgorny, line 33:
```suggestion
The Package URL project is chosen for this endeavour. This is especially useful
```

beckermr (changes requested):
A few comments here:

1. Unless the conda-forge package itself decides to vendor something that the upstream package does not vendor, then listing vendored deps is the responsibility of the upstream package. A given package should not be responsible for listing "deps of deps" as usual.
2. Another CEP had comments suggesting that the build tooling add a PURL for the conda package itself that is automatically generated. This is problematic for a few reasons.
   - It is wasteful in terms of storage and repodata size. I don't actually care that much, but I don't want to let this point go unsaid. IOW, we do need to specify how to compute PURLs for conda packages, but that is for downstream users of the conda package if they are referencing it. We can provide tooling for them to compute it, but we don't need to pre-compute it everywhere.
   - The point of the PURL, as I understand it, is to introduce, per their docs, "...a standardized URL-based syntax that uniquely identifies software packages, independent of their ecosystem or distribution channel." We want to use the PURL to know what has been repackaged into this package. If we are repackaging another conda artifact (e.g., the intel repackage feedstock in conda-forge), then we include the PURL of the conda artifact there. Including the PURL to identify the package itself is saying "I have packaged myself into myself as a package," which is true but also useless. In a graph sense, it is a self-loop on a node.
   - A PURL is not a substitute for package signing and identification. We need actual package signing for identification.

cep-9999.md
bollwyvl: Re-phrasing here from #63: if not SBOM territory, then at least SBOM _cartographic terms_ for the "functional equivalent" and "vendor/static link" cases, covered as part of closed namespaces in the SBOM specs:

- CycloneDX 
  - [`pedigree`](https://cyclonedx.org/docs/1.7/json/#metadata_tools_oneOf_i0_components_items_pedigree), one of: `ancestors`, `descendents`,  `variants`, `commits`, `patches`, `notes`
- SPDX
  - [`RelationshipType`](https://spdx.github.io/spdx-spec/v3.0.1/model/Core/Vocabularies/RelationshipType/) between artifacts... some of which are packages, but but may be licenses, vulnerability reports, patches, etc.

A strawman narrowing of those terms, maybe only the first needing to land initially:

| what we're trying to say | in CycloneDXese | to SPDXish |
|-|-|-|
|  _this conda package fulfills the purpose of that pypi package_ | `ancestors` | `descendantOf` |
|  _this conda package statically contains, and therefore exposes you to the vulnerabilities of, that rust crate_ | `variants` | `hasStaticLink` |

PR #78, opened 2024-05-31
CEP-17 add language-agnostic plugins

beckermr:
This PR proposes a CEP to add language agnostic plugins.

PR #115, opened 2025-03-11
[CEP XXXX] OCI Storage of Conda Artifacts

beckermr:
This draft CEP has an updated specification for storing conda packages as OCI artifacts. It is an updated form of the specification in PR #70, given the feedback on the previous PR.

**[Rendered CEP](https://github.com/conda/ceps/blob/oci-again/cep-XXXX.md)**

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -14,6 +14,7 @@
 - [ ] The `# CEP XXXX - ` header has been edited accordingly.
 - [ ] The CEP status in the table has been changed to approved.
 - [ ] The last modification date in the table has been updated accordingly.
+- [ ] The table in the README has been updated with the new CEP entry.
 - [ ] The `pre-commit` checks are passing.
 
 
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ for conda's implementation, all major changes should be submitted as
 | [0021](cep-0021.md) | Run-exports in sharded Repodata                                             |
 | [0022](cep-0022.md) | Frozen environments                                                         |
 | [0023](cep-0023.md) | Text spec input files                                                       |
+| [0024](cep-0024.md) | Specification of <code>environment.yml</code> input files                   |
+| [0025](cep-0025.md) | Versioning of Existing conda Standards                                      |
+| [0026](cep-0026.md) | Identifying Packages and Channels in the conda Ecosystem                    |
 
 ## References
 
--- a/cep-0025.md
+++ b/cep-0025.md
@@ -1,0 +1,44 @@
+# CEP 25 - Versioning of Existing conda Standards
+
+<table>
+<tr><td> Title </td><td> CEP 25 - Versioning of Existing conda Standards and Specs </td></tr>
+<tr><td> Status </td><td> Approved  </td></tr>
+<tr><td> Author(s) </td><td> Matthew R. Becker &lt;becker.mr@gmail.com&gt;</td></tr>
+<tr><td> Created </td><td> Mar 14, 2025</td></tr>
+<tr><td> Updated </td><td> Mar 31, 2025</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/119 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+## Abstract
+
+As the conda ecosystem is being standardized and evolving, the existing standards
+(in many cases only documented by the code that implements them) have been called
+`v1` in some cases (e.g., `.tar.bz2` versus `.conda`) and `v0` in others (e.g., `v0` versus `v1` recipe formats).
+This differing nomenclature is confusing. In this CEP, we resolve this issue by establishing conventions on how to number existing and future `conda` standards.
+
+## Specification
+
+All existing conda standards which satisfy the following conditions
+
+- are documented formally in CEPs or are only effectively documented through code implementations
+- have not been superseded by a newer standard labeled `v1`
+
+MUST be versioned at `v1`.
+
+The numbering of build recipe formats is exempt from this CEP due to the fact
+that the new recipe format in [CEP 14](https://github.com/conda/ceps/blob/main/cep-0014.md) specified `schema_version: 1`.
+
+Standards MUST use a version specifier that consists only of numbers and a single period `.`, has a major version,
+and has at most a minor version (i.e. matches the regex `^[0-9]+($|\.[0-9]+$)`). A `v` MAY be prepended
+for clarity, but MUST be removed before comparing versions. Version comparisons and ordering MUST follow
+`conda`'s version comparison and ordering conventions.
+
+## Backwards Compatibility
+
+The `v1` versioning is fully backwards compatible with all current standards, except recipe formats
+which have been explicitly excluded.
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/cep-0026.md
+++ b/cep-0026.md
@@ -1,0 +1,239 @@
+# CEP 26 - Identifying Packages and Channels in the conda Ecosystem
+
+<table>
+<tr><td> Title </td><td> CEP 26 - Identifying Packages and Channels in the conda Ecosystem </td>
+<tr><td> Status </td><td> Approved </td></tr>
+<tr><td> Author(s) </td><td>
+  Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt; <br />
+  Matthew R. Becker &lt;becker.mr@gmail.com&gt; <br />
+  Cheng H. Lee &lt;clee@anaconda.com&gt;
+</td></tr>
+<tr><td> Created </td><td> Mar 11, 2025</td></tr>
+<tr><td> Updated </td><td> Apr 17, 2025</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/116 </td></tr>
+<tr><td> Implementation </td><td> N/A </td></tr>
+</table>
+
+## Abstract
+
+This CEP aims to standardize names and other strings used to identify packages, artifacts and
+channels in the conda ecosystem.
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the
+rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the
+implementation.
+
+### Identifying package artifacts
+
+The conda ecosystem distinguishes between two types of packages:
+
+- Distributable package names: represented by a concrete, downloadable, extractable conda artifact.
+- Virtual package names: not backed by any concrete artifact. They only exist on the client side.
+
+#### Package names
+
+A distributable package name MUST only consist of lowercase ASCII letters, numbers, hyphens,
+periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT
+include two consecutive separators (hyphen, period, underscore).
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and
+underscores. They MUST NOT use two consecutive separators, with one exception: they MUST start with
+two underscores.
+
+Distributable package names MUST match the following case-insensitive regex:
+`^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
+
+Virtual package names MUST follow this regex: `^__[a-z0-9][._-]?([a-z0-9]+(\.|-|_|$))*$`.
+
+In all cases, the maximum length of a package name MUST NOT exceed 64 characters.
+
+#### Version strings
+
+Version strings MUST only consist of digits, periods, lowercase ASCII letters, underscores, plus
+symbols, and exclamation marks. Additional rules apply but are out of scope in this CEP and will be
+discussed separately.
+
+The maximum length of a version string MUST NOT exceed 64 characters.
+
+#### Build strings
+
+Builds strings MUST only consist of ASCII letters, numbers, periods, plus symbols, and underscores.
+They MUST match this regex `^[a-zA-Z0-9_\.+]+$`.
+
+The maximum length of a build string MUST NOT exceed 64 characters.
+
+#### Artifact extensions
+
+Artifact extensions MUST only consist of lowercase ASCII letters, numbers and periods. They must
+start and end with a letter or a number. They MUST NOT include two consecutive periods. They MUST
+match this regex `^[a-z0-9](\.?[a-z0-9])*$`.
+
+The maximum length of a file extension MUST NOT exceed 16 characters.
+
+> The conda ecosystem currently recognizes two artifact extensions: `tar.bz2` and `conda`,
+versioned `v1` and `v2` respectively.
+
+#### Distribution strings
+
+A "distribution string" MAY be used to identify a package artifact, without specifying the
+extension or channel. It MUST match the following syntax:
+
+```text
+[<subdir>/]<package name>-<version string>-<build string>
+```
+
+Distribution strings apply to distributable packages. They are used as the name of
+the directories where artifacts are extracted in the package cache, for example.
+
+Virtual packages MAY be also identified by a distribution string, but in those cases a subdir MUST NOT be present.
+
+> Note: Despite the similarity, distribution strings are not `MatchSpec`-like specifiers and MUST
+> NOT be used as such.
+
+#### Filenames
+
+The filename of distributable conda artifacts is obtained by adding the artifact extension to its
+distribution string (without the subdir, if present). It MUST match this syntax:
+
+```text
+<package name>-<version string>-<build string>.<extension>
+```
+
+The maximum length of a filename MUST NOT exceed 211 characters.
+
+Virtual conda packages do not exist on disk and SHOULD NOT need filename standardization.
+
+### Identifying channels
+
+A conda channel is defined as a URL where one can find one or more `repodata.json` files arranged
+in one subdirectory (_subdir_) each. `noarch/repodata.json` MUST be present to consider the parent
+location a channel.
+
+#### Channel base URLs
+
+The base URL for the arbitrary location of a repodata file is defined as:
+
+```text
+<scheme>://[<authority>][/<path>/][/label/<label name>]/<subdir>/repodata.json
+```
+
+with `<scheme>`, `<authority>` and `<path>` defined by [RFC
+3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2).
+
+Taken the channel definition above, the base URL without trailing slashes is thus:
+
+```text
+<scheme>://[<authority>][/<path>/][/label/<label name>]
+```
+
+For example, given `https://conda.anaconda.org/conda-forge/noarch/repodata.json`, the part leading
+to `noarch/repodata.json` and thus base URL is `https://conda.anaconda.org/conda-forge`. For local
+repodata such as `file:///home/username/channel/noarch/repodata.json`, the channel base URL is
+`file:///home/username/channel`.
+
+When present, each path component MUST only contain lowercase ASCII letters, numbers, underscores,
+periods, and dashes. They MUST NOT start with a period or a dash. They SHOULD start and end with a
+letter or a number. If present, each path component MUST match this regex:
+
+```re
+^[a-z0-9_][a-z0-9_.-]*$
+```
+
+For `file://`-based channel URLs, the path component rules MAY be understood as recommendations
+only.
+
+The maximum length of an individual path component in a channel base URL MUST NOT exceed 128
+characters. The maximum length of a channel base URL SHOULD NOT exceed 256 characters.
+
+To avoid ambiguous `MatchSpec` grammar, the last path component of a channel base URL SHOULD NOT
+match any `subdir` identifiers. If it does, the behavior in this ambiguous case is not defined
+and implementation dependent.
+
+#### Channel names
+
+For convenience, the channel _name_ is defined as the concatenation of `scheme`, `authority` and
+`path` components of a channel URL. At least one of `authority` or `path` SHOULD be present. In
+their absence, the channel name MUST be considered empty, regardless the scheme. Empty channel
+names SHOULD NOT be used.
+
+When the scheme and authority fields are missing, the full URL can be inferred with these rules:
+
+- If the channel name matches the regex `^\.{0,2}[/\\].*$`, or if it matches the regex
+  `^[A-Z]:([\\/].*)?$` (for Windows drives), it SHOULD be understood as the path component of a
+  `file://` URL.
+- Otherwise, the tool SHOULD provide a user-configurable mechanism to use a default scheme and
+  authority, with the provided channel name taken as the rest of the path component. At the time of
+  this CEP's writing, most tools assume the default URL scheme and authority to be
+  `https://conda.anaconda.org`.
+
+#### Subdir names
+
+Channel subdir names MUST either be the literal `noarch` or a string following the syntax
+`{os}-{arch}`, where `{os}` and `{arch}` MUST only consist of lowercase ASCII letters and numbers.
+Non-`noarch` subdirs MUST match this regex: `^[a-z0-9]+-[a-z0-9]+$`.
+
+The maximum length of a subdir name MUST NOT exceed 32 characters.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward
+slashes, periods, and whitespace. They MUST start with a letter. They MUST match this regex:
+`^[a-zA-Z][0-9a-zA-Z_\-\./]*$`. The last `/`-delimited component of a label
+SHOULD NOT match any `subdir` identifier. If it does, the behavior in this ambiguous
+case is undefined and implementation dependent.
+
+The label `nolabel` is reserved and MUST only be used for conda packages which have no other
+labels. In other words, in the space of labels, the empty set is represented by the labels
+`nolabel`.
+
+A URL for a package, repodata, etc. without a label component MUST be assumed to have the default
+label `main`.
+
+The maximum length of a label name MUST NOT exceed 128 characters.
+
+## Backwards compatibility
+
+The conda subdir and package name regexes are backwards compatible with the current `conda`
+implementation (25.3) and all existing packages on the `defaults` and `conda-forge` channels,
+except for the `__anaconda_core_depends` package on the `defaults` channel, which was [deprecated
+in April 2025][anaconda-core-depends-depr]. See [this
+comment](https://github.com/conda/ceps/pull/116#discussion_r1992234677).
+
+The regex for labels was pulled from an anaconda.org error message describing the set of valid
+labels.
+
+As of 2025-03-12T19:00Z, of the ~1.9M channel names on anaconda.org:
+
+- 7,219 violate the regex `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`;
+- 98 violate the regex `^[a-z0-9][a-z0-9_.-]*$` (allowing channel names to end with `_`, `.`, or
+  `-`); and
+- 6 violate `^[a-z0-9_][a-z0-9_.-]*$` (allowing channel names to start with `_`). Of those six,
+  five start with `.`, and the other starts with `~`.
+
+See [this comment](https://github.com/conda/ceps/pull/116#discussion_r1992154574) for more details.
+The authors have excluded the channel names in the last case that start with `.` or `~` given
+possible security implications. A low percentage, ~0.4%, of channels do not match the
+recommendations for channel names above, but are allowed.
+
+The maximum lengths allowed for the different fields have been chosen so the resulting path
+components (directory names, filenames) comfortably fit in a the 255-char maximum limit some
+filesystems impose. As of 2025-03-01T13:00Z, there are no violations of these limits in any of the
+packages published for `conda-forge`, `bioconda` and `defaults`. See [this
+comment](https://github.com/conda/ceps/pull/116#issuecomment-2763392999) and [this
+comment](https://github.com/conda/ceps/pull/116#issuecomment-2759130187) for more details.
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).
+
+<!-- links -->
+
+[RFC2119]: https://www.ietf.org/rfc/rfc2119.txt
+
+[anaconda-core-depends-depr]: https://conda.zulipchat.com/#narrow/channel/457607-general/topic/conda-oci.20incubation/near/512865665


cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+This CEP defines how to store conda packages in an *OCI registry* as *OCI artifacts* (see the [Open Containers Initiative](https://opencontainers.org/) for more details on these terms). Importantly, this CEP does NOT describe how to store conda repodata in an OCI registry or how to generate repodata from an OCI registry. We leave these topics for future CEPs.
+
+## Specification
+
+The conda OCI package specification has three main parts. First, we define how the data in a conda package is stored as an OCI artifact. Second, we codify the allowed characters and format of conda channels, subdirs, labels, and package names. Finally, we define how to generate the container `<name>` and container image `<tag>` (i.e., `<name>:<tag>`) for a conda package in an OCI registry.
+
+Below we use the following terms, some of which are defined by the OCI specs and others of which are specific to conda.
+
+OCI-define terms:
```
jaimergp, line 23:
```suggestion
OCI-defined terms:
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+- *OCI registry*: A service that implements the OCI Distribution Spec (e.g., `ghcr.io`).
```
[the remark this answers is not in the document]
jaimergp: Ah you linked below, nvm.

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+#### conda OIC Artifact Blob/Layer Media Types
+
+We define the following custom media types that MUST be used for the storage of conda packages in an OCI registry.
+
+| Blob type        | Content type              | mediaType                                   |
+|------------------|---------------------------|---------------------------------------------|
+| conda package v1 | .tar.bz2 package          | application/vnd.conda.package.v1            |
+| conda package v2 | .conda package            | application/vnd.conda.package.v2            |
+| package info     | `info` folder as gzip     | application/vnd.conda.info.v1.tar+gzip      |
+| package index    | `index.json` file         | application/vnd.conda.info.index.v1+json    |
```
jaimergp, line 55:
```suggestion
| package index    | `info/index.json` file         | application/vnd.conda.info.index.v1+json    |
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### conda OCI Artifact Blob/Layer Structure
+
+A valid conda package as an OCI artifact MUST have the following layers:
+
+- a layer holding either the `.tar.bz2` or `.conda` package file with the mediaType `application/vnd.conda.package.v1` or `application/vnd.conda.package.v2`, respectively.
```
jaimergp, line 61: Some packages (especially in defaults) are packaged as both .tar.bz2 and .conda. Can't those channels include both layers in the same artifact?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+- a layer holding the `info` folder as a gzipped tarball with extension `.tar.gz` and mediaType `application/vnd.conda.info.v1.tar+gzip`.
+- a layer holding the `index.json` file as plain JSON with the mediaType `application/vnd.conda.info.index.v1+json`.
```
jaimergp, line 63:
```suggestion
- a layer holding the `info/index.json` file as plain JSON with the mediaType `application/vnd.conda.info.index.v1+json`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+#### conda Manifest Annotations
+
+The manifest MUST have the following [Annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#annotations):
+
+- `org.conda.oci.schema` with the value `1` indicating the `v1` specification (defined by this document).
+- `org.conda.package.name` with the name of the conda package.
+- `org.conda.package.version` with the version of the conda package.
+- `org.conda.package.build` with the build string of the conda package
+
+Additional annotations are NOT allowed.
```
jaimergp, line 76: I think we cannot guarantee this because providers like GHCR might add their own too? At any case, we can say something like:

```suggestion
Additional annotations under the `org.conda` namespace are NOT allowed.
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+### Allowed Characters and Formats for conda Channels, Subdirs, Labels, and Package Names
+
+The follow regexes define valid conda channel names, labels, and conda package names:
```
jaimergp, line 80: I assume this only standardizes names for OCI, not the whole ecosystem?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+### Mapping conda Packages with Channels, Subdirs, Labels to OCI Repositories and Tags
+
+Per the [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec) v1.0, the `<name>` and `<tag>` of an OCI repository are subject to the following regexes:
+
+- `<name>`: `^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$`
+- `<tag>`: `^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$`
+
+The [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec) further notes that "Many clients impose a limit of 255 characters on the length of the concatenation of the registry hostname (and optional port), /, and <name> value."
+
+These constraints motivate the following set of rules for mapping conda packages to OCI repositories and tags. More explanation is given below in the Rational section and the example implementation illustrate how these rules work together to create a human-readable and easily parsable OCI repository `<name>` and `<tag>` for a conda package.
```
jaimergp, line 102:
```suggestion
These constraints motivate the following set of rules for mapping conda packages to OCI repositories and tags. More explanation is given below in the Rationale section and the example implementation illustrate how these rules work together to create a human-readable and easily parsable OCI repository `<name>` and `<tag>` for a conda package.
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+The `<channel>` and `<subdir>` MUST be used as-is without modification. The OCI-forms of the package name, label, version, and build string are defined below.
+
+#### Encoding Package Names, Channels, and Subdirs into OCI Repository `<name>`s
+
+The package name MUST be encoded to OCI-form as follows:
+
+- If the package name starts with an `_`, the `_` MUST be replaced with `z`.
+- Otherwise, if the package name does NOT start with an `_`, the package name MUST be prepended by `c`.
+- If the combined string `<channel>/<subdir>/<OCI-encoded package name>` exceeds 128 characters in length, the OCI-encoded package name MUST be replaced by the SHA1 hash of the OCI-encoded package name in hexadecimal form as `h<SHA1 in hexidecimal>`.
```
jaimergp, line 128:
```suggestion
- If the combined string `<channel>/<subdir>/<OCI-encoded package name>` exceeds 128 characters in length, the OCI-encoded package name MUST be replaced by the SHA1 hash of the OCI-encoded package name in hexadecimal form as `h<SHA1 in hexadecimal>`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+The `<channel>` and `<subdir>` MUST be used as-is without modification. The OCI-forms of the package name, label, version, and build string are defined below.
+
+#### Encoding Package Names, Channels, and Subdirs into OCI Repository `<name>`s
+
+The package name MUST be encoded to OCI-form as follows:
+
+- If the package name starts with an `_`, the `_` MUST be replaced with `z`.
+- Otherwise, if the package name does NOT start with an `_`, the package name MUST be prepended by `c`.
```
jaimergp, line 127: Can you elaborate on why these characters were chosen? I assume `z` is just to sort last, and `c` for `conda` just to have something else?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+<channel>/<subdir>/h<hexdigest of the SHA1 hash of OCI-encoded package name>:h<hexdigest of the SHA1 hash of the original OCI tag of the conda package>
+```
+
+The `<channel>` and `<subdir>` MUST be used as-is without modification. The OCI-forms of the package name, label, version, and build string are defined below.
+
+#### Encoding Package Names, Channels, and Subdirs into OCI Repository `<name>`s
+
+The package name MUST be encoded to OCI-form as follows:
+
+- If the package name starts with an `_`, the `_` MUST be replaced with `z`.
```
jaimergp, line 126: Why replaced and not prepended?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+```
+<channel>/<subdir>/<OCI-encoded package name>
+```
+
+#### Encoding Labels, Versions and Build Strings into OCI `<tag>`s
+
+The label, version, and build string are encoded to OCI-form as follows. First the following mapping rules MUST be applied to the label, version and build string in the order listed below:
+
+- `_` -> `__`
```
jaimergp, line 146: Only single underscores are allowed as input, right? Because OCI only allows two consecutive underscores IIRC.

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+- `-` -> `_D`
+- `+` -> `_P`
+- `!` -> `_N`
+- `=` -> `_E`
+- `:` -> `_C`
+- `/` -> `_S`
+- ` ` -> `_B`
+- `\t` -> `_T`
+- `\r` -> `_R`
+- `\n` -> `_L`
```
jaimergp, line 156: Since only two consecutive underscores are allowed, what happens with an input like `something_+else`? It would become `something___Pelse`, which has three underscores. Maybe `_` needs to become `_U`?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### Further Odds and Ends
+
+- If either the OCI-encoded form of package name or the OCI tag for the conda package are hashed, then both MUST be hashed. In other words, only the two forms of the OCI artifact `<name>:<tag>` combination defined above, either unhashed or hashed, are allowed.
+- Once the OCI repository `<name>` and `<tag>` are computed for the conda artifact, the rest of the OCI Distribution Spec MUST be followed.
+- If an OCI registry implementation errors due to length when dealing with the full OCI form of the conda package (e.g., the combination of the registry URL and full OCI repository `<name>` for the conda package exceed some internal limit in the implementation), then that package cannot be stored in that registry and an error MUST be raised.
+- Within conda repodata, the download URL for a conda package in an OCI registry MUST always be prepended by `oci://` to indicate that the download URL is an OCI URL. The registry being used, any required ports etc. are prepended to the conda package OCI repository `<name>:<tag>` to form the full download URL per the OCI v1.0 specification.
```
jaimergp, line 188: I think this topic might require further details, maybe in the upcoming OCI-compatible repodata CEP?

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+## Specification
+
+The conda OCI package specification has three main parts. First, we define how the data in a conda package is stored as an OCI artifact. Second, we codify the allowed characters and format of conda channels, subdirs, labels, and package names. Finally, we define how to generate the container `<name>` and container image `<tag>` (i.e., `<name>:<tag>`) for a conda package in an OCI registry.
+
+Below we use the following terms, some of which are defined by the OCI specs and others of which are specific to conda.
+
+OCI-define terms:
+
+- *OCI registry*: A service that implements the OCI Distribution Spec (e.g., `ghcr.io`).
+- *OCI repository*: A collection of container images with the same repository `<name>` in an OCI registry.
```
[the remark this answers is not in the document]
jaimergp: Ok, no, the repository is actually the OCI image or container name. Confusing but I guess that's their official terminology. I would have understood it better like this:

```suggestion
- *OCI repository*: A collection of OCI artifacts with the container or image `<name>` in an OCI registry.
```

cep-XXXX.md
beckermr: No they cannot. conda index, the last I checked, assumed those packages hold the same information, so we should pick one and we should prefer the `.conda` one. I have added updated text to the spec.

cep-XXXX.md
beckermr: To keep the names as short as possible. We could instead just prepend everything with a `c` at the cost of a single extra character.

cep-XXXX.md
beckermr: I was following the suggestion from @baszalmstra.

cep-XXXX.md
beckermr: you are correct! fixed!

cep-XXXX.md
beckermr: Yep. Changed to `_U` per your suggestion!

cep-XXXX.md
beckermr: I want to include this clause here so that we can start including these urls in existing repodata files even if the repodata focused spec does not exist yet. Is there anything unclear here?

beckermr:
Good catch @jaimergp! I reran my notebooks with `_` going to `_U` and everything works great. 

Interestingly enough, we apparently have no build strings with a double underscore in either `defaults` or `conda-forge`! All OCI-encoded conda artifact names I produced using `_` -> `__` from both of those channels passed the OCI regexes from the Distribution Spec. Fun!

cep-XXXX.md
```diff
@@ -0,0 +1,229 @@
...
+### Allowed Characters and Formats for conda Channels, Subdirs, Labels, and Package Names
+
+The follow regexes define valid conda channel names, labels, and conda package names for the ENTIRE conda ecosystem:
+
+- channel: `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`
+- subdirs: `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`
+- label: `^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`
+- package name: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`
+
+All channels, subdirs, labels, and package names MUST conform to their respective regex in the list above regardless of whether or not they are package inside an OCI registry or a tradition conda package index (e.g., anaconda.org).
```
beckermr, line 87:
```suggestion
All channels, subdirs, labels, and package names MUST conform to their respective regex in the list above regardless of whether or not they are package inside an OCI registry or a traditional conda package index (e.g., anaconda.org).
```

cep-XXXX.md
jaimergp: I think we need to be a bit more specific then. Something like:

> The download URL OCI-mirrored packages MUST use the `oci://` scheme. More specifically, it MUST follow the syntax `oci://<authority>[:port]/[path]/<name>:<tag>`.

Note that `<name>:<tag>` in the path component of the URL is technically illegal; or the colon needs to be %-encoded. We could also add it as one more path component (`.../<name>/<tag>`) or a qualifier (`.../<name>?tag=<tag>`). 

The thing is that I can't find an actual spec in the OpenContainers org where they recommend `oci://` as the scheme, but there are some real world examples of apps having come up with that convention, like `helm` or `kustomizer`.

cep-0021.md
```diff
@@ -0,0 +1,229 @@
+# CEP 21 - OCI Storage of Conda Artifacts
+
+<table>
+<tr><td> Title </td><td> OCI Storage of Conda Artifacts </td>
+<tr><td> Status </td><td> Proposed </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;<br> Hind Montassif &lt;hind.montassif@quantstack.net&gt;<br> Matthew R. Becker &lt;becker.mr@gmail.com&gt;</td></tr>
```
jaimergp, line 6:
```suggestion
<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;<br /> Hind Montassif &lt;hind.montassif@quantstack.net&gt;<br /> Matthew R. Becker &lt;becker.mr@gmail.com&gt;<br /> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;</td></tr>
```

cep-0021.md
```diff
@@ -0,0 +1,229 @@
...
+
+#### Further Odds and Ends
+
+- If either the OCI-encoded form of package name or the OCI tag for the conda package are hashed, then both MUST be hashed. In other words, only the two forms of the OCI artifact `<name>:<tag>` combination defined above, either unhashed or hashed, are allowed.
+- Once the OCI repository `<name>` and `<tag>` are computed for the conda artifact, the rest of the OCI Distribution Spec MUST be followed.
+- If an OCI registry implementation errors due to length when dealing with the full OCI form of the conda package (e.g., the combination of the registry URL and full OCI repository `<name>` for the conda package exceed some internal limit in the implementation), then that package cannot be stored in that registry and an error MUST be raised.
+- Within conda repodata, the download URL for a conda package in an OCI registry MUST always be prepended by `oci://` to indicate that the download URL is an OCI URL. The registry being used, any required ports etc. are prepended to the conda package OCI repository `<name>:<tag>` to form the full download URL per the OCI v1.0 specification.
```
jaimergp, line 188:
```suggestion
- The download URL for a conda package in an OCI registry MUST use the `oci://` scheme to indicate that the artifact will be downloaded from an OCI endpoint. More specifically, it MUST follow the syntax `oci://<authority>[:<port>]/[<path>/]<name>:<tag>`.
```

cep-XXXX.md
jaimergp: Yes, I think these need to be put in a separate CEP (which can be added later or in parallel) where we standardize the naming schemes for all the components in the ecosystem.

cep-XXXX.md
beckermr: Yeah these are not real urls. They are instead what you feed to the OCI runtime utility like the docker command line tool.

cep-0021.md
beckermr: There would be no extra <path> component here as far as I know, right?

jaimergp: Nope, just what you suggested but using the established URL terms (authority, path, etc).

cep-XXXX.md
beckermr: @jaimergp thoughts here?

jaimergp: I'm not sure if a single character is worth the trouble of having a rule for this, specially given how you've assessed that we are far off from the limit in general. But no strong feelings against leaving things as they are written right now.

jaimergp:
Nothing too big, just a couple of observations:

- The copyright statement needs to be put back in.
- A References section that compiles the different URLs mentioned would be welcome.
- I would wait until this CEP is approved to assign a number. There are a few ongoing PRs that might get voted before this one and then it would be confusing. e.g. there's a [PR named "CEP 17"](https://github.com/conda/ceps/pull/78), but CEP 17 ended up being [this one](https://github.com/conda/ceps/blob/main/cep-0017.md). This might be reflecting a problem in how we mint CEP numbers. Happy to discuss further!

cep-XXXX.md
jaimergp: Yes, but they have different SHA's. With `.conda` we need to assume ZSTD decompressors are available, but I guess that's a given now.

cep-XXXX.md
jaimergp: Opened https://github.com/conda/ceps/pull/116

beckermr:
Ah if we don't plan to assign numbers can we adjust the linter then? It being red but only due to the number is pretty annoying since it also spell checks, lints, etc.

beckermr:
Ack @jaimergp, our analysis of the double underscore issues was wrong. Build strings, version strings, and labels go in OCI tags which allow any number of underscores in a row

```
^[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}$
```

So we can and should use the `_` -> `__` encoding to make things visually cleaner.

beckermr:
OK @jaimergp This is ready for one more look.

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+Package names in OCI-form which are not hashed can be decoded by reversing the encoding rules above.
+
+For example, the package name `_libgcc_mutex` is encoded to OCI-form as `c_libgcc_mutex`, and the package name `c_libgcc_mutex` is encoded to OCI-form as `cc_zlibgcc_mutex`. A very long package name like `gblah000...000` would be encoded to its OCI-form `cgblah000...000`, then hashed via SHA1, yielding something like `h<hexdigest of SHA1 hash>`.
+
+The channel and subdir MUST be used as-is without modification.
+
+The channel, subdir and OCI-encoded package name (or hashed OCI-encoded package name if required) MUST be combined with `/` to form the OCI repository `<name>`
+
+```
+<channel>/<subdir>/<OCI-encoded package name>
```
beckermr, line 138:
```suggestion
<channel>/<subdir>/<OCI-encoded package name or its SHA1 hash in hexidecimal>
```

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+- `\t` -> `_T`
+- `\r` -> `_R`
+- `\n` -> `_L`
+
+This encoding can be undone by applying the rules in reverse, starting at the bottom of the list and moving to the top. Depending on the context, some labels may be percent-encoded for use in URLs. The percent-encoding MUST be undone before the label is encoded via the list of rules above.
+
+After encoding each component to OCI-form, the version, build string, and (optional) label MUST be combined with a `-` to form the OCI `<tag>` as either
+
+```
+<version>-<build>-<label>
```
beckermr, line 162:
```suggestion
<OCI-encoded version>-<OCI-encoded build>-<OCI-encoded label>
```

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+or if the label is not specified (i.e. the package is on the `main` label), then
+
+```
+<version>-<build>
```
beckermr, line 168:
```suggestion
<OCI-encoded version>-<OCI-encoded build>
```

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+```
+
+with no trailing `-`.
+
+As implied above, if no label is explicitly specified for a package, then package is by definition on the `main` label. This stipulation means the following two OCI tags for conda packages are equivalent:
+
+```
+v1.0.0-h123456__0-main
+v1.0.0-h123456__0
```
beckermr, line 177:
```suggestion
v1.0.0-h123456_U0-main
v1.0.0-h123456_U0
```

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+
+#### Further Odds and Ends
+
+- If either the OCI-encoded form of package name or the OCI tag for the conda package are hashed, then both MUST be hashed. In other words, only the two forms of the OCI artifact `<name>:<tag>` combination defined above, either unhashed or hashed, are allowed.
+- Once the OCI repository `<name>` and `<tag>` are computed for the conda artifact, the rest of the OCI Distribution Spec MUST be followed.
+- If an OCI registry implementation errors due to length when dealing with the full OCI form of the conda package (e.g., the combination of the registry URL and full OCI repository `<name>` for the conda package exceed some internal limit in the implementation), then that package cannot be stored in that registry and an error MUST be raised.
+- The download URL for a conda package in an OCI registry MUST use the `oci://` scheme to indicate that the artifact will be downloaded from an OCI endpoint. More specifically, it MUST follow the syntax `oci://<authority>[:<port>]/<name>:<tag>`. The `<authority>` is the URI of the OCI registry (e.g., `ghcr.io`).
```
beckermr, line 187:
```suggestion
- The repodata download URL for a conda package in an OCI registry MUST use the `oci://` scheme to indicate that the artifact will be downloaded from an OCI endpoint. More specifically, it MUST follow the syntax `oci://<authority>[:<port>]/<name>:<tag>`. The `<authority>` is the URI of the OCI registry (e.g., `ghcr.io`).
```

cep-XXXX.md
```diff
@@ -0,0 +1,242 @@
...
+- conda packages whose OCI artifact `<name>:<tag>` are not hashed retain human-readability.
+- For conda packages whose OCI artifact `<name>:<tag>` are not hashed, the underlying conda package information can be decoded from the OCI artifact `<name>:<tag>` without needing to access the OCI registry or consult a lookup table.
+- For conda packages whose OCI artifact `<name>:<tag>` are hashed, the underlying conda package information can be extracted from the OCI Annotations stored in the OCI manifest.
+- All labels on `anaconda.org` can be used as-is.
+- Relabeling a conda package in an OCI registry is a cheap operation since it involves retagging an existing image.
+
+Some specific choices were made to ease parsing and avoid edge cases:
+
+- The encoding of labels, versions, and build strings is purposefully short to avoid excessively increasing the length of OCI tags.
+- The encoding of labels, versions, and build strings only generates at most two `_`s in a row and so obeys the OCI spec.
```
beckermr, line 204:
```suggestion
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+
+- The [channel-mirrors project](https://github.com/channel-mirrors) with its associated package [conda-oci-mirror](https://github.com/channel-mirrors/conda-oci-mirror) developed a proof-of-concept implementation of conda packages and repodata stored in an OCI registry. This specification is based on the experience gained from that project, but is not fully backwards compatible with the original implementation. See Backwards Compatibility above. This original implementation and implied specification is referred to as `v0`.
+
+## Sample Implementation
+
+### conda-oci
+
+The [conda-oci](https://github.com/conda-incubator/conda-oci) repository has reference implementations of the procedures defined above for computing the OCI artifact `<name>:<tag>` and for converting a conda artifact into an OCI artifact with a manifest and layers/blobs.
```
jaimergp, line 223: This is not ready yet, right?

beckermr: Nope. My plan is to have it ready before the vote happens. The code parsing/validation is done, but I need to add in the package structure, CI, and code to convert artifacts.

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+
+All channels, subdirs, labels, and package names MUST conform to their respective regex in the list above if the package is to be stored in an OCI registry.
+
+Further, the following rule applies to labels:
+
+- The label `NOLABEL` is reserved and MUST only be used for conda packages which have no other labels. In other words, in the space of labels, the empty set is represented by the label `NOLABEL`.
+
+### Mapping conda Packages with Channels, Subdirs, Labels to OCI Repositories and Tags
+
+Per the [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec) v1.0, the `<name>` and `<tag>` of an OCI repository are subject to the following regexes:
```
beckermr, line 95:
```suggestion
Per the [OCI Distribution Spec](https://github.com/opencontainers/distribution-spec) v1, the `<name>` and `<tag>` of an OCI repository are subject to the following regexes:
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+
+Conda-specific terms:
+
+- *OCI-encoded package name*: The name of a conda package that is encoded to a specific format (defined below) for use in an OCI repository `<name>`. The language "encoded to OCI-form" is used below to refer to this process.
+- *OCI-encoded label/version/build string*: A conda package label, version, or build string that is encoded to a specific format (defined below) for use in an OCI image `<tag>`. The language "encoded to OCI-form" is used below to refer to this process.
+- *hashed OCI-encoded <item>*: An `<item>` that is first encoded to its OCI-form and then hashed to a specific format (defined below) for use in an OCI repository `<name>` and/or `<tag>`.
+
+For further details on the OCI-defined terms, see the full OCI specifications, the [Image Spec](https://github.com/opencontainers/image-spec) and the [Distribution Spec](https://github.com/opencontainers/distribution-spec).
+
+This specification is labeled `v1` and uses only `v1.0` of the OCI specification.
```
beckermr, line 40:
```suggestion
This specification is labeled `v1` and uses only OCI specifications compatible with the OCI `v1` specification.
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+- `:` -> `_C`
+- `/` -> `_S`
+- ` ` -> `_B`
+- `\t` -> `_T`
+- `\r` -> `_R`
+- `\n` -> `_L`
+
+This encoding can be undone by applying the rules in reverse, starting at the bottom of the list and moving to the top. Depending on the context, some labels may be percent-encoded for use in URLs. The percent-encoding MUST be undone before the label is encoded via the list of rules above.
+
+After encoding each component to OCI-form, the version, build string, and (optional) label MUST be combined with a `-` to form the OCI `<tag>` as either
```
beckermr, line 159:
```suggestion
After encoding each component to OCI-form, the version, build string, and label MUST be combined with a `-` to form the OCI `<tag>`
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+<OCI-encoded version>-<OCI-encoded build>-<OCI-encoded label>
```
beckermr, line 162:
```suggestion
<OCI-encoded version>-<OCI-encoded build>(-<OCI-encoded label>)
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+```
+
+or if the label is not specified (i.e. the package is on the `main` label), then
```
beckermr, line 165:
```suggestion
where the `-<OCI-encoded label> component is present is the package label is not `main`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,241 @@
...
+
+```
+<OCI-encoded version>-<OCI-encoded build>
+```
+
+with no trailing `-`.
+
```
beckermr, line 172:
```suggestion
```

beckermr:
Comment from @schuylermartin45: move to better hash to avoid issues with hash-collision attacks

Comment from @Callek: move to SHA256 as a compromise

cep-XXXX.md
```diff
@@ -0,0 +1,237 @@
...
+where the `-<OCI-encoded label>` component MUST be present if the package label is not `main`.
+
+In the hashed form, the OCI-form of the package name and the OCI tag (i.e., `<OCI-encoded version>-<OCI-encoded build>(-<OCI-encoded label>)`) are each separately hashed via SHA1 and then replaced by the string `h<hexdigest of SHA1 hash>` as follows:
+
+```
+<channel>/<subdir>/h<hexdigest of the SHA1 hash of OCI-encoded package name>:h<hexdigest of the SHA1 hash of the original OCI tag of the conda package>
+```
+
+The `<channel>` and `<subdir>` MUST be used as-is without modification. The OCI-forms of the package name, label, version, and build string are defined below.
```
jaimergp, line 122: What if the channel name along is too long to be encoded? Would it be ok to have top-level `h<hash>` entries with the corresponding `org.conda.channel.{name,subdir,label}` annotation metadata?

beckermr: Yes maybe just doing that all the time is simpler actually. I was trying to increase readability, but the complexity cost is high.

beckermr:
i've pushed a huge update @jaimergp. comments welcome!

beckermr:
Looking more at the OCI spec and the working code from conda-OCI-mirror, we'll need to specify a few more things including

- what we do with the top-level config
- any attributes attached to each layer
- how we set the top-level artifact mediaType

beckermr:
pre-commit.ci autofix

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+<OCI conda channel base URL>/<OCI-compatible subdir>
+```
+
+where `<OCI-compatible subdir>` is defined below.
+
+The following regexes MUST be satisfied by the `<OCI-compatible channel path>`, `<OCI-compatible subdir>`, and `<OCI-compatible label>`:
+
+- `<OCI-compatible channel path>`: `^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$` (same as the regex for an OCI repository `<name>`)
+- `<OCI-compatible label>`: `^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$` (same as the regex for an OCI repository `<name>`)
+- `<OCI-compatible subdir>`: `^[a-z0-9]+(-[a-z0-9]+)*$`
```
jaimergp, line 145: This is the same as the one proposed in https://github.com/conda/ceps/pull/116, right? We should update this part to refer to that one once accepted.

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+the OCI repository `<name>` regex):
+
+- `<OCI conda channel base URL>/<OCI-compatible subdir>/r*`: reserved for repodata JSON artifacts, run exports JSON artifacts, and
+  future CEP-defined extensions for sharded repodata (defined in [CEP 16][cep16]).
+- `<OCI conda channel base URL>/<OCI-compatible subdir>/c*`: reserved for OCI-encoded conda package names (The character `c`
+  is prepended to the conda package name to indicate that it is an OCI-encoded conda package name as described below.)
+- `<OCI conda channel base URL>/<OCI-compatible subdir>/h*`: reserved for hashed OCI-encoded conda package names
+- `<OCI conda channel base URL>/<OCI-compatible subdir>/p*`: reserved for repodata patch instructions
+
+These designations are used to prevent namespace collisions between the various components of an OCI conda channel.
```
jaimergp, line 160: Should we add a clause that prevents any other initial letters from being used altogether? Something like:

```suggestion
These designations are used to prevent namespace collisions between the various components of an OCI conda channel. Any other string prefixes MUST NOT be used.
```

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+- All package names MUST be prepended by the character `c`.
+- If the combined string `<OCI conda channel base URL>/<OCI-compatible subdir>/<OCI-encoded package name>` exceeds 128 characters
+  in length, the`<OCI-encoded package name>` MUST be replaced by the SHA256 hash of the `<OCI-encoded package name>`
+  in hexadecimal form as `h<SHA256 in hexadecimal>`. The letter `h` is reserved to indicate a hashed OCI-encoded package name as described above.
+
+`<OCI-encoded package name>`s that are not hashed can be decoded by reversing the encoding rules above.
+
+For example, the package name `_libgcc_mutex` is encoded to OCI-form as `c_libgcc_mutex`, and the package name `c_libgcc_mutex`
+is encoded to OCI-form as `cc_zlibgcc_mutex`. A very long package name like `gblah000...000` would be encoded to its OCI-form
+`cgblah000...000`, then hashed via SHA256, yielding something like `h<hexdigest of SHA256 hash>`.
```
jaimergp, line 183: Let's add a note here saying that the original components can be found as annotation metadata, as discussed below.

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+
+### Conda Artifact OCI Tags
+
+The OCI tag for a conda package MUST be computed from the packages `<OCI-encoded version>` and `<OCI-encoded build string>` as follows.
+
+First, the `<OCI-encoded version>` and `<OCI-encoded build string>` are computed from conda package's `<version>` and `<build string>`
+respectively. To do this, one MUST apply the mapping rules below to the conda package's `<version>` or `<build string>`
+in the order listed, top to bottom:
+
+- `_` -> `__`
```
jaimergp, line 193: We are reverting the `_U` thing from last iteration, right? Just making sure it's not an oversight because in tag names we did find that multiple sequential underscores are ok.

jaimergp: Ah yes, you mentioned this below.

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+| Blob type            | Content type                | mediaType                                      |
+|----------------------|-----------------------------|------------------------------------------------|
+| conda package v1     | .tar.bz2 package            | application/vnd.conda.package.v1               |
+| conda package v2     | .conda package              | application/vnd.conda.package.v2               |
+| package info         | `info` folder as gzip       | application/vnd.conda.info.v1.tar+gzip         |
+| package index        | `info/index.json` file      | application/vnd.conda.info.index.v1+json       |
+| repodata.json        | `repodata.json` file        | application/vnd.conda.repodata.v1+json         |
+| repodata.json.zst    | `repodata.json.zst` file    | application/vnd.conda.repodata.v1+json+zst     |
+| repodata.json.gz     | `repodata.json.gz` file     | application/vnd.conda.repodata.v1+json+gzip    |
+| repodata.json.bz2    | `repodata.json.bz2` file    | application/vnd.conda.repodata.v1+json+bz2     |
```
jaimergp, line 268: Do we need to add `v2` repodatas (CEP 15)?

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+| channeldata.json     | `channeldata.json` file     | application/vnd.conda.channeldata.v1+json      |
+| run_exports.json     | `run_exports.json` file     | application/vnd.conda.run_exports.v1+json      |
+| run_exports.json.zst | `run_exports.json.zst` file | application/vnd.conda.run_exports.v1+json+zst  |
+| run_exports.json.gz  | `run_exports.json.gz` file  | application/vnd.conda.run_exports.v1+json+gzip |
+| run_exports.json.bz2 | `run_exports.json.bz2` file | application/vnd.conda.run_exports.v1+json+bz2  |
+| repodata_from_packages.json | `repodata_from_packages.json` file | application/vnd.conda.repodata_from_packages.v1+json |
+| repodata_from_packages.json.zst | `repodata_from_packages.json.zst` file | application/vnd.conda.repodata_from_packages.v1+json+zst |
+| repodata_from_packages.json.gz | `repodata_from_packages.json.gz` file | application/vnd.conda.repodata_from_packages.v1+json+gzip |
+| repodata_from_packages.json.bz2 | `repodata_from_packages.json.bz2` file | application/vnd.conda.repodata_from_packages.v1+json+bz2 |
```
jaimergp, line 277: These follow the same schema as other repodatas, but are unpatched. Do we need a separate media type? What about `current_repodata`?

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+
+Additional layers are NOT allowed in the OCI artifact. If a conda package exists in both `.tar.bz2` and `.conda` format, the `.conda`
+version MUST be chosen to be in the OCI artifact.
+
+The manifest MUST have the following [OCI Annotations][annotations]
+
+- `org.conda.oci.schema` with the value `1` indicating the `v1` specification (defined by this document).
+- `org.conda.package.name` with the name of the conda package.
+- `org.conda.package.version` with the version of the conda package.
+- `org.conda.package.build` with the build string of the conda package
```
jaimergp, line 303: Shall we add channel, label and subdir too?

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+<OCI conda channel base URL>/<OCI-compatible subdir>/run_exports.json
+<OCI conda channel base URL>/<OCI-compatible subdir>/patch_instructions.json
+```
+
+MUST have a layer with the data as JSON and the appropriate mediaType from the table above. Additional layers with
+`gzip`, `bz2`, or `zst` compressed copies of the data MAY be included.
+
+The manifest MUST have the following [OCI Annotation][annotations]:
+
+- `org.conda.oci.schema` with the value `1` indicating the `v1` specification (defined by this document).
```
jaimergp, line 323: Same here: shall we annotate channel, subdir and label in case they end up being hashed?

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+  notable feedback included the suggestion of
+  [symbol name mangling techniques and the first version of the shortened encoding used above](https://github.com/conda/ceps/pull/70/files#discussion_r1740438239)
+  by [Bas Zalmstra](https://github.com/baszalmstra).
+- The [channel-mirrors project](https://github.com/channel-mirrors) with its associated package
+  [conda-oci-mirror](https://github.com/channel-mirrors/conda-oci-mirror) developed a proof-of-concept implementation of
+  conda packages and repodata stored in an OCI registry. This specification is based on the experience gained from that project, but is not
+  fully backwards compatible with the original implementation. See the [Backwards Compatibility](#backwards-compatibility) section above.
+  This original implementation and implied specification is referred to as `v0`.
+- One alternative to excluding certain labels and channels from distribution via an OCI conda channel is to define a further custom
+  encoding of the channel path parts of the URL. For example, one could use an encoding inspired by percent-encoding for URLS. This encoding
```
jaimergp, line 380:
```suggestion
  encoding of the channel path parts of the URL. For example, one could use an encoding inspired by percent-encoding for URLs. This encoding
```

cep-XXXX.md
beckermr: So to be clear, I didn't end up hashing everything like we discussed. It is just the package name that is hashed. The rationale here is two-fold

1. Packages move between channels and so it didn't make sense to me to hash the channel and label with the package name. 

2. The hashes would occupy a specific namespace in the space of channels. While collisions themselves through normal operations would be rare, I was concerned that there might be security implications for this down the road.

cep-XXXX.md
beckermr: Ah good point. I'll condense to a single repodata schema and add current repodata. Unfortunately current repodata starts with a "c" and so conflicts with a package named _repdata.json. We'll have to resolve that conflict somehow.

cep-XXXX.md
beckermr: They would not be hashed. See above.

cep-XXXX.md
beckermr: Yes

cep-XXXX.md
beckermr: subdir yes, but not channel or label.

cep-XXXX.md
jaimergp: Argh, do we need to prefix something to all `repodata`s then? Because in that case we should prepend the same prefix to all non-package information for consistency. Something like `m`. So `repodata.json` would be stored under `mrepodata.json`, and `patch_instructions.json` would be under `mpatch_instructions.json`?

cep-XXXX.md
beckermr: ah sorry. these entries are never hashed so no need here.

cep-XXXX.md
beckermr: We never hash the subdir so no need actually.

cep-XXXX.md
```diff
@@ -0,0 +1,424 @@
...
+```text
+<OCI conda channel base URL>/<OCI-compatible subdir>/<OCI-encoded package name>
+```
+
+The contents and mediaTypes of OCI blobs that make up the conda package artifact, along with how to compute the OCI tag, are defined below as well.
+
+The `<OCI-encoded package name>` MUST be computed according to the following rules:
+
+- All package names MUST be prepended by the character `c`.
+- If the combined string `<OCI conda channel base URL>/<OCI-compatible subdir>/<OCI-encoded package name>` exceeds 128 characters
```
beckermr, line 175:
```suggestion
- If the `<OCI-encoded package name>` exceeds 64 characters
```

cep-XXXX.md
jaimergp: Sure, that can work. It would just be odd because in a way we are limiting `repodata.json`-like files to _not_ start with `h` or `c`, and at the same time allowing all the other letters implicitly, while also requiring some copies to start with `r`? So I'd rather have a clear and simple rule for non-package files, so they all must start with the same prefix.

beckermr: I see what you mean. That is indeed better so let's go with `m` for `metadata`. I will adjust the CEP.

jaimergp:
All of that is currently valid and sound. I'm just worried that we are being lucky now with no conflicts in the filename prefixes (I do second dropping `current_repodata.json`, we don't need it these days, and we can consider it an Anaconda.org implementation detail if there's such a need).

I just don't see the pain in prefixing metadata files with `m`. If we don't, we might run into a situation where we want to add a new type of files, and the only way out would be to add OCI-specific sub-subdirs like `conda-forge/linux-64/packages/` and `conda-forge/linux-64/metadata/`, and that seems like more painful.

But if you think the burden of prefixing `m` to everything is not worth it, I won't block it. I'd be happy to hear what others think too.

PR #116, opened 2025-03-11
[CEP 26] Identifying Packages and Channels in the conda Ecosystem

jaimergp:
#### [Rendered CEP](https://github.com/jaimergp/ceps/blob/naming/cep-XXXX.md)

Comes from https://github.com/conda/ceps/pull/115#discussion_r1989305660.

Addresses the long needed standardization of package names and channel names.

I took @beckermr's regexes and added some more details and structure. Very open to feedback in how descriptive vs prescriptive we should get. So far, only trying to describe roughly what we have in `conda`. I'd like to know more about how this is handled in `rattler`. cc @baszalmstra @wolfv


cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+Version strings MAY specify an epoch field, which MUST be an integer separated by a single exclamation mark from the rest of the version information. If not specified, the epoch is considered to be `0`. The exclamation mark MUST NOT be used anywhere else.
+
+The version string SHOULD start with a digit.
+
+The plus sign SHOULD separate the regular version information from the local version information.
+The plus sign SHOULD NOT be used for any other purposes.
+
+Multiple separator characters (periods, underscores, plus signs) SHOULD NOT be used contiguously.
+
+The following regex does not cover all rules, but all valid version strings MUST match it: `^([0-9]+!)?[0-9][0-9a-z.(_|-)+]*$`.
```
jaimergp, line 49: This is probably incomplete, but it's a beginning. This does NOT standardize version _specifiers_ (e.g. `>=2.0` or `1.7.*`), just version _strings_ (`2.0.0` and `1!1.3.10post1+h12345`).

cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Package names
+
+A regular package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT start with two consecutive underscores.
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start with two underscores.
+
+This means that regular package names are limited to those which match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
+
+Virtual package names MUST follow this other regex: `^__([a-z0-9]+(\.|-|_|$))+$`.
```
beckermr, line 34: The text descriptions here are not complete. The regex also requires that no two separators appear in a row, one after the other..

cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+- Regular package names: represented by a concrete, downloadable, extractable conda artifact.
+- Virtual package names: not backed by any physical artifact. They only exist on the client side.
+
+### Package names
+
+A regular package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT start with two consecutive underscores.
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start with two underscores.
+
+This means that regular package names are limited to those which match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```
beckermr, line 32:
```suggestion
This means that regular package names MUST match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```

cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+
+Virtual package names MUST follow this other regex: `^__([a-z0-9]+(\.|-|_|$))+$`.
+
+### Version strings
+
+Version strings MUST only consist of digits, periods, lowercase ASCII letters, underscores, hyphens, plus signs and exclamation marks. Hyphens and underscores MAY be used interchangeably, but they MUST NOT be mixed in the same version string.
```
beckermr, line 38: We need to look at this carefully. Right now a lot of the conda code base uses `.rsplit("-", maxsplit=2)` to extract package name, version and build strings. Further, the VersionOrder class in conda maps all hyphens to underscores if only hyphens are in the version string. I have also never seen conda build output a package name with a dash in the version string. I think what happens is that hyphens are normalized to underscores if possible, but that hyphens are not allowed in version strings used in conda package names, the repodata, and basically any other place.

cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+Virtual conda packages SHOULD NOT need filename standardization.
+
+### Channels
+
+#### Channel names
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. In other words, they MUST match  this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, slashes, periods, colons and whitespace. They MUST start with a letter. They MUST match this regex: `^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`.
```
beckermr, line 72: The reserved `NOLABEL` label for the label empty set needs to be specified here.

beckermr (changes requested):
A backwards compatibility section needs to be added with information pulled from the OCI CEP.

beckermr (changes requested):
To be honest, I'd leave allowed versions out of this CEP all together. That topic will require a significant amount of work that will block the rest of these items which are basically good to go.

cep-9999.md
beckermr: I also noticed that trailing underscores are allowed but they are not used to generate empty version parts in the parsing algorithm.

jaimergp: Yea I was also confused by that, let's descope it out of this CEP for now.

cep-9999.md
jaimergp: Added this point to clarify, thanks! PTAL.

jaimergp:
I wonder if we should introduce maximum lengths across all categories, so we don't end up with something stupid like a package name of 50K characters.

cep-9999.md
beckermr: Well the issue is how compatible across registries we want things to be. If people are using NOLABEL as an actual label on anaconda.org, then that package could not be mirrored to an OCI registry. So I think we'd ideally want it everywhere.

beckermr:
A maximum of 128 seems good to me.

cep-9999.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Filenames
+
+Regular conda artifacts MUST be named following this scheme:
+
+```
+<package name>-<version string>-<build string>.<extension>
+```
+
+Virtual conda packages SHOULD NOT need filename standardization.
```
beckermr, line 54: This section is not 100% clear to me. I get its purpose, but the term "regular" needs more explanation. Also, I could see cases where the internals of a system use other names, but present the standardized name to users. That'd be ok to me. We're also getting close to MatchSpec territory here, which I want to avoid.

Maybe something like this is close to our intent?

```suggestion
### Canonical URIs and Filenames for conda Artifacts

The canonical URI for a conda artifact is defined as

```
<channel>(/<label>)/<subdir>/<package name>-<version string>-<build string>.<extension>
```

The optional `<label>` component MAY be percent-encoded for use as a URL. The canonical URI MUST NOT be allowed to be used in contexts where a `MatchSpec` is expected.

Virtual conda packages SHOULD NOT need URI/filename standardization.
```

cep-9999.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Channels
+
+#### Channel names
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. In other words, they MUST match  this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, slashes, periods, colons and whitespace. They MUST start with a letter. They MUST match this regex: `^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.
```
beckermr, line 64:
```suggestion
#### Label names

Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, slashes, periods, colons and whitespace. They MUST start with a letter. They MUST match this regex: `^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.

The label `NOLABEL` is reserved and MUST only be used for conda packages which have no other labels. In other words, in the space of labels, the empty set is represented by the label `NOLABEL`.
```

beckermr (changes requested):
I added a few more comments and clarifications. Thoughts @jaimergp ?

cep-9999.md
```diff
@@ -0,0 +1,78 @@
...
+
+This CEP aims to standardize names and other strings used throughout the conda ecosystem.
+
+## Specification
+
+### Artifacts
+
+The conda ecosystem distinguishes between two types of packages:
+
+- Regular package names: represented by a concrete, downloadable, extractable conda artifact.
```
beckermr, line 23: I see now where "regular" comes from. Sorry I missed that. Maybe we can pick another term? Unfortunately I don't have any great suggestions. :/

beckermr, line 23:
```suggestion
- Distributable package names: represented by a concrete, downloadable, extractable conda artifact.
```

cep-9999.md
```diff
@@ -0,0 +1,78 @@
...
+- Virtual package names: not backed by any physical artifact. They only exist on the client side.
+
+### Package names
+
+A regular package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT include two consecutive separators (hyphen, period, underscore).
```
beckermr, line 28:
```suggestion
A distributable package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT include two consecutive separators (hyphen, period, underscore). 
```

cep-9999.md
```diff
@@ -0,0 +1,78 @@
...
+- Regular package names: represented by a concrete, downloadable, extractable conda artifact.
+- Virtual package names: not backed by any physical artifact. They only exist on the client side.
+
+### Package names
+
+A regular package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT include two consecutive separators (hyphen, period, underscore). 
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST NOT use two consecutive separators, with one exception: they MUST start with two underscores.
+
+This means that regular package names MUST match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```
beckermr, line 32:
```suggestion
This means that distributable package names MUST match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```

cep-9999.md
jaimergp: Yea this is one of the areas where I want a CEP too. Standardized terminology for each thing. Package, artifact, distribution... is too loaded. Let's go for distributable package for now!

cep-9999.md
jaimergp: Not sure about mixing channel and labels here. anaconda.org does `<channel>/label/<label>/<subdir>`, but `channel` itself can have slashes (see `pkgs/main`), so it gets tricky. Also, thinking about it, we need to do something about label names that overlap with existing subdirs. Otherwise the URI is non parsable.

Also, is `<channel>` here a channel name or a full URI/URL?

beckermr: Hrmmmm hold on. Anaconda.org uses `/label/<label>`?

beckermr: Oh no you are right! I meant to match anaconda.org.

For the defaults label, is it really `pkgs/main`? I can only get repodata from this url

```
https://conda.anaconda.org/main/osx-arm64
```

These two urls give 404:

- `https://conda.anaconda.org/pkgs/main/osx-arm64`
- `https://conda.anaconda.org/pkgs%2Fmain/osx-arm64`

cep-9999.md
```diff
@@ -0,0 +1,80 @@
...
+<package name>-<version string>-<build string>.<extension>
+```
+
+Virtual conda packages SHOULD NOT need filename standardization.
+
+### Channels
+
+#### Channel names
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. In other words, they MUST match  this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
```
jaimergp, line 60: @beckermr I think we will need some more details here. This is the definition for user and organization names in anaconda.org. They _happen_ to be channel names because the conda clients interpret the path component of conda.anaconda.org URLs as channel names, but that's just a coincidence.

`conda` can happily take `-c pkgs/main` as a channel name.

How do we define a channel name then? Specially if they are not hosted in anaconda.org. I would even argue that a "channel name" includes the label because from the perspective of `conda` CLI, the name of conda.anaconda.org/conda-forge/label/rc is `conda-forge/label/rc`.

jaimergp: When local channels are used, the whole `file://` path is the channel "name". When non `channel_alias`-hosted channels are used, the whole URL is the channel "name".

beckermr: Does conda automatically use repo.anaconda.com if it sees `pkgs/main`?

cep-9999.md
beckermr: So @JeanChristopheMorinPerso what does anaconda.org actually allow for users and orgs?

cep-9999.md
jaimergp: Yep:

```
$ CONDA_CHANNELS=pkgs/main conda info

     active environment : None
       user config file : /Users/jrodriguez/.condarc
 populated config files : /Users/jrodriguez/.local/conda/.condarc
                          /Users/jrodriguez/.condarc
          conda version : 25.1.1
    conda-build version : not installed
         python version : 3.12.8.final.0
                 solver : libmamba (default)
       virtual packages : __archspec=1=m1
                          __conda=25.1.1=0
                          __osx=15.3.1=0
                          __unix=0=0
       base environment : /Users/jrodriguez/.local/conda  (writable)
      conda av data dir : /Users/jrodriguez/.local/conda/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://conda.anaconda.org/conda-forge/osx-arm64
                          https://conda.anaconda.org/conda-forge/noarch
          package cache : /Users/jrodriguez/.local/conda/pkgs
                          /Users/jrodriguez/.conda/pkgs
       envs directories : /Users/jrodriguez/.local/conda/envs
                          /Users/jrodriguez/.conda/envs
               platform : osx-arm64
             user-agent : conda/25.1.1 requests/2.32.3 CPython/3.12.8 Darwin/24.3.0 OSX/15.3.1 solver/libmamba conda-libmamba-solver/24.11.1 libmambapy/2.0.5
                UID:GID : 501:20
             netrc file : None
           offline mode : False
```

The logic is spread around the channel.py module (canonical_name and others).

cep-9999.md
beckermr: Ah I see. This URL works fine

```
https://repo.anaconda.com/pkgs%2Fmain
```

In other words, the `/` is part of the channel name.

I think the way around this is that we specify that in the canonical URI, the channel is percent-encoded except the leading bits like `https://`, etc. and includes the label possibly, per @jaimergp's points below.

This would cover all cases but I would want to adjust the OCI spec to move labels into the container `<name>` to match.

JeanChristopheMorinPerso: @beckermr Can you clarify what you mean here? I have a feeling that it was already answered in [#general > conda-oci incubation](https://conda.zulipchat.com/#narrow/channel/457607-general/topic/conda-oci.20incubation) but I might be wrong.

beckermr: My question specifically is 

What regex is used to test strings when folks try to make org or user accounts on anaconda.org?

I did not see an answer on zulip but I may have missed it. Sorry if I did!

cep-9999.md
```diff
@@ -57,17 +63,25 @@ Virtual conda packages SHOULD NOT need filename standardization.
 
 #### Channel names
 
-Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. In other words, they MUST match  this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+A channel name is defined as the
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. They MUST NOT contain two consecutive separators (hyphen, underscore, period). In other words, they MUST match this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+The maximum length of a label name MUST NOT exceed 128 characters.
```
beckermr, line 70: not sure if we can enforce this, but it should say `channel` here

beckermr:
Given the discussion above, it appears that treating the label separate from the channel is a mistake in the following sense. 

A conda `channel` formally has a single set of repodata files associated with it (the platforms plus noarch). 

In this sense, `labels` on anaconda.org are separate channels (as they have their own separate repodata). It is just that they are built out of subsets of packages from a parent channel (that subset denoted by the label). 

The notation `<channel>/label/<label>` is basically a reserved pattern in the space of channels to refer to the results of this process. 

So when we declare allowed channel names, we should not allow channel names like `conda-forge/label/main` since that would overlap with the actual `main` label on `conda-forge`. 

I think the net result of all of this is that `label` is a reserved word in the space of channel names.

cep-9999.md
```diff
@@ -0,0 +1,94 @@
...
+
+Virtual conda packages SHOULD NOT need filename standardization.
+
+### Channels
+
+#### Channel names
+
+A channel name is defined as the
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST start and end with a letter or a digit. They MUST NOT contain two consecutive separators (hyphen, underscore, period). In other words, they MUST match this regex:  `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
```
chenghlee, line 68: As of 2025-03-12T19:00Z, out of 1,918,426 channel/org/names on anaconda.org:

* 7,219 violate the regex `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`;
* 98 violate the regex `^[a-z0-9][a-z0-9_.-]*$` (allowing channel names to end with `_`, `.`, or `-`); and
* 6 violate `^[a-z0-9_][a-z0-9_.-]*$` (allowing channel names to start with `_`)

Of those six, five start with `.`, and the other starts with `~`. Which I would claim, we want to disallow in any case.

beckermr: Can you provide statistics on usage of the channels in each of those cases? 

I'd very much like to leave behind channels that start with an `_`, and I want to know how destructive that decision will be.

jjhelmus:
The requirement here on package names is more restrict than what is supported by conda and conda-build right now. Among other differences is that non-ASCII unicode character are allowed packages names in conda and can be produced by conda-build including a test which [checks this behavior](https://github.com/conda/conda-build/blob/25.1.2/tests/test-recipes/metadata/unicode_all_over/love-feedstock/meta.yaml).

If the requirements on package name are to be restricted it would be good if the CEP specified how violations are expected to be handled if at all.

jjhelmus:
There are some edge cases where it is helpful to have concrete versions of virtual testing such as testing, debugging and locking, especially on non-native systems. For example conda-lock [creates a local channel](https://github.com/conda/conda-lock/blob/main/conda_lock/virtual_package.py) with realized virtual packages. 

I don't believe this CEP prohibits the existence of these artifacts but it would be reasonable to include a note to this effect. That said concrete packages that have virtual packages name should not appear in general purpose channels.

jjhelmus:
> Yep. Correct @jjhelmus. I've checked defaults and conda-forge for violations. IMHO tools should simply fail.

Should tools be actively checking for compatibility and if so at what level; creation, repodata, download/install? 

As a concrete example should a ♥ package be buildable? included in repodata? installable?  

For reference right now the ♥ package can be: 
* built by conda-build
* indexed by conda-index
* installed by conda using the classic solver

The following fail with the package:
* Uploading to anaconda.org (`[ERROR] ('package name ♥ not valid', 400)`)
* Installing with the libmamba solver (`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe2 in position 45: invalid continuation byte`, could be an issue on my end)

Another option is to leave the behavior of these packages undefined, which IMHO is okay.

cep-9999.md
```diff
@@ -0,0 +1,94 @@
...
+- Distributable package names: represented by a concrete, downloadable, extractable conda artifact.
+- Virtual package names: not backed by any physical artifact. They only exist on the client side.
+
+### Package names
+
+A distributable package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT include two consecutive separators (hyphen, period, underscore).
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and underscores. They MUST NOT use two consecutive separators, with one exception: they MUST start with two underscores.
+
+This means that distributable package names MUST match the following case-insensitive regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```
chenghlee, line 32: As of 2025-03-12T19:00Z, of the ~209K unique distributeable [conda] package names on anaconda.org, only 3 violate the `^[a-z0-9_][a-z0-9_.-]*$` naming pattern:
* one is the `__anaconda_core_depends` package Anaconda is working on renaming;
* one is the name of an existing virtual package under a "hack" label; and
* one contains a space but is a broken artifact with zero downloads

(I'm still looking for package names with consecutive separators; the database I'm querying doesn't support backreferences in its regexps.)

cep-9999.md
```diff
@@ -61,27 +61,29 @@ Virtual conda packages SHOULD NOT need filename standardization.
 
 ### Channels
 
+A conda channel is defined as a location where one can find one or more `repodata.json` files arranged in one subdirectory (_subdir_) each. `noarch/repodata.json` MUST be present to consider the parent location a channel.
```
beckermr, line 64: Ah sorry. This was my way of thinking about channels. For this CEP, we only want to codify the naming conventions and leave what is in the channel to another CEP.

beckermr: The other option is to propose this as a way to get to a working definition of what should be allowed for a channel name. For example, we say "In anaconda.org and conda/mamba/rattler, a channel is defined in practice as a location where..."

This leaves room for future specs and limits scope.

jaimergp:
> Given the discussion above, it appears that treating the label separate from the channel is a mistake in the following sense.
> 
> A conda `channel` formally has a single set of repodata files associated with it (the platforms plus noarch).
> 
> In this sense, `labels` on anaconda.org are separate channels (as they have their own separate repodata). It is just that they are built out of subsets of packages from a parent channel (that subset denoted by the label).
> 
> The notation `<channel>/label/<label>` is basically a reserved pattern in the space of channels to refer to the results of this process.
> 
> So when we declare allowed channel names, we should not allow channel names like `conda-forge/label/main` since that would overlap with the actual `main` label on `conda-forge`.
> 
> I think the net result of all of this is that `label` is a reserved word in the space of channel names.

Hm, maybe we can just consider this an implementation detail of anaconda.org. "Labels" are just a way of creating repodata.json files from a collection of conda artifacts. By default, `main` renders the repodata in the user/org channel. Changing that default label renders the repodata in `user_or_org/label/the_new_label`. But it could have been very well `user_or_org/the_new_label`. anaconda.org also supports label-less packages which go nowhere, so there's no need to establish that a label is always a subset of main, because something under a label doesn't have to be in main or anywhere really.

This begs the question of whether we want to standardize labels at all, or simply define an equivalent mechanism in the OCI mirror. Labels control repodata indexing by creating a different channel location for the selected packages. The labels can be applied after upload and publication. Since it concerns repodata generation, this has not been covered by the OCI CEP. We only need to figure out an OCI-equivalent way of doing this (one way could be ["attached artifacts" via the referrers API](https://oras.land/docs/concepts/reftypes/))

In that regard, we don't need to standardize label names separately; only as part of what a channel name is. To that effect, I've added a short paragraph now.

cep-9999.md
jaimergp: But I don't think we can name a channel without defining what it is. I like your definition a lot. We can add it as a contextual note if you don't want this CEP to be the authority for that definition, but this is the problem with all these CEPs: we are building on top of shaky foundations all the time.

If we extend the definition to "path component of URI before subdir", then we discover we _need_ to include slashes. If you want to only standardize the last path component before the subdir, then that's a different regex, without the slashes.

In OCI, we can have as many path components as we want, because we will parse things form the right: `channel/subdir/package-name`. `channel` there can be anything, including `my/long/path/to/channel/name` and `conda-forge/label/rc`.

cep-9999.md
chenghlee: > Can you provide statistics on usage of the channels in each of those cases?

Will likely take me some time to dig into that. Since channel names and user names are roughly equivalent .org, it's not the easiest task to separate "logged traffic due to using this as a conda channel" from "logged traffic because someone logged in and did soemthing else".

beckermr:
Yeah I am not so sure. The eventual CEP specifying what standard HTTP-based conda server looks like will want a way to specify that subsets of packages on a channel should be grouped in different ways. This feature is pretty essential to package maintenance. 

We don't have to specify how repodata is generated from labels in this CEP, but I think we should reserve the namespace.

cep-9999.md
beckermr: Right I see your point. The distinction really is about the physical thing ("finding a file") versus a set of rules for specifying how to find information ("how do I construct a string to feed to conda to have it do a thing?"). In that sense, a conda channel is more akin to specification built on top of a URI. 

This CEP should specify rules of how to build the URI.

chenghlee:
> Oops, `label` is a valid channel name 😂 https://anaconda.org/label

As long as we specifiy the construct is "the channel name, followed by the string literal 'label', followed by the label name", I'm not sure that's actually a problem?

jaimergp:
The problem I see is with URLs like `my.server.org/label/label/linux-64/linux-64/repodata.json`. Is the channel name `label`, with label `linux-64` and subdir `linux-64`? Is the channel name `linux-64`, and happens to be in a path that has two `label` components before it?

chenghlee:
> We can specify parsing from right to left?
> ...

Give me a bit. I _think_ I can write an EBNF grammar that will capture that.

jaimergp:
Then, for `my.server.org/label/linux-64/linux-64/repodata.json` (one less `label` path component), the server name is `my.server.org`, `` (empty), or `label/linux-64`?

beckermr:
By the logic above, this string

> my.server.org/label/linux-64/linux-64/repodata.json

would be parsed as

0. remove the repodata.json
1. subdir = linux-64
2. Splitting `my.server.org/label/linux-64/` on the first label from the right produces (`my.server.org`, `linux-64`). So the channel is `my.server.org` and the label is `linux-64`.

I don't see why `my.server.org` can't be a valid channel?

chenghlee:
> I don't see why `my.server.org` can't be a valid channel?

Just tried `conda search --override-channels -c http://localhost:8080 openblas`, and that worked just fine, so I'd argue that `https://my.server.org` should be a perfectly valid channel.

What might break, though, is `https://my.server.org/label/linux-64/repodata.json`. Maybe. But, in that case, since `label` is not preceded by [another] `label`, I suppose the channel name is `label`, no label, and subdir name is `linux-64`.

beckermr:
Yep @chenghlee. If you split on the `label` and the label you extract is the empty string, then there is no label.

beckermr:
Just to be very clear, in this spec the channel should include everything to the start of the URI. conda itself can strip the anaconda.org out of that bit for display purposes, but this is a UX feature.

beckermr:
Right. We can fix that in the cli. That is a display issue.

jaimergp:
> Just to be very clear, in this spec the channel should include everything to the start of the URI.

Including or excluding the scheme? I assume no authentication, but maybe we do have encode the port? That means we need to accept slashes and colons too.

cep-9999.md
jaimergp: I think we have conflated "name", "display name" and "location" while ironing these out. For clarity, my current understanding is that we want to specify how to build a URI to determine a channel location. "name" and "display name" are irrelevant here. Is that accurate?

jaimergp:
That's what I think too, but then mapping that to OCI will present more challenges 😬

chenghlee:
> Just to be very clear, in this spec the channel should include everything to the start of the URI.

Modulo the arguably pedantic distinctions between URIs ([RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)) and URLs ([RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738)), I would agree with that. To support right to left parsing in any sensible way, I think "channel" should include everything else in the URL &mdash; scheme, host, port, user, password, etc. included.

> That's what I think too, but then mapping that to OCI will present more challenges 😬

True; and I think there's maybe to definitely a separate discussion of what makes a "channel" unique. It's not just the OCI mapping that creates such semantic issues; e.g., (`cf-staging` -> `conda-forge`) and mirrors in various forms would have the same problems. To a large extent, I think this comes from the fact that conda packages are far more mobile across "channels" (whatever that may mean) than packages in other ecosystems.

cep-9999.md
beckermr: That's my opinion. Once we know what a conda uri looks like, we can specify how an OCI registry would fit into that scheme. OCI Registries already allow ports, but we'll need to encode some characters to match the OCI spec.

beckermr:
I need to read those rfcs. I don't know the difference myself.

beckermr:
I think we can defer some of the mapping issues to a spec on mirroring and a global index of channels. 

I should reframe the OCI cep to simply specify what an OCI uri looks like, how it gets encoded to be used with an OCI instance, and not force it to be tied to specific things in other

jaimergp:
The URL/URI RFCs plus WHATWG URLs is a [fun read](https://github.com/whatwg/url/issues/703) 😂

cep-9999.md
jaimergp: Good. I'll reword this part tomorrow then to accommodate the feedback so far.

chenghlee:
> I need to read those rfcs. I don't know the difference myself.

90% of the time, I forget what the exact distinctions are. But I'm also aware that someone taking a cursory glance at [the example URIs](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3) in RFC 3986 could pedantically argue that `tel:+1-816-555-1212` should be a valid channel, and none of us have the time, energy, or blood pressure medication to argue "absolutely not". :laughing:

cep-9999.md
```diff
@@ -0,0 +1,96 @@
...
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods, slashes and underscores. They MUST start and end with a letter or a digit. They MUST NOT contain two consecutive separators (hyphen, underscore, period, slash). In other words, they MUST match this regex: `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+The maximum length of a channel name SHOULD NOT exceed 128 characters.
+
+#### Subdir names
+
+Channel subdir names MUST only consist of lowercase ASCII letters, numbers, hyphens, underscores and periods. They MUST start and end with a letter or a number. They MUST NOT contain two consecutive separators (hyphen, underscore, period). They MUST match this regex: `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+The maximum length of a subdir name MUST NOT exceed 32 characters.
```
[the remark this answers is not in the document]
jaimergp: I was even thinking that we don't need to be super lenient with subdirs. Basically `^[a-z0-9]+(-[a-z0-9]+)*$` is enough, right? This way files with extensions are out of the picture.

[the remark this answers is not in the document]
beckermr: Ah yes we can do that too!

cep-9999.md
```diff
@@ -0,0 +1,96 @@
...
+
+Channel subdir names MUST only consist of lowercase ASCII letters, numbers, and hyphens. They MUST start and end with a letter or a number. They MUST NOT contain two consecutive hyphens. They MUST match this regex: `^[a-z0-9]+(-[a-z0-9]+)*$`.
+
+The maximum length of a subdir name MUST NOT exceed 32 characters.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward slashes, periods, colons and whitespace. They MUST start with a letter. They MUST match this regex: `^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.
+
+The label `NOLABEL` is reserved and MUST only be used for conda packages which have no other labels. In other words, in the space of labels, the empty set is represented by the label `NOLABEL`.
```
beckermr, line 84:
```suggestion
The labels `nolabel` is reserved and MUST only be used for conda packages which have no other labels. In other words, in the space of labels, the empty set is represented by the labels `nolabel`.
```

cep-9999.md
```diff
@@ -0,0 +1,96 @@
...
+
+Version strings MUST only consist of digits, periods, lowercase ASCII letters, underscores, plus
+signs and exclamation marks. Additional rules apply but are out of scope in this CEP and will be
+discussed separately.
+
+The maximum length of a version string MUST NOT exceed 128 characters.
+
+### Build strings
+
+Builds strings MUST only consist of ASCII letters, numbers, periods, and underscores. They MUST match this regex `^[a-zA-Z0-9_\.]+$`.
```
beckermr, line 48:
```suggestion
Builds strings MUST only consist of ASCII letters, numbers, periods, and underscores. They MUST match this regex `^[a-zA-Z0-9_\.+]+$`.
```

We have a few build strings in conda-forge with `+` in them. Is there a good reason to not allow `+` @jaimergp?

jaimergp: Nah, just an oversight. Didn't see any pluses in my local `noarch` + `osx-arm64` repodata caches.

cep-9999.md
beckermr: Ahhhh this is in conda too: https://github.com/conda/conda/blob/4.8.3/conda/gateways/connection/adapters/s3.py

beckermr: Looks like we can pursue a pretty general approach with a set of allowed schemes, the authority section, etc 

```
<scheme>://<authority>/<channel path parts>[/label/<label parts>]/<subdir>/...
```

xhochy: > @jaimergp I reworded the OCI CEP to be focused around valid URLs for OCI channels. I wonder if that is an appropriate model for rewording this CEP.
> 
> I've heard that @xhochy and @wolfv have s3 channels working and I am wondering if the implementation there can help guide us to write down the correct URL.

Paging @delsner and @pavelzw who work on S3 channels.

pavelzw: right now, you can also specify channels like this in rattler/pixi/rattler-build:

```
s3://<bucket-name>/<channel-path-parts>/<subdir>/...
```

the channel-path-parts could be empty, though, for example `s3://conda-forge` would search for the S3 bucket called `conda-forge` and in it the files `/noarch/repodata.json`, ...

you could map the authority to bucket name theoretically.

---

also, in S3, you can specify different endpoint URLs/regions.
rattler uses [aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust/) and conda uses [boto3](https://github.com/boto/boto3).
both can be configured through environment variables like `AWS_REGION`, `AWS_CONFIG_FILE`, ... to further specify where these buckets are hosted, for example if you want to use cloudflare or minio.

rattler and pixi both also expose a custom configuration if you don't want to rely on the aws crate to handle this through env variables: https://pixi.sh/latest/advanced/s3/#using-pixis-configuration

```toml
[project.s3-options.my-bucket]
endpoint-url = "https://my-s3-host"
region = "us-east-1"
force-path-style = false
```

cep-9999.md
```diff
@@ -0,0 +1,96 @@
...
+
+Channel names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods, slashes and underscores. They MUST start and end with a letter or a digit. They MUST NOT contain two consecutive separators (hyphen, underscore, period, slash). In other words, they MUST match this regex: `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`.
+
+The maximum length of a channel name SHOULD NOT exceed 128 characters.
+
+#### Subdir names
+
+Channel subdir names MUST only consist of lowercase ASCII letters, numbers, and hyphens. They MUST start and end with a letter or a number. They MUST NOT contain two consecutive hyphens. They MUST match this regex: `^[a-z0-9]+(-[a-z0-9]+)*$`.
```
chenghlee, line 76: I think we should constrain this even more and say the subdir name must either:

* Be the string literal `noarch`; or
* Follow the pattern `{os}-{architecture}`, where each of `{os}` and `{architecture}` MUST contain only lowercase ASCII letters or numbers (`[a-z0-9]`).

I'm not sure we gain anything by allowing more than one hyphen in the subdir name; the regexp you have here (`^[a-z0-9]+(-[a-z0-9]+)*$`) allows, e.g., `linux-or-vms-on-x86-64-or-vax` as a subdir name.

chenghlee, line 76: I think we should constrain this even more and say the subdir name must either:

* Be the string literal `noarch`; or
* Follow the pattern `{os}-{architecture}`, where each of `{os}` and `{architecture}` MUST contain only lowercase ASCII letters or numbers (`[a-z0-9]`).

I'm not sure we gain anything by allowing more than one hyphen in the subdir name; the regexp that's currently here (`^[a-z0-9]+(-[a-z0-9]+)*$`) allows, e.g., `linux-or-vms-on-x86-64-or-vax` as a subdir name.

beckermr: Maybe we specify a specific allow-list stored at https://schemas.conda.org/ as json or something?

jaimergp:
Reworded the channel section following the approach in your OCI CEP @beckermr. Also addressed your comment about subdirs @chenghlee.

cep-XXXX.md
```diff
@@ -0,0 +1,119 @@
...
+to `noarch/repodata.json` and thus base URL is `https://conda.anaconda.org/conda-forge/conda-forge`. For local repodata such as `file:///home/username/channel/noarch/repodata.json`, the
+channel base URL is `file:///home/username/channel`.
+
+For convenience, the channel _name_ is defined as the concatenation of `scheme`, `authority` and
+`path` components. `scheme` MAY be omitted for `http`, `https` and `file`. At least one of
+`authority` or `path` SHOULD be present. In their absence, the channel name MUST be considered
+empty, regardless the scheme. Empty channel names SHOULD NOT be used.
+
+When present, each path component SHOULD only contain lowercase ASCII letters, numbers, underscores
+and dashes. They SHOULD start and end with a letter or a number.
```
beckermr, line 93: Is there a reason we do not allow consecutive separators at all in conda package names, but we do allow them in channel path components?

cep-XXXX.md
```diff
@@ -0,0 +1,119 @@
...
+
+The label `nolabel` is reserved and MUST only be used for conda packages which have no other labels. In other words, in the space of labels, the empty set is represented by the labels `nolabel`.
+
+The maximum length of a label name SHOULD NOT exceed 128 characters.
+
+## Backwards compatibility
+
+The conda channel, subdir, label, and package name regexes are backwards compatible with the current `conda` implementation (25.3) and all existing packages on the `defaults` and `conda-forge` channels, except the `__anaconda_core_depends` package on the `defaults` channel.
+
+The regex for labels above was pulled from an anaconda.org error message describing the set of valid labels. Finally, anaconda.org usernames/channel names are case insensitive and cannot begin with an underscore.
```
beckermr, line 115: It would be good to collect the stats from @chenghlee on anaconda.org and put them as text here.

cep-XXXX.md
beckermr: if we want to cater to OCI, then this will need to be restricted to the OCI name regex.

beckermr: The regex needed by OCI for labels and channel path components is

```
^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$
```

which is the `v1.*` OCI regex for container `<name>`s. 

I do not have stats on how many labels violate this on anaconda.org. The number of channels from usernames/orgs that violate it is around 0.4% which is small enough that I excluded them in the OCI CEP. 

Maybe we simply exclude them here too?

beckermr: Also this is the stat from @chenghlee 

> As of 2025-03-12T19:00Z, of the ~1.9M channel/org/names on anaconda.org:
> 
> 7,219 violate the regex ^[a-z0-9]+((-|_|.)[a-z0-9]+)*$;
> 98 violate the regex ^[a-z0-9][a-z0-9_.-]*$ (allowing channel names to end with _, ., or -); and
> 6 violate ^[a-z0-9_][a-z0-9_.-]*$ (allowing channel names to start with _)
> Of those six, five start with ., and the other starts with ~. Which I would claim, we want to disallow in any case (since they correspond to hidden files, directory traversal, or a user home directory in *NIX).
> 
> Edit: With the caveat that I used case-insensitive matching when evaluating the above. But I'm okay with requiring conda-compatible software to normalize channel names by lower-casing.

So the requirement that things start and end with a letter already excludes the 0.4% regardless of the issue of how many separators appear in a row. Thus we probably should just use the OCI name regex and be done with it?

jaimergp: Ok, I can add this as a SHOULD for now. Or do you want it as MUST?

beckermr: Looking at this more, I think we should go all one way (very liberal) or the other (restrict to OCI). In between the those cases doesn't make much sense to me. 

So I think we specify that channel path components MUST match `^[a-z0-9_][[a-z0-9_.-]$` and let the OCI spec exclude more.

cep-XXXX.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+Virtual conda packages do not exist on disk and SHOULD NOT need filename standardization.
+
+#### Distribution strings
+
+A "distribution string" MAY be used to identify a package artifact without specifying the extension or the channel. It MUST match the following syntax:
+
+```text
+<package name>-<version string>-<build string>
```
beckermr, line 75: We need to use space-separated and/or `=`-separated forms of this as valid matchspecs in recipes. We may want to be clearer about the scope of this section or add a carve out for matchspecs

jaimergp: Yea this is not a MatchSpec, is a dist str, which is not allowed anywhere, just a convention like filenames are. I'll clarify.

cep-XXXX.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+For example, given `https://conda.anaconda.org/conda-forge/noarch/repodata.json`, the part leading
+to `noarch/repodata.json` and thus base URL is `https://conda.anaconda.org/conda-forge/conda-forge`. For local repodata such as `file:///home/username/channel/noarch/repodata.json`, the
+channel base URL is `file:///home/username/channel`.
+
+For convenience, the channel _name_ is defined as the concatenation of `scheme`, `authority` and
+`path` components. `scheme` MAY be omitted for `http`, `https` and `file`. At least one of
+`authority` or `path` SHOULD be present. In their absence, the channel name MUST be considered
+empty, regardless the scheme. Empty channel names SHOULD NOT be used.
```
beckermr, line 108: I may have made this mistake in the OCI CEP too. What you mean when you say the scheme MAY be omitted is that we allow things like `conda-forge`. Literally removing the scheme above would look like `://conda-forge` which is not what we mean.

jaimergp: Oops, good point. I'll add a note.

cep-XXXX.md
```diff
@@ -0,0 +1,153 @@
...
+
+The regex for labels was pulled from an anaconda.org error message describing the set of valid labels.
+
+As of 2025-03-12T19:00Z, of the ~1.9M channel names on anaconda.org:
+
+- 7,219 violate the regex `^[a-z0-9]+((-|_|.)[a-z0-9]+)*$`;
+- 98 violate the regex `^[a-z0-9][a-z0-9_.-]*$` (allowing channel names to end with `_`, `.`, or `-`); and
+- 6 violate `^[a-z0-9_][a-z0-9_.-]*$` (allowing channel names to start with `_`). Of those six, five start with `.`, and the other starts with `~`.
+
+Given the low percentage (~0.4%) of mismatches, authors have decided to ignore those entries and err on the side of strictness to maximize compatibility with other storage solutions that have more restrictive naming requirements.
```
beckermr, line 145: May not be true given our discussion above.

cep-XXXX.md
jaimergp: Works for me.

cep-XXXX.md
```diff
@@ -0,0 +1,156 @@
...
```
beckermr, line 148:
```suggestion
The authors have excluded the channel names in the last case that start with `.` or `~` given possible security implications. A low percentage, ~0.4%, of channels do not match the recommendations for channel names above, but are allowed.
```

jaimergp:
> Some of limits appear arbitrary or not well-motivated. 

Yes they are pretty much arbitrary 😂 Wanted to start the discussion somewhere. I think we should base these thresholds on existing data. I'd like to know the longest

* channel name in anaconda.org and prefix.dev
* label name in anaconda.org
* package name in anaconda.org and prefix.dev
* version string
* build string

Maybe we already have these numbers? I recall asking for these but not sure if we ever got an answer back.

With those we can more or less see where the community is at after ~12 years of conda, which should be sufficient sampling.

beckermr:
@chenghlee can you pull out these maximum lengths from anaconda.org?

beckermr:
@chenghlee can you get these stats for us or should ask someone else?

chenghlee:
> Considering only conda packages and channels that have at least one published conda package:
> ...

Mind you, that analysis did not account for when said packages were created nor when they were last downloaded. As a conda steering council member, I'm not exactly opposed to setting [new] limits that would invalidate some existing packages, especially when it comes to versions and build string.

jaimergp:
One more argument for a conservative 64-char limit: some clients might add an extra extension for downloads in progress like `.partial`, or lock the file with `.lock`, etc; so it's better to have some extra space there.

beckermr:
We should check defaults, bioconda, and conda-forge specifically for the limits. I think all of those packages are below the limits given my OCI analysis for defaults and conda-forge. I did not check bioconda.

beckermr (approved):

beckermr:
pre-commit.ci autofix

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+The authors have excluded the channel names in the last case that start with `.` or `~` given possible
+security implications. A low percentage, ~0.4%, of channels do not match the recommendations for channel
+names above, but are allowed.
+
+The maximum lengths allowed for the different fields have been chosen so the resulting path components
+(directory names, filenames) comfortably fit in a the 255-char maximum limit some filesystems impose.
+As of 2025-03-01T13:00Z, there are no violations of these limits in any of the packages published for
+`conda-forge`, `bioconda` and `defaults`. See
+[this comment](https://github.com/conda/ceps/pull/116#issuecomment-2763392999) and
+[this comment](https://github.com/conda/ceps/pull/116#issuecomment-2759130187)for more details.
```
beckermr, line 193:
```suggestion
[this comment](https://github.com/conda/ceps/pull/116#issuecomment-2759130187) for more details.
```

beckermr:
This PR is ready for a vote I think. @jaimergp shall we go ahead?

jaimergp:
> This PR is ready for a vote I think. @jaimergp shall we go ahead?

Let's leave a period for comments until next Wednesday, so we can close the two open ones, and start this one. WDYT?

beckermr:
We should mark it ready for review and bump steering.

jaimergp:
@conda/steering-council, we'd like to call a vote for this starting next Wednesday. We'll wait until then to receive feedback and comments.

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
+# CEP XXXX - Names in conda packages and channels
+
+<table>
+<tr><td> Title </td><td> CEP XXXX - Names in conda packages and channels </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt; <br /> Matthew R. Becker &lt;becker.mr@gmail.com&gt;</td></tr>
```
jaimergp, line 6: @chenghlee, would you like to be part of the author list? You've helped a lot!

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+For example, given `https://conda.anaconda.org/conda-forge/noarch/repodata.json`, the part leading
+to `noarch/repodata.json` and thus base URL is `https://conda.anaconda.org/conda-forge/conda-forge`. For
+local repodata such as `file:///home/username/channel/noarch/repodata.json`, the
+channel base URL is `file:///home/username/channel`.
+
+For convenience, the channel _name_ is defined as the concatenation of `scheme`, `authority` and
+`path` components. The `<scheme>://` part MAY be omitted if `scheme` is one of `http`, `https` or
+`file`. At least one of `authority` or `path` SHOULD be present. In their absence, the channel name
```
chenghlee, line 134: > The `<scheme>://` part MAY be omitted if `scheme` is one of `http`, `https` or
`file`.

Should we restrict this to just `file`? It seems odd to me that if I do something like `mkdir -pv conda.anaconda.org/mychannel/noarch`, multiple conda-compatible clients now have to figure out and agree if the schema-less argument `--channel conda.anaconda.org/mychannel/` refers to a local relative (`file://`-like) path or a HTTP(S) service.

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+MUST be considered empty, regardless the scheme. Empty channel names SHOULD NOT be used.
+
+When present, each path component MUST only contain lowercase ASCII letters, numbers, underscores,
+periods, and dashes. They MUST NOT start with a period or a dash. They SHOULD start and end with a
+letter or a number. If present, each path component MUST match this regex:
+
+```re
+^[a-z0-9_][a-z0-9_.-]*$
+```
```
chenghlee, line 143: This might create a potential problem with `file://` URLs, especially when it comes to user home directories.

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+Channel subdir names MUST either be the literal `noarch` or a string following the syntax `{os}-{arch}`,
+where `{os}` and `{arch}` MUST only consist of lowercase ASCII letters and numbers. Non-`noarch` subdirs
+MUST match this regex: `^[a-z0-9]+-[a-z0-9]+$`.
+
+The maximum length of a subdir name MUST NOT exceed 32 characters.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward slashes,
+periods, colons and whitespace. They MUST start with a letter. They MUST match this regex:
```
chenghlee, line 159: I'm inclined to disallow colons, since they're disallowed on CIFS and Windows (except for drive letters) and get interpreted as directory delimiters on macOS.

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+`^[a-zA-Z][0-9a-zA-Z_\-\.\/:\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.
+
+The label `nolabel` is reserved and MUST only be used for conda packages which have no other labels.
+In other words, in the space of labels, the empty set is represented by the labels `nolabel`.
+
```
beckermr, line 164:
```suggestion

A URL for a package, repodata, etc. without a label component MUST be assumed to have the default label `main`.

```

cep-XXXX.md
```diff
@@ -0,0 +1,201 @@
...
+3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2).
+
+Taken the channel definition above, the base URL without trailing slashes is thus:
+
+```text
+<scheme>://[<authority>][/<path>/][/label/<label name>]
+```
+
+For example, given `https://conda.anaconda.org/conda-forge/noarch/repodata.json`, the part leading
+to `noarch/repodata.json` and thus base URL is `https://conda.anaconda.org/conda-forge/conda-forge`. For
```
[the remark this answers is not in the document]
jaimergp: Oops, yes, this is a typo.

cep-XXXX.md
jaimergp: I'll add an exception for `file://` channels, downgrading it to a SHOULD.

cep-XXXX.md
jaimergp: Applied suggestions.

cep-XXXX.md
jaimergp: Added a new section for this only. Let me know what you think.

beckermr: Looks good. @chenghlee @jaimergp Do we know what rules `conda` uses when doing this type of inference? Do they match what is in the CEP? The rules don't have to match, but it'd be good to compare so we know what is currently done.

cep-XXXX.md
```diff
@@ -0,0 +1,215 @@
...
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward slashes,
+periods, and whitespace. They MUST start with a letter. They MUST match this regex:
+`^[a-zA-Z][0-9a-zA-Z_\-\.\/\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.
+
+The label `nolabel` is reserved and MUST only be used for conda packages which have no other labels.
+In other words, in the space of labels, the empty set is represented by the labels `nolabel`.
+
+A URL for a package, repodata, etc. without a label component MUST be assumed to have the default label `main`.
+The maximum length of a label name MUST NOT exceed 128 characters.
```
beckermr, line 179:
```suggestion

The maximum length of a label name MUST NOT exceed 128 characters.
```

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+where `{os}` and `{arch}` MUST only consist of lowercase ASCII letters and numbers. Non-`noarch` subdirs
+MUST match this regex: `^[a-z0-9]+-[a-z0-9]+$`.
+
+The maximum length of a subdir name MUST NOT exceed 32 characters.
+
+#### Label names
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward slashes,
+periods, and whitespace. They MUST start with a letter. They MUST match this regex:
+`^[a-zA-Z][0-9a-zA-Z_\-\.\/\s]*`. Even if allowed, label names SHOULD NOT contain any whitespace.
```
beckermr, line 173: @chenghlee @jaimergp as long as we're deviating from anaconda.org here (with the removal of `:`s), we might as well remove whitespace `\s` as well?

cep-XXXX.md
jaimergp: Not really, I think.

For environment names (not channels), there are two CLI parts:

- `conda create` disambiguates via `-n` (name) and `-p` (path, prefix).
- `conda activate` does not, and instead relies on [this logic](https://github.com/conda/conda/blob/866f84041cb40a4122fe5e8154d67eb28817884a/conda/activate.py#L344-L353). This is similar to what we propose here.

For channels, it's a bit more [involved](https://github.com/conda/conda/blob/866f84041cb40a4122fe5e8154d67eb28817884a/conda/models/channel.py#L442-L472). They strip path components until there's a match in `custom_channels`, or give up and append it to `channel_alias` (conda.anaconda.org).

cep-XXXX.md
jaimergp: I'm ok with that.

jaimergp: I can also add back `:` and include it in the SHOULD NOT part.

cep-XXXX.md
beckermr: ok LGTM.

cep-XXXX.md
beckermr: I think both `:` and `\s` are bad ideas so I am happy removing them both.

jaimergp: Farewell to \s then!

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
+# CEP XXXX - Names in conda packages and channels
+
+<table>
+<tr><td> Title </td><td> CEP XXXX - Names in conda packages and channels </td>
```
beckermr, line 4:
```suggestion
<tr><td> Title </td><td> CEP XXXX - Conda Channel URLs </td>
```

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+<tr><td> Created </td><td> Mar 11, 2025</td></tr>
+<tr><td> Updated </td><td> Apr 1, 2025</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/116 </td></tr>
+<tr><td> Implementation </td><td> N/A </td></tr>
+</table>
+
+## Abstract
+
+This CEP aims to standardize names and other strings used to identify packages, artifacts and
+channels in the conda ecosystem.
```
beckermr, line 20:
```suggestion
This CEP aims to standardize conda channel URLs, package names, and other strings used to identify packages, artifacts and channels in the conda ecosystem.
```

beckermr (commented):
Shall we retitle the CEP to reflect where we ended up?

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the
+rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the
+implementation.
+
+### Identifying package artifacts
+
+The conda ecosystem distinguishes between two types of packages:
+
+- Distributable package names: represented by a concrete, downloadable, extractable conda artifact.
+- Virtual package names: not backed by any physical artifact. They only exist on the client side.
```
trallard, line 37:
```suggestion
- Virtual package names: not backed by any concrete artifact. They only exist on the client side.
```

Align for consistency across definitions

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+A distributable package name MUST only consist of lowercase ASCII letters, numbers, hyphens, periods
+and underscores. It MUST start with a letter, a number, or a single underscore. It MUST NOT include
+two consecutive separators (hyphen, period, underscore).
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and
+underscores. They MUST NOT use two consecutive separators, with one exception: they MUST start with
+two underscores.
+
+This means that distributable package names MUST match the following case-insensitive
+regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```
trallard, line 50: Nit: let's move this before `Virtual packages` (line no. 45) for readability too

```suggestion
Distributable package names MUST match the following case-insensitive
regex: `^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+
+#### Filenames
+
+Distributable conda artifacts MUST have a filename following this scheme:
+
+```text
+<package name>-<version string>-<build string>.<extension>
+```
+
+The maximum length of a filename MUST NOT exceed 192 characters.
```
trallard, line 90:
```suggestion
The maximum length of a filename MUST NOT exceed 211 characters.
```

cep-XXXX.md
```diff
@@ -0,0 +1,216 @@
...
+
+```text
+<package name>-<version string>-<build string>
+```
+
+Distribution strings apply to both distributable and virtual packages. They are used as the name of the
+directories where artifacts are extracted in the package cache, for example.
+
+> Note: Despite the similarity, distribution strings are not `MatchSpec`-like specifiers and
+> MUST NOT be used as such.
```
trallard, line 107: This would be best moved _before_ filenames as the latter inherit from the `Distribution strings` spec

jjhelmus:
Is there interest in adding information about preferred package names, as in what separator to prefer when multiple are possible? For example, `importlib-metadata`, `importlib_metadata`, `importlib.metadata`. Perhaps recommending, but not requiring, that the names match the [normalized name](https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization) used in Python packaging (`importlib-metadata`). If this was widely followed, matching packages between the two ecosystems would be a bit easier.

cep-XXXX.md
```diff
@@ -0,0 +1,226 @@
+# CEP XXXX - Names in conda packages and channels
```
beckermr, line 1:
```suggestion
# CEP XXXX - Identifying Packages and Channels in the conda Ecosystem
```

beckermr:
We have some draft CEPs on PURLs and other cross-ecosystem issues. I think it makes sense to put that info there or generalize those CEPS?

jaimergp:
> Is there interest in adding information about preferred package names, as in what separator to prefer when multiple are possible?

I feel that's very ecosystem dependent. What works on PyPI might not be the same thing for CPAN, R, NPM or others. Also, if we decided to go for it and offer recommendations, I'd rather have that as a separate PR because it feels it would be a bit bikesheddy and I want to submit the vote for this one so we can iterate on the OCI CEP and other efforts.

cep-XXXX.md
```diff
@@ -0,0 +1,226 @@
...
+their absence, the channel name MUST be considered empty, regardless the scheme. Empty channel
+names SHOULD NOT be used.
+
+When the scheme and authority fields are missing, the full URL can be inferred with these rules:
+
+- If the channel name matches the regex `^\.{0,2}[/\\].*$`, or if it matches the regex
+  `^[A-Z]:([\\/].*)?$` (for Windows drives), it SHOULD be understood as the path component of a
+  `file://` URL.
+- Otherwise, it SHOULD be understood as a `http[s]://` URL. The tool SHOULD assume a default scheme
+  and authority (e.g. `https://conda.anaconda.org`), and take the rest as a path component.
```
chenghlee, line 165: How about
> Otherwise, the tool SHOULD provide a user-configurable mechanism to prepend a default scheme and authority, with provided channel name taken as the rest of the path component. At the time of this CEP's writing, most tool assume the default URL scheme and authority to be `https://conda.anaconda.org`.

I think this better captures the intent (i.e., support the `-c`/`--channel` option as they exist today), while recognizing that condarc already provides mechanisms to have the default schema not be `http[s]://` for certain channels.

cep-XXXX.md
```diff
@@ -0,0 +1,228 @@
...
+> The conda ecosystem currently recognizes two artifact extensions: `tar.bz2` and `conda`,
+versioned `v1` and `v2` respectively.
+
+#### Distribution strings
+
+A "distribution string" MAY be used to identify a package artifact without specifying the extension
+or the channel. It MUST match the following syntax:
+
+```text
+<package name>-<version string>-<build string>
```
beckermr, line 88: We may want to add a possible `<subdir>` part here too?

```suggestion
[<subdir>/]<package name>-<version string>-<build string>
```

cc @jaimergp

jaimergp: Good point. Let me rephrase the filename bit too so it doesn't add confusion, though.

cep-XXXX.md
```diff
@@ -0,0 +1,232 @@
...
+^[a-z0-9_][a-z0-9_.-]*$
+```
+
+For `file://`-based channel URLs, the path component rules MAY be understood as recommendations
+only.
+
+The maximum length of an individual path component in a channel base URL MUST NOT exceed 128
+characters. The maximum length of a channel base URL SHOULD NOT exceed 256 characters.
+
+To avoid ambiguous `MatchSpec` grammar, the last path component of a channel base URL MUST NOT match any `subdir` identifiers.
```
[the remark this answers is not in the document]
beckermr: This is not formally a requirement that can always be met, since the `subdirs` change. For example a channel might be `blah/bar/foo-128` and be valid. Then if someone comes along and declares a subdir `foo-128` later for a new platform and arch, we now have channels that were valid, but are now invalid. This seems like a rare edge case, but I wonder if we should think a bit more on this?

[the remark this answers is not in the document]
jaimergp: We can downgrade to SHOULD and add some language so folks avoid those pitfalls. e.g. don't register a channel named `linux-128` just in case 😂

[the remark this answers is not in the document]
beckermr: What is your proposal on changing matchspecs? We'd need a separator besides `/` or to force more items into the key=value bits.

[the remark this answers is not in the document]
jaimergp: We would deprecate the `channel/subdir` notation and instead point users to use `subdir=<value>` in the bracket specs (which is supported already):

```python
>>> from conda.models.match_spec import MatchSpec
>>> MatchSpec("conda-forge/linux-64::python")
MatchSpec("conda-forge/linux-64::python")
>>> MatchSpec("conda-forge::python[subdir=linux-64]")
MatchSpec("conda-forge/linux-64::python")
>>> MatchSpec("conda-forge/linux-64::python") == MatchSpec("conda-forge::python[subdir=linux-64]")
True
```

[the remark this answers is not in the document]
beckermr: Right. So to be clear these two things can be parsed reliably

- `<channel base URL>(/label/<label>)/<subdir>/<package or repodata or other file>`
- `<channel base URL>(/label/<label>)`

but this these are ambiguous

- `<channel base URL>(/label/<label>)/<subdir>`
- `conda-forge/label/linux-64`

The only thing we actually need is a way to tell the parser that even if a path component matches a subdir, to ignore that match.

Could we allow users to optionally use percent-encoding to encode the individual parts to resolve any ambiguity?

For example, suppose a channel is made that is

- `foo/bar/linux-128`

Then later, the subdir `linux-128` is added. We could let users denote this channel as

- `foo/bar/linux%2D128/label/osx%2D64/osx-64`
- `foo/bar/linux%2D128/osx-64`

with the idea that anything that is percent encoded is meant to be decoded when used as a URL, but to remain encoded when parsing.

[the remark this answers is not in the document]
jaimergp: In case of ambiguity we can also check for the existence of `my-channels/linux-64/repodata.json` and decide based on that, but doing IO to resolve parsing ambiguity sounds so wrong.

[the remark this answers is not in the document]
jaimergp: Yep, I thought about that but then it wouldn't actually correspond to the path on disk 🥲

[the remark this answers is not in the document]
jaimergp: Maybe we just recommend against it and defer to tools how to resolve the ambiguity. Future CEPs could resolve this by proposing a different file tree for channels.

beckermr, line 154:
```suggestion
To avoid ambiguous `MatchSpec` grammar, the last path component of a channel base URL SHOULD NOT match any `subdir` identifiers. If it does, the behavior in this ambiguous
case is not defined and implementation dependent.
```

beckermr (commented):
@jaimergp how do these changes look given our discussion?

cep-XXXX.md
```diff
@@ -0,0 +1,235 @@
...
+<table>
+<tr><td> Title </td><td> CEP XXXX - Identifying Packages and Channels in the conda Ecosystem </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td>
+  Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt; <br />
+  Matthew R. Becker &lt;becker.mr@gmail.com&gt; <br />
+  Cheng H. Lee &lt;clee@anaconda.com&gt;
+</td></tr>
+<tr><td> Created </td><td> Mar 11, 2025</td></tr>
+<tr><td> Updated </td><td> Apr 3, 2025</td></tr>
```
```suggestion
<tr><td> Updated </td><td> Apr 15, 2025</td></tr>
```

jaimergp:
Love it, thanks @beckermr

jaimergp:
@cj-wright, @mariusvniekerk, @ocefpaf, @marcelotrevisani, @msarahan, @mbargull, @jakirkham, @jezdez, @baszalmstra: your mid-period reminder to vote, thanks!

cep-XXXX.md
```diff
@@ -0,0 +1,239 @@
...
+To avoid ambiguous `MatchSpec` grammar, the last path component of a channel base URL SHOULD NOT
+match any `subdir` identifiers. If it does, the behavior in this ambiguous case is not defined
+and implementation dependent.
+
+#### Channel names
+
+For convenience, the channel _name_ is defined as the concatenation of `scheme`, `authority` and
+`path` components of a channel URL. At least one of `authority` or `path` SHOULD be present. In
+their absence, the channel name MUST be considered empty, regardless the scheme. Empty channel
+names SHOULD NOT be used.
```
mbargull, line 163: This deviates, e.g., from how `conda`'s implementation uses (or at least used to use in the past) the term `name`.
E.g., for `conda.models.channel.Channel` a `https://conda.anaconda.org/conda-forge/label/main` would (using the default channel alias) yield `conda-forge/label/main` as the `Channel.name`.
One reason for this that `conda` itself has (had?) no concept of "label" at all but only distribution related tooling (e.g., `conda.anaconda.org`'s server or tools like `anaconda-client` or `conda-smithy`) use that concept.

This is not to say it does not make sense to include the "label" concept at all -- it's certainly good to have this distribution-specific part specified since it's widely used, e.g., in `conda-forge`.
We should just be careful to align meanings of channel names, labels, etc. in specs and tooling (which this proposal is/can be used to get it going, of course).

cep-XXXX.md
```diff
@@ -0,0 +1,239 @@
...
+
+Channel label names MUST only consist of ASCII letters, digits, underscores, hyphens, forward
+slashes, periods, and whitespace. They MUST start with a letter. They MUST match this regex:
+`^[a-zA-Z][0-9a-zA-Z_\-\./]*$`. The last `/`-delimited component of a label
+SHOULD NOT match any `subdir` identifier. If it does, the behavior in this ambiguous
+case is undefined and implementation dependent.
+
+The label `nolabel` is reserved and MUST only be used for conda packages which have no other
+labels. In other words, in the space of labels, the empty set is represented by the labels
+`nolabel`.
```
mbargull, line 193: I'm not aware of `nolabel` reservation -- is this an existing thing for `conda.anaconda.org` or is it something to do with the OCI work mentioned in the comments?

mbargull (commented):
I'm very thankful that you continue to work on/write up how specification-like documents for all these previously only "implementation-defined" (and thus bound to be diverging) workings.

I vote to accept this proposal since it is a very good starting point to align implementations in package managers as well as distribution tools overall.

Unfortunately I wasn't able to take part in the discussion beforehand (but expect to be able to in the near future), so I apologize to not have given input before.
There are certain points (channel name/label definition, maximum length of package name/version/buildstring, etc.) that are either different from current definitions or possibly too restrictive.
E.g., maximum component lengths are worded with `MUST NOT exceed` even though we have existing cases that do exceed the given limits already -- hence it might make sense to have higher `MUST NOT exceed` limits and possibly have the current lower limits declared as `SHOULD NOT exceed`.
In my eyes in makes sense to declare this proposal as "provisional" since adjusting implementation to be "compliant to CEP XXXX" could be counter productive until those points have been worked on in later refinements to this proposal.

cep-XXXX.md
jaimergp: Related to OCI.

jaimergp:

* Total possible voters: 16. 
* Valid votes: 12 / 16 = 75%.

Yes votes (12 / 100%):

1. @xhochy
1. @chenghlee
1. @ocefpaf
1. @marcelotrevisani
1. @mbargull
1. @jezdez
1. @wolfv
1. @jaimergp
1. @baszalmstra
1. @beckermr
1. @hind-m
1. @trallard

No votes (0 / 0.00%).

Abstain votes (0 / 0.00%).

Not voted (4):

1. @cj-wright
1. @mariusvniekerk
1. @msarahan
1. @jakirkham

Invalid votes (0).

12/16 votes = 75% meets the minimum quorum. 
12/12 YES votes meets the approval threshold (60%).

jaimergp:
Thank you @mbargull (and welcome back!). I'll take a note of your feedback in a new issue so we can address this in a future revision.

beckermr:
I want to add in relation to @mbargull's comments that 

1. We cheeked conda-forge, bioconda, and defaults pretty extensively when writing these things down. There are a few things we purposefully decided to exclude because they were both rare and they made the spec more complicated. 

2. You seemed to indicate that you knew of things that violated the rules above. If they are not in the items we already know about (which are documented in the CEP) please post them here or in the new issue. 

3. Indeed some of the terms and usage do differ from conda's internals. Specifically for channel names, we've generalized to full URLs in anticipation of a less anaconda.org-centric ecosystem where packages are mirrored regularly. These changes are purposeful and should not be rolled back in a future CEP.

cep-XXXX.md
```diff
@@ -0,0 +1,239 @@
...
+include two consecutive separators (hyphen, period, underscore).
+
+Virtual package names MUST only consist of lowercase ASCII letters, numbers, hyphens, periods and
+underscores. They MUST NOT use two consecutive separators, with one exception: they MUST start with
+two underscores.
+
+Distributable package names MUST match the following case-insensitive regex:
+`^(([a-z0-9])|([a-z0-9_](?!_)))[._-]?([a-z0-9]+(\.|-|_|$))*$`.
+
+Virtual package names MUST follow this regex: `^__[a-z0-9][._-]?([a-z0-9]+(\.|-|_|$))*$`.
```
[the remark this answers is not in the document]
jaimergp: We want to disallow consecutive separators in the string, like `numpy--base` or `matplotlib.-extra`.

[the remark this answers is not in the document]
jaimergp: Maybe like this?
```re
^([a-z0-9]+|(_[a-z0-9]+))[._-]?([a-z0-9]+(\.|-|_|$))*$
```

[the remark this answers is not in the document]
beckermr: I've found these things are tricky and the negative lookahead was all that I could get to work when I was working on similar things.

PR #129, opened 2025-08-29
CEP XXXX: Improving dependency export infrastructure

### LINKED ISSUE #77: Adding `host_exports` (like `run_exports` except between `build` & `host` only)

h-vetinari:
First draft after discussion in #77. Does not contain (much) specification yet, because I'm unsure how to go about changing the schema of v1 recipes (does it need a bump in the schema version, or do we specify build tools must translate between them?), and how to deal with the repodata side of things. This is my first CEP, please excuse my lack of experience with a lot of the underlying details.

Help on these questions would be much appreciated! I decided to write up the design in more comprehensive form than originally in this [comment](https://github.com/conda/ceps/issues/77#issuecomment-3186740038) though, in order to hopefully facilitate more effective discussion of how to solve the transition issues posed by the new design.

Closes #77 (eventually)

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -169,6 +169,23 @@ the ground up based on all the lessons learned in the last decade would be much 
 lowering the barrier to entry into the ecosystem, and helping existing contributors navigate packaging
 challenges more easily.
 
+### The many uses of exports
+
+What began as a simple idea (e.g. to inject a runtime dependency on a shared library from `host:`
+to `run:`) expressible with a single verb, _to export_ (and its nominalization into `run_exports:`)
+has turned into a key concept across many use-cases. As a consequence, exports have become ubiquitous.
+
+Not only that, but we need to describe fundamentally different processes about exports:
+
+- Attaching an export (e.g. to a package `foo`)
+- Applying an export (e.g. when building a package `bar` depending on `foo`)
+- Ignoring an export (e.g. if `bar` needs `foo` present, but doesn't want the export itself)
+- Inheriting an export (when talking about `package:`-global builds influencing their `outputs:`,
+  which may be `files:`-based and thus have no `build:`/`host:` environments of their own, yet
+  still need to reflect the runtime dependencies of their content)
+
+These actions are currently either implicit or dispersed incongruously across the recipe.
+
 ## Design
 
 We begin with the following observations based on the above:
@@ -186,6 +203,7 @@ We begin with the following observations based on the above:
   which have so far been under-served by the existing run-export infrastructure.
 - Run-exports are not applied when building `noarch` packages, except if the export is of type `noarch:`.
 - The v1 recipe format unified all requirement-related topics (including run-exports) under `requirements:`.
+- Exports and their associated actions should be consolidated as well.
 
 Based on this, we propose the following pattern:
 
@@ -199,28 +217,60 @@ requirements:
     - [...]
   # relying on the surrounding "requirements" key for context
   exports:
-    host_to_run:            # matches `weak:` run-export
-      - a_shared_library
-    build_to_host:          # "host-export"
-      - a_host_constraint =*=*foo
-    build_to_run:           # produces same effect as `strong:` run-export when used together with build_to_host
-      - a_compiler_runtime
-    host_to_constraints:    # matches `weak_constrains:` (v0) / `weak_constraints:` (v1)
-      - a_run_constraint
-    build_to_constraints:   # matches `strong_constrains:` (v0) / `strong_constraints:` (v1)
-      - a_run_constraint
-    # all the above _do not_ apply when building `noarch: generic` or `noarch: python` packages
-    noarch_to_run:          # matches `noarch:` run-export; _does_ apply when building noarch packages
-      - a_dependency_exported_when_consumer_is_noarch
-    # potential future additions, not proposed by this CEP, but included for completeness
-    build_to_build:
-      - a_compile_time_only_dependency
-    host_to_host:
-      - a_compile_time_only_dependency
+    # exports that the output itself will apply when *used* for building other packages
+    attach:
+      host_to_run:            # matches `weak:` run-export
+        - a_shared_library
+      build_to_host:          # "host-export"
+        - a_host_constraint =*=*foo
+      build_to_run:           # produces same effect as `strong:` run-export when used together with build_to_host
+        - a_compiler_runtime
+      host_to_constraints:    # matches `weak_constrains:` (v0) / `weak_constraints:` (v1)
+        - a_run_constraint
+      build_to_constraints:   # matches `strong_constrains:` (v0) / `strong_constraints:` (v1)
+        - a_run_constraint
+      # all the above _do not_ apply when building `noarch: generic` or `noarch: python` packages
+      noarch_to_run:          # matches `noarch:` run-export; _does_ apply when building noarch packages
+        - a_dependency_exported_when_consumer_is_noarch
+      # potential future additions (see CEP-XXXY), but included for completeness
+      build_to_build:
+        - a_compile_time_only_dependency
+      host_to_host:
+        - a_compile_time_only_dependency
+
+    # exports from dependencies in the current output that should be ignored when *building*
+    ignore:
+      to_any:           # default; `exports: ignore: foo` maps to `exports: ignore: to_any: foo`
+        from:
+          - zlib
+        by_name:
+          - libzlib
+      to_run:
+        [...]           # same inner schema
+      to_host:
+        [...]           # same inner schema
+      # potential future addition (see CEP-XXXY)
+      to_build:
+        [...]
+
+    # exports to be inherited from staging output while *building* the output;
+    # NOT PROPOSED here, but relevant for https://github.com/conda/ceps/pull/102
+    inherit:
+      to_any:           # default; `exports: inherit: foo` maps to `exports: inherit: to_any: foo`
+        from:
+          - name-of-staging-output
+      to_{run,host,build}:
+        [...]           # same inner schema
 ```
 
-As indicated by the comments, `host_to_run:` matches the existing weak run-export. If taken together with
-`build_to_run:` this produces the same effect of a strong run-export. Similarly for `host_to_constraints:`
+This unifies all export-related things under `requirements.exports`, as well as collecting the
+different actions `{attach,ignore}` (plus potentially `inherit` in the future) in one place. We'll look
+at the `ignore:` side further down; let us start with the heart of the matter: to `attach:` an export
+to the package being built.
+
+As indicated by the comments, `host_to_run:` matches the existing weak run-export attached to a package.
+If taken together with `build_to_run:`, this produces the same effect of a strong run-export.
+Similarly for `host_to_constraints:`
 and `build_to_constraints:`. The other keys introduce new functionality, which is explained below.
 
 This design has the advantage that it's
@@ -256,7 +306,7 @@ together with constraints (such as `_fortran_modules_abi`) attached to packages 
 
 While the `build:` to `host:` to `run:` order described above reduces procedural complexity, it still enlarges
 the space of possible scenarios. Specifically, there are now at least two distinct points where exports happen,
-once for `build_to_*:` and once for `host_to_run:`. This leads to new questions, of how chained exports interact.
+once for `build_to_*:` and once for `host_to_run:`. This leads to new questions how chained exports interact.
 
 For example, if we have
 
@@ -264,8 +314,9 @@ For example, if we have
 # output: a_complicated_package
 requirements:
   exports:
-    build_to_host:
-      - some_package_with_a_run_export
+    attach:
+      build_to_host:
+        - some_package_with_a_run_export
 ```
 
 and
@@ -274,8 +325,9 @@ and
 # output: some_package_with_a_run_export
 requirements:
   exports:
-    host_to_run:
-      - the_export_in_question
+    attach:
+      host_to_run:
+        - the_export_in_question
 ```
 
 and then consume it
@@ -295,7 +347,7 @@ requirements:
 
 the question arises how to handle export of a package _that's not explicitly named_ in the recipe.
 
-Obviously this is quite an impactful question, as it is certainly not desirable to trigger the exports of all
+Obviously this is quite an impactful question, as it is certainly not desirable to apply the exports of all
 packages that happen to transitively make it into a build or host environment.
 
 Taken in isolation, it would be understandable to conclude from the above that the rule "only packages named
@@ -326,8 +378,9 @@ To illustrate this better, let us look at some example recipes:
 requirements:
   # omitted: regular build/host/run environments
   exports:
-    host_to_run:
-      - ${{ pin_compatible("libA") }}
+    attach:
+      host_to_run:
+        - ${{ pin_compatible("libA") }}
 ```
 
 This matches the operations of `run_export:` so far. Before we consider the recipe of `libB`, let us consider
@@ -363,11 +416,12 @@ requirements:
     # regular host_to_run export from libA
     # - libA >={{ver_A}},<{{next_ver_A}}
   exports:
-    host_to_run:
-      - ${{ pin_compatible("libB") }}
-    # export that injects libA whenever libB is a named dependency in a host environment!
-    host_to_host:
-      - ${{ pin_compatible("libA") }}
+    attach:
+      host_to_run:
+        - ${{ pin_compatible("libB") }}
+      # export that injects libA whenever libB is a named dependency in a host environment!
+      host_to_host:
+        - ${{ pin_compatible("libA") }}
 ```
 
 The main additional complication this introduces (and why it is not being proposed in this CEP yet), is that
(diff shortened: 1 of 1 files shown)


h-vetinari:
pre-commit.ci autofix

h-vetinari:
pre-commit.ci autofix

h-vetinari:
> because I'm unsure how to go about changing the schema of v1 recipes (does it need a bump in the schema version, or do we specify build tools must translate between them?)

Wolf mentioned in private that we don't necessarily have to go to a v2 schema over this, because despite being renamed semantically, the new keys would just be _extending_ the v1 schema, not breaking it. Of course, we'd then have to mandate mutual exclusivity between `exports:` and `run_exports:` etc., but I think that's probably a gentler approach to this than taking this relatively minor issue as cause for "recipe v2".

The same approach (consider the new keys if present, error if not mutually exclusive with the old way) could even by used by conda-build to support the CEP[^1], which would be great because a lot of our compiler feedstocks that would need this the most are not necessarily ready to be ported to v1 yet. :)

If people are in agreement over this approach, I can try to write up specification for it.

In any case, I think this is in a good enough state to ask for a first round of feedback; I'd be very curious to hear the thoughts of
@chenghlee @isuruf @jezdez @msarahan @wolfv, people from mamba, prefix, cf/core, and anyone else interested in this!

[^1]: somewhat informally perhaps, since it's currently formulated atop of the v1 format.

h-vetinari:
Thanks for the feedback! :)

> I was thinking it might be useful to have an `build_to_host_and_run` as well (aka the `strong` run exports).

My current thinking is that there are quite few cases that actually _need_ this; mostly it's a big hammer to achieve _either_ `build_to_host:` _or_ `build_to_run:`. So I have a preference to keeping them separate, and letting recipes be explicit in the cases where something needs to be exported to both.

> I also thought that `noarch_to_run` is maybe somewhat confusing. Maybe `export_when_noarch` or something, as an alternative?

I only found out about `noarch:` run-exports while going through the CB code base for this CEP, so it's a bit of a late addition to the design, and perhaps I'm overlooking a more elegant option. That said, I kinda feel that "noarch is a state of ~mind~ building", and as such it still makes sense to plug into the pattern `<condition>_to_<target>`. But it's definitely the odd one out in any case, so I'm open for different names on this.

> This is going to be terrible to implement in this way as it is basically a conditional dependency and thus influences the solve.

I don't see how it affects the solve? To my (limited) understanding of conda-build, there's [already](https://github.com/conda/conda-build/blob/25.7.0/conda_build/render.py#L507-L523) a separate pass to add run-exports before anything is being resolved. Also conceptually, you don't need to resolve anything to do `build_to_build:` injection, the only thing you need is to look up the fully static `exports.json` of the packages in the respective environment, and apply the right ones, before going to the resolver.

> We need to use the conditional dependencies here and do something like a run dependency that looks like: [...] Which would be something that we should add to the CEP about repodata v2.

My first reaction to that is... not great, but perhaps I'm not seeing yet what conditions you see would make this necessary. I'm fully onboard with repodata v2 fixing a bunch of things, but at first glance, that recipe snipped looks like a solution in search of a problem to me. I don't see why I'd want to qualify run-dependencies w.r.t. which environment the package gets installed in. That sounds like a way bigger headache to me (in terms of increased complexity, especially on a conceptual level) than doing a pre-processing pass for exports that looks up some static data.

That said, I'm not super-attached to `build_to_build:` and `host_to_host:`. I've included them as a favour to Isuru, but my main goal here is providing a clean design and solving the modules ABI situation. If people prefer to leave transitive exports for another CEP, that's also fine - at least the current design leaves the door open for a natural way to add them, even if it comes later.

baszalmstra:
I am curious what constraints you would want to impose. But maybe thats something we should discuss in the other CEP

h-vetinari:
> I am curious what constraints you would want to impose. But maybe thats something we should discuss in the other CEP

Happy to discuss this in the other CEP

wolfv:
I had another think on this and while this could be solved as conditional dependency, it could still be listed under `build_to_build`. Since we have run-exports available from the repodata (shards) we could include them in the solve as well. It would be slightly less elegant vs. conditional dependencies from the implementation perspective of resolvo / rattler though.

Or we could normalize this in rattler build and turn these "exports" into conditional dependencies :)

isuruf:
> I had another think on this and while this could be solved as conditional dependency, it could still be listed under build_to_build.

We could add the following condition.

`build_to_build` condition has to match the `run` condition in the upstream package.

i.e. we need to have

```
requirements:
  run:
    - foo >=2
  exports:
    build_to_build:
      - foo >=2
```

baszalmstra:
Currently run exports are only included from packages on which the recipe directly depends, so not for transitive dependencies but if I read the above discussion correctly we have to make an exception for `build_to_build` and `host_to_host` exports, is that correct?

AntoinePrv:
I don't know if conditional dependencies will be faster for the solver. After all it's NP-hard so three solves might not be slower than one with more constraints (to be fair I'm unsure about it either way). Also for most recipe that need this, solving time pales in comparison to compile time.

Another thing about the doing it with conditional dependencies do end up in the final package dependencies.
So we would have to delay the adoption of this CEP until conditional dependencies are adopted, implemented by majors client, and a reasonable time is given to users to upgrade.

Looking at the recipe I don't think conditional dependencies do a good job at understanding what they are for. I also think being conditional on an environment variable is fragile (although that could be changed for some static key passed to the solver that is only set to true by conda/rattler passing it to the solver).
The `X_to_Y` is more descriptive in my opinion and does not prevent a future CEP to mandate that the build tool need to transform these keys into conditional dependencies.
Heck, even a ~conditional dependency enabled~ solver today could already read the relevant repodata entries and add them based on which environment they are creating no?

h-vetinari:
> I don't know why everyone is talking about conditional dependencies.

Because adding dependencies to the same environment (e.g. from host to host) is 100% equivalent to a dependency that triggers under certain circumstances (like being in host).

> if `foo` is in `host` and `bar` is a dependency of `foo`, we need a way to add `bar` conditions in `run`.

This is yet another case from what you've described before (a variant of `host_to_run:`, instead of `host_to_host:` as previously). Unless foo exports bar by itself, bar is completely invisible from the POV of the build that's consuming foo, and I _definitely_ don't want to start applying any exports from packages not explicitly named in the recipe.

Please respect my decision that in this CEP, your use-case is out of scope. You can write your own CEP for that (either building on top of this, or competing with mine if you must).

isuruf:
> You can write your own CEP for that 

Sure

h-vetinari:
> I read the whole thread and think the proposed solution is a good addition. (From the perspective of writing recipes.)

Thanks a lot for the feedback @cbouss, I appreciate you taking the time!

That reminds me that I wanted to post a comment[^1] from a recent [discussion](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Yet.20another.20.60stdlib.28.22c.22.29.60.20discussion/near/560278679) on zulip which covers an open point of the design here: how to deal with `host_to_run:` exports of packages that were themselves injected via `build_to_host:` (i.e. not explicitly named in the consuming recipe).

Concretely, if we have
```yaml
# output: a_complicated_package
requirements:
  exports:
    build_to_host:
      - some_package_with_a_run_export
```
and
```yaml
# output: some_package_with_a_run_export
requirements:
  exports:
    host_to_run:
      - the_export_in_question
```
and then consume it
```yaml
# output: mypkg
requirements:
  build:
    - a_complicated_package
  host:
    # from a_complicated_package's build_to_host
    # - some_package_with_a_run_export
  run:
    # ...should the run-export of some_package_with_a_run_export get triggered here?!
    # - the_export_in_question
```
the question is whether, why and how we trigger an export of a package _that's not explicitly named_ in the recipe. Obviously this is quite an impactful question, and we certainly _do not_ want to trigger all exports of packages that happen to transitively make it into an environment.

On zulip, I discussed this in the context of how injected exports could/would work. Despite not wanting to cover self-exports in this CEP, I've used the motivating use-case for them as an example here, because it illustrates the situation more concretely than just some abstract arithmetic between environments.

As I note at the end, _this CEP_ could go either way (i.e. whether `mypkg` gets a run-requirement of `the_export_in_question` or not). However, I'm trying to take a wider view of the design space here, and to surface potential problems early. The flip side is that it's possible to go too far ahead in the "what if" exercise and get bogged down without ever achieving the first step. In any case, I'd be interested what people think.

[^1]: I've edited the comment to leave out the bits not relevant here.
------

Let's take the [clearest](https://github.com/conda/ceps/issues/77#issuecomment-3210307252) [case](https://github.com/conda/ceps/pull/129#issuecomment-3254627079) that I'm aware of that's been in contention around self-exports vs. conditional dependencies. I'm rephrasing here for consistent language: `libB` depends on `libA` as usual, but compiling against `libB` needs *the specific version* of `libA` used to build `libB`, e.g. due to the way ABI and headers between `libA` and `libB` interact. 

In other words, if `libB 1.0.0 *_1` is built against `libA=1` and `libB 1.0.0 *_2` is built against `libA=2`, we need to match the `libA` version constraint of `mypkg` that (generically) depends on`libB` with the *specific* constraints on the artefact of `libB` that was in `host:` at build time.

Let's look at the recipes:

```yaml
# output: libA
requirements:
  [...]  # regular build: / host: / run:
  exports:
    host_to_run:
      - ${{ pin_compatible("libA") }}
```
That's the easy part, which we already know as `run_export:` today. For `libB`, the recipe would use conditional dependencies
```yaml
# output: libB
requirements:
  [...]  # regular build:
  host:
    - libA 
  run:
    # regular run-export from libA
    # - libA >={{ver_A}},<{{next_ver_A}}
    # additional requirement when compiling against libB; same effect as host_to_host
    - ${{ pin_compatible("libA") }}; if env.HOST  # how to spell this is TBD!
  exports:
    host_to_run:
      - ${{ pin_compatible("libB") }}
```

Now, whether as a conditional dependency or a `host_to_host:` export, using `libB` *necessarily* needs to inject a constraint on `libA` that's not present in the consuming recipe

```yaml
# output: mypkg
requirements:
  [...]  # regular build:
  host:
    - libB
    # injected!
    # - libA  # matches libA-constraint of libB
  run:
    # regular run-export from libB
    # - libB >={{ver_B}},<{{next_ver_B}}
    # run-export from injected libA!
    # - libA >={{ver_A}},<{{next_ver_A}}
    - some_regular_dep
```
In my mental model, the process is as follows:
* Solve `build:` env
* determine exports: for  packages that have been named ***or injected*** in `build:`, collect any `build_to_host:` or `build_to_run:`, call them B2H, B2R
* Add B2H to `host:`, solve
* determine exports: for  packages that have been named ***or injected*** in `host:`, collect `host_to_run:`, call them H2R
* Save `run:` + B2R + H2R as run deps of the resulting artefact.

The "***or injected***" part is the most open aspect of the design space, but I believe the example with `libA/libB` above shows that we cannot rely only on named packages in such cases. So my proposed rule would be: "we apply exports from packages that are either explicitly named in the environment, or have been injected via exports or conditional dependencies". The second part of this rule only exists because there are cases that cannot be solved with just the first part.

Coming back to your example [for automatically injecting `${{ stdlib("c") }}`], this could be written as
```yaml
# mycompiler
requirements:
  run:
    - ${{ stdlib("c") }}
    # conditional dependency when in build, which then participates in run-exports; same as build_to_build
    - ${{ stdlib("c") }}; if env.BUILD
```
This would work with the "***or injected***" scheme above, but as you can see, it's neither very elegant nor obvious why you'd have to repeat the same dependency for `mycompiler`, and how that triggers the export from `${{ stdlib("c") }}` in `mypkg` [from using `mycompiler`].

IMO that's because conditional dependencies and self-exports are features that are only truly needed for niche cases; any feature can be misused, so any additional expressivity needs to be guarded (e.g. by the linter etc.). As a consequence, I'm convinced that we shouldn't abuse this mechanism for something which *every* compiled recipe needs. In other words, `${{ stdlib("c") }}` is too large a use-case to fit into this niche, and the consequences of trying to hide it are not worth the benefits.

This comment is already too long, so I'll just note that a more restricted form of [this PR] without the "***or injected***" is possible, and that this would already solve a lot of the cases where we need host-exports, e.g. the ABI of C++/Fortran modules. It's only when we get to stuff like the `libA/libB` case above that we really need to go beyond the "only named packages can contribute exports"; though I do think (compare the list of steps above) that the same mechanism would also be quite natural for chaining e.g. `build_to_host:` with `host_to_run:` exports.

h-vetinari:
> Obviously this is quite an impactful question, and we certainly _do not_ want to trigger all exports of packages that happen to transitively make it into an environment.

One thing that gnawed at me since posting that comment on zulip is that I don't like the relationship "conditional dependencies participate in run-exports", which seems too magical, and is by far not explicit enough in the recipe IMO.

However, I just had an idea that might solve this case: we can use `export: host_to_host: ...` as the _syntax_ to indicate this unusual mechanism, but _implement_ it using conditional dependencies. IOW, `libB` from above would look like
```yaml
# output: libB
requirements:
  exports:
    host_to_run:
      - ${{ pin_compatible("libB") }}
    host_to_host:
      # implemented not as an export, but as a conditional dependency!
      - ${{ pin_compatible("libA") }}
```


That would IMO be the best of both worlds: explicit syntax, as well as a sane environment resolution process, and conditional requirements don't have to be imbued with some magical pixie dust (pun intended 😉).

This would then give the following dependencies between the different proposals

```mermaid
---
config:
  securityLevel: loose
---
flowchart TB
    this["this CEP"]-->se["self exports CEP"]
    cond["conditional dependencies CEP"]-->se
```

h-vetinari:
> Nice that we are finding paths forward!

Glad to hear it. It's not for lack of wanting to solve the issue that I had descoped self-exports; now I'm beginning to see a path that allows the various pieces to work together in a non-hacky way (where before I couldn't see it at all).

> Is this similar to the solution I proposed in [#129 (comment)](https://github.com/conda/ceps/pull/129#issuecomment-3249504891), or does it differ in some way? If that’s the case, can you elaborate how it’s different?

It's different in that we do not have to add a condition like "`build_to_build:` has to match `run:`", but we still maintain the fact that it doesn't require multiple solves.

I've been getting down into the nitty gritty details at least one level deeper (e.g. @jaimergp opened another rabbit hole under my feet about the mechanics of `ignore_exports:`, and I had an illuminating chat with @baszalmstra about the steps that (can) happen between the metadata and the resolver); I'm planning to expand the CEP with the design conclusions from those discussions, which should hopefully tell the whole story better than yet another very long comment.

h-vetinari:
OK, I actually needed to see a third verb related to exports (`inherit:` in #102) for it to click for me that it makes no sense to add `ignore_exports:` next to `exports:`. So the most recent commit here changes this to
```yaml
requirements:
  exports:
    inherit:  # exports to be inherited from cache stage when the output is being *built*
    ignore:   # exports to be ignored (from its dependencies) when the output is being *built*
    attach:   # exports that the output itself will apply when *used* for building other packages
```

I'm not super attached (ha!) to the verb `attach:`, but as I try to explain in the updated introduction, we have too many actions that piggyback on the verb "export" currently (itself describing the cross-environment dependency injection; whereas we also need language to inherit/ignore/attach such exports). If you have a better idea (or any other feedback), I'm all ears!

JeanChristopheMorinPerso:
>attach:   # exports that the output itself will apply when *used* for building other packages

@h-vetinari Maybe `apply`? If I look at what you suggest:

>    inherit:  # exports to be **inherited** from cache stage when the output is being built
>    ignore:   # exports to be **ignored** (from its dependencies) when the output is being built
>    attach:   # exports that the output itself will **apply** when used for building other packages

The description you give says "apply", which kind of makes sense I think. Or maybe `inject` could work too, considering that the export will be injected into the current artifact (itself).

h-vetinari:
Thanks for the comment!

> The description you give says "apply", which kind of makes sense I think. Or maybe `inject` could work too, considering that the export will be injected into the current artifact (itself).

We "apply" exports at a very different time from when we're building the current artefact that will carry the export, so I think that would be a confusing verb to use for the act of imbuing the artefact with the export in the first place.

I think "inject" would be perfectly fine (though I still have a slight preference[^1][^2][^3] for "attach").

[^1]: I think it depends a bit on one's mental image of what a package looks like metaphorically; if you picture it as a closed container, then "inject" or "insert" makes more sense; if you picture it as a collection of things strapped together, "attach" makes more sense.

[^2]: Since it came to me while struggling to find a good formulation above: "imbue" would also work (though probably too obscure).

[^3]: I'll admit it's tempting to have all three verbs start with the same letter; my first thought was "instate", as "install" is too overloaded here.

PR #130, opened 2025-08-30
readability edits for CEP-21

h-vetinari:
While reading https://github.com/conda/ceps/blob/main/cep-0021.md for work on #129, I noticed some minor issues, which I'm suggesting to fix here.

The big issue that I don't know how to fix is that the "Specification" section is IMO mislabeled, as it only contains discussion of reasons that CEP 12 had brought for not embedding run-exports into the regular repodata. The only language that uses authoritative keywords (MAY, MUST) is under "Patching" (though that's also mixed with non-technical language, e.g. "We do not see a reason how [...]"), so presumably that must be the entirety of the specification?

CC @baszalmstra


cep-0021.md
```diff
@@ -25,24 +25,25 @@ However, not all channels on the default server (conda.anaconda.org) provide thi
...
 CEP 12 mentions the following reasons for splitting the information into two files:
 
 > * It would require extending the repodata schema, currently not formally standardized.
 > * It would increase the size of the already heavy repodata.json files.
 > * (Typed) repodata parsers would need to be updated to handle the new field.
+
 We propose that these reasons no longer hold with [sharded repodata](https://github.com/conda/ceps/blob/main/cep-0016.md).
 
-**It would require extending the repodata schema, currently not formally standardized.**
+### Re: "It would require extending the repodata schema, currently not formally standardized."
```
h-vetinari, line 38: The screenshot above also shows that the current highlight makes it seem like part of the regular text of the CEP. I don't care about the exact formatting, as long as it is clear that it's a point being addressed.

baszalmstra:
I understand the edits but now the part about that the run exports are stored inside the repodata is part of the discussion. Maybe we should then clarify this in the (new) specification section with MUST?

h-vetinari:
I don't know the process for editing accepted CEPs; in any case, I rely on you the author to make those changes, I just wanted to get the ball rolling. :)

If you want I can give you collaborator rights on my fork, then you can push directly into this PR.

PR #142, opened 2025-12-10
CEP about serving sigstore attestations in Conda repositories

wolfv:
This CEP adds specifications how to serve sigstore attestations in Conda repositories. 

A preview implementation is available on beta.prefix.dev under the following URL:

https://beta.prefix.dev/wolf-channel/linux-64/signed-package-1.2.1-hb0f4dca_0.conda.v0.sigs

## Checklist for submitter


cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+```json
+[
+  <Sigstore Bundle>,
+  <Sigstore Bundle>,
+  ...
+]
+```
+
+Each element in the array MUST be a valid [Sigstore Bundle] as defined by the Sigstore specification. The bundle format supports multiple versions; implementations SHOULD support at least bundle versions v0.2 and v0.3.
```
facutuesca: I don't know if we should specify bundle versions here. This CEP defines the how to serve attestations (format, location) so that clients can access them. But IMHO, the bundle itself should be a black box to this CEP, other than "each element must be a valid sigstore bundle" I don't think we should go into internals (which are covered by CEP 27)

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+  }
+]
+```
+
+### HTTP Status Codes
+
+| Status Code | Meaning |
+|-------------|---------|
+| `200 OK` | Attestations returned successfully (may be empty array) |
+| `404 Not Found` | The package does not exist (distinct from "no attestations") |
```
facutuesca: Adding more context from my older comment about this being backwards compatible:
I think 404 should have the meaning "the package does not exist AND/OR the channel does not support attestations".
Otherwise (new) clients communicating with (old) channels could try to retrieve the `.sigs` file, see 404, and interpret "the package does not exist", where the reality is that the channel does not support yet this CEP

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+| Third-party auditor | Proves the package passed security review |
+
+When multiple attestations are present, they MUST all refer to the same artifact (same filename and SHA256 hash). Clients MAY choose which attestations to verify based on their trust policy.
+
+### Mirror Behavior
+
+Mirrors and proxies SHOULD:
+
+1. Fetch and cache `.sigs` files alongside packages
+2. Serve cached attestations without modification
```
facutuesca: Can attestations for a file be uploaded post-facto? If not, might that be a future feature? (like the example above, where a third party auditor might review and upload and attestation after the artifact is publicly available)

If so, this section should account for how mirrors should update changed `.sigs` files. Is there a way for mirrors (or any client) to detect such changes?

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+Clients implementing attestation verification SHOULD follow this workflow:
+
+1. **Download package** from the channel (or use the SHA256 sum from the repodata.json)
+2. **Fetch attestations** from `<package_url>.sigs`
+3. **Verify each attestation** according to the client's trust policy:
+   - Verify the Sigstore bundle signature
+   - Verify the certificate chain to Fulcio root
+   - Verify the transparency log inclusion proof
+   - Verify the in-toto subject matches the downloaded package
```
facutuesca: I would not put this here, I would just refer to the verification process specified in CEP 27

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+        - "https://github.com/foobar"
+```
+
+| Setting | Values | Behavior |
+|---------|--------|----------|
+| `enabled` | `true`/`false` | Enable or disable attestation fetching and verification |
+| `require` | `error` | Fail if attestations are missing or invalid |
+| | `warn` | Log warning but continue if attestations are missing or invalid |
+| | `ignore` | Silently continue (still verify if attestations exist) |
+| `trusted_identities` | List of patterns | Only accept attestations from matching Sigstore identities |
```
facutuesca: is this practical? I could see the usecase for a private channel where all artifacts are uploaded by the same trusted identity, but in the general usecase there will be one identity per package, so it doesn't make sense to list them all here (rather than relying on a lockfile with identities). I don't think we should require this of the clients, instead we might want to require the clients supporting identities in lockfiles.

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Offline and Air-gapped Environments
+
+For offline verification, clients MAY cache `.sigs` files alongside packages in local repositories
+The Sigstore bundle format is self-contained and supports offline verification once the Sigstore trust root is available locally.
```
facutuesca: Maybe we should mention here that the trust root should be updated periodically, since part of its security guarantees come from the fact that target/keys can be revoked

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+| Compromised signing identity | Transparency log enables detection; trust policy limits blast radius |
+
+### Limitations
+
+This scheme does NOT protect against:
+
+1. Legitimate signers publishing malicious packages
+2. Compromise of the Sigstore infrastructure itself
+3. Incorrect client trust policies
+4. Attacks before attestation was added to the package
```
facutuesca: I'm not sure we should rehash the security model here (which should be covered by CEP 27).

cep-xxxx-serving-attestations.md
wolfv: Your proposal would be to do some kind of TOFU from the lockfile? Ie. warn / error when an identity changes?

I think for a lot of channels, the publishers are much more homogenous vs. PyPI. E.g. one could at least choose the Github organization with wildcard, or even the repository for an entire channel (e.g. see the `rust-forge`: https://github.com/wolfv/rust-forge)

cep-xxxx-serving-attestations.md
aovasylenko: i'd also want to understand it better, if we need always return empty [] file for any package's *.sigs request.
if we just serve channel from s3 bucket - it could require additional server layer implementation

evankanderson (commented):
I swung by, since this got posted to the [OpenSSF #general channel](https://openssf.slack.com/archives/C019M98JSHK/p1773415327814499), and I was curious.  Feel free to disregard.

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,270 @@
...
+
+[CEP 27] defines a standard attestation format for the conda ecosystem using [in-toto] statements and [Sigstore] bundles. However, it explicitly leaves the distribution mechanism as future work:
+
+> "This CEP does not specify a distribution mechanism for attestations (i.e., Sigstore bundles containing attestations)."
+
+Without a standardized distribution mechanism, clients cannot reliably discover and retrieve attestations. This CEP addresses that gap by defining a simple, RESTful endpoint that:
+
+1. **Enables client verification**: Clients can fetch attestations alongside packages and verify them before installation.
+
+2. **Supports multiple attestations**: A single package may have multiple attestations (e.g., from the build system, from the channel on upload, from third-party auditors).
```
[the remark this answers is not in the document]
facutuesca: > While this CEP describes the endpoint as "RESTful", it seems like it may be a read-only endpoint?

These are not in contradiction, right?

> If you want to support additional attestations from third-parties, you'll need to figure out access control for users "vouching" for other packages. That seems like a diversion from the main goal of this CEP, so it might be worth calling out third-party auditors as a future capability that you don't want to prevent.

Not necessarily? A third party could generate attestations, and a channel admin could upload them to their respective packages, without giving access to the third party. This section is characterizing the distribution mechanism, the "attestations from third-party auditors" is just an example of a type of attestation that may exist.

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,270 @@
...
+
+The repodata index is changed to include a new `attestations` field that MUST contain the SHA256 hash of the signatures file.
+
+```
+{
+  "name": "foobar",
+  "version": "1.2.3",
+  "attestations": "37517e5f3dc66819f61f5a7bb8ace1921282415f10551d2defa5c3eb0985b570"
+}
+```
```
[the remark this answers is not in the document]
facutuesca: +1

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,270 @@
...
+    trusted_identities:
+        - "https://github.com/foobar"
+```
+
+| Setting              | Values           | Behavior                                                        |
+| -------------------- | ---------------- | --------------------------------------------------------------- |
+| `enabled`            | `true`/`false`   | Enable or disable attestation fetching and verification         |
+| `require`            | `error`          | Fail if attestations are missing or invalid                     |
+|                      | `warn`           | Log warning but continue if attestations are missing or invalid |
+|                      | `ignore`         | Silently continue (still verify if attestations exist)          |
```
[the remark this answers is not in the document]
facutuesca: I vote for making the 3 fields mandatory, with no default values

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Client Behavior
+
+### Verification Workflow
+
+Clients implementing attestation verification SHOULD follow this workflow:
+
+1. **Download package** from the channel
+2. **Fetch attestations** from `<package_url>.sigs`
+3. **Verify each attestation** against the configuration.
```
facutuesca: make explicit that the verification process is defined in CEP-27

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,261 @@
...
+|                      | `warn`           | Log warning but continue if attestations are missing or invalid |
+|                      | `ignore`         | Silently continue (still verify if attestations exist)          |
+| `trusted_identities` | List of patterns | Only accept attestations from matching Sigstore identities      |
+
+### Offline and Air-gapped Environments
+
+For offline verification, clients MAY cache `.sigs` files alongside packages in local repositories.
+The Sigstore bundle format is self-contained and supports offline verification once the Sigstore trust root is available locally.
+
+Note: clients MUST periodically update the sigstore trust root to ensure no keys were revoked.
```
facutuesca: the reason is more general, it's to ensure the client doesn't miss any trust root changes (which could be key revocation, but also new keys added)

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+1. Legitimate signers publishing malicious packages
+2. Compromise of the Sigstore infrastructure itself
+3. Incorrect client trust policies
+4. Attacks before attestation was added to the package
+
+## Backwards Compatibility
+
+This proposal is fully backwards compatible:
+
+1. **Existing channels**: No changes required; clients will receive `404` for `.sigs` endpoints
```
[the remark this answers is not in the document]
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+Without a standardized distribution mechanism, clients cannot reliably discover and retrieve attestations. This CEP addresses that gap by defining a simple, RESTful endpoint that:
+
+1. **Enables client verification**: Clients can fetch attestations alongside packages and verify them before installation.
+
+2. **Supports multiple attestations**: A single package may have multiple attestations (e.g., from the build system, from the channel on upload, from third-party auditors).
+
+3. **Works with existing infrastructure**: The sidecar file approach integrates naturally with static file hosting, CDNs, and mirrors.
+
+4. **Follows ecosystem conventions**: Similar approaches are used by PyPI ([Integrity API][PyPI Integrity]), npm ([provenance attestations][npm provenance]), and RubyGems.
```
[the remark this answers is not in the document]
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Multiple Attestations
+
+A package MAY have multiple attestations from different sources. Common scenarios include:
+
+| Source | Purpose |
+|--------|---------|
+| Build system (e.g., GitHub Actions) | Proves the package was built from specific source code |
+| Channel operator | Proves the channel accepted and published the package |
+| Third-party auditor | Proves the package passed security review |
```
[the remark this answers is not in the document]
facutuesca: this has not been addressed yet

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,295 @@
...
+
+TODO: specify further the desired behavior of mirrors.
+
+## Client Behavior
+
+### Verification Workflow
+
+Clients implementing attestation verification SHOULD follow this workflow:
+
+1. **Download package** from the channel (or use the SHA256 sum from the repodata.json)
```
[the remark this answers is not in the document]
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
facutuesca: this thread can be resolved

cep-xxxx-serving-attestations.md
```diff
@@ -0,0 +1,261 @@
...
+]
+```
+
+### HTTP Status Codes
+
+| Status Code     | Meaning                                                      |
+| --------------- | ------------------------------------------------------------ |
+| `200 OK`        | Attestations returned successfully (may be empty array)      |
+
+Channels that support attestations MUST always return `200 OK` with an empty array `[]`, even when the package does not exist.
```
facutuesca: My [previous comment](https://github.com/conda/ceps/pull/142#discussion_r2748217893) was about the ambiguity of defining a 404 error on the `.sigs` endpoint as "this package does not exist", when channels that don't implement this CEP would return 404 on the `.sigs` endpoint for every package.

I think we should return 404 when the package does not exist, but define the meaning of 404 as:
- For channels implementing this CEP, 404 means the package does not exist
- For channels not implementing this CEP, `.sigs` always returns 404

and make it clear that clients should not use the `.sigs` endpoint to determine if a package exists or not

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+  "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+  described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+>
+> More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the
+  rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the
+  implementation.
+
+## Abstract
```
facutuesca, line 22: Maybe we should update the abstract, since the CEP defines both the endpoint to distribute the attestations but also the new configuration clients should use to consume them.

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+<tr><td> Requires </td><td> CEP 27 (Publish Attestation) </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+  "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+  described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+>
+> More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the
+  rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the
+  implementation.
```
facutuesca, line 20: any particular reason for adding this?

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+
+The field is the single discovery and integrity mechanism for attestation sidecars:
+
+1. **Discovery**: Clients learn from repodata alone whether a `.sigs` file exists, avoiding a network round-trip for packages without attestations.
+2. **Integrity**: Clients MUST verify that the fetched `.sigs` bytes hash to `sha256` before using the sidecar (see [Verification Workflow](#verification-workflow)). This prevents a mirror or intermediary from stripping or replacing attestations without detection.
+3. **Change detection**: When attestations are added after a package was first published, the channel publishes an updated `.sigs` file and updates the field. Mirrors and clients re-fetch the sidecar when the hash changes.
+4. **Resource bounds**: `size` allows clients to enforce a download limit before fetching. Clients MAY refuse to download sidecars larger than a locally configured limit; such a refusal MUST be handled like a retrieval failure (see [Configuration](#configuration)).
```
facutuesca: does this new `size` field add extra protection? AFAICT, the way the CEP says to use it is for clients to not download the `.sigs` file if the `size` field exceeds some limit.

What is it protecting against?

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+| `warn`    | warn    | warn              | warn                 |
+| `ignore`  | silent  | warn              | warn                 |
+
+"Fail" means the package MUST NOT be installed; "warn" means the client MUST log a warning and MAY continue. `ignore` accepts packages without attestations silently, but retrieval and verification failures are potential tampering signals and MUST NOT be silently swallowed: they are handled as under `warn`.
+
+#### `trusted_identities` matching
+
+Each entry is a pattern matched against the certificate identity (the SubjectAlternativeName of the Sigstore signing certificate). Matching is case-sensitive and literal, except that `*` matches any sequence of characters, including `/`.
+For example, `https://github.com/conda-forge/*` matches `https://github.com/conda-forge/numpy-feedstock/.github/workflows/build.yml@refs/heads/main` but not `https://github.com/conda-forge-evil/...`, because the literal prefix includes the trailing slash.
+A future revision of this CEP MAY extend entries to additionally pin the OIDC issuer.
```
facutuesca: the configuration should include the issuer in some form. `(issuerA, identityA)` is different from `(issuerB, identityA)`, and we should not allow both when `identityA` is specified.

Also, the paragraph below:

> **Sigstore verification** binds each attestation to a signing identity. An attacker cannot forge attestations for identities they do not control, but an attacker who controls the distribution path can substitute attestations signed by an identity they *do* control. The `trusted_identities` policy is what turns bundle verification into a guarantee about who produced the package.

The policy is a guarantee about who produced the package only if we also check the OIDC issuer.

cep-XXXX.md
wolfv: I think it can help against accidentally downloading too large blobs (which could be a DOS for improper configured clients). Also helps with displaying a nice progress bar :) But not a huge security win.

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+3. **Fetch the sidecar** from `<package_url>.sigs` and verify that its bytes hash to the `attestations.sha256` value from repodata. On a mismatch or fetch failure, the client SHOULD retry once with refreshed repodata, since the sidecar may have been updated concurrently. A persistent mismatch is a retrieval failure and MUST be handled per the `require` policy.
+4. **Verify each attestation** using the verification process defined by [CEP 27] or by the attestation's predicate type (see [Attestation Requirements](#attestation-requirements)).
+5. **Accept or reject** the package based on the verification results and the `require` policy.
+
+### Configuration
+
+Clients SHOULD support the following configuration options:
+
+```yaml
+# Example ~/.condarc configuration
```
travishathaway: I just want to comment here that this is most likely not how the configuration will actually end up looking in conda (see my experimental [conda-sigstore](https://github.com/travishathaway/conda-sigstore) for a more realistic example).

I do think this example is super useful for quickly understanding all the moving pieces of the configuration though. Maybe just remove the reference to `.condarc` so we don't confuse this with what conda clients will actually implement?

```suggestion
# Abstract example of Sigstore configuration:
```

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+
+| Package URL                                                                         | Attestation URL                                                                          |
+| ----------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
+| `https://conda.anaconda.org/conda-forge/linux-64/numpy-2.0.0-py312h1234567_0.conda` | `https://conda.anaconda.org/conda-forge/linux-64/numpy-2.0.0-py312h1234567_0.conda.sigs` |
+| `https://prefix.dev/my-channel/noarch/my-package-1.0.0-pyhd8ed1ab_0.conda`          | `https://prefix.dev/my-channel/noarch/my-package-1.0.0-pyhd8ed1ab_0.conda.sigs`          |
+| `https://example.com/channel/win-64/pkg-1.0-0.tar.bz2`                              | `https://example.com/channel/win-64/pkg-1.0-0.tar.bz2.sigs`                              |
+
+### Response Format
+
+The `.sigs` file MUST contain a JSON array of one or more [Sigstore bundles][Sigstore Bundle]. Each bundle represents one attestation for the package.
```
baszalmstra, line 71: Why an json array instead of a jsonl?

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+
+- `sha256` (REQUIRED): the SHA256 hash of the exact bytes of the `.sigs` file as served.
+- `size` (REQUIRED): the size of the `.sigs` file in bytes.
+
+The field MUST be present if and only if at least one attestation is available for the package. Its absence means the package has no attestations; clients MUST NOT treat an absent field as an error and SHOULD NOT request the `.sigs` URL in that case.
+
+The field is the single discovery and integrity mechanism for attestation sidecars:
+
+1. **Discovery**: Clients learn from repodata alone whether a `.sigs` file exists, avoiding a network round-trip for packages without attestations.
+2. **Integrity**: Clients MUST verify that the fetched `.sigs` bytes hash to `sha256` before using the sidecar (see [Verification Workflow](#verification-workflow)). This prevents a mirror or intermediary from stripping or replacing attestations without detection.
```
baszalmstra, line 125: If the mirror or intermediary has access to the channel they can also just change the repodata itself no?

baszalmstra (commented):
A big gap I see with this proposal is how this should be handled in lock-files. Given that `.sigs` is not immutable should we store the attestation sha256 in the lock-file as well. But what happens if on the server/mirror the .sigs are updated? Should we store the original .sigs in the lock-file?

cep-XXXX.md
facutuesca: Are we talking about maliciously crafted blobs, which are big on purpose to cause a DOS? Or legitimate normal-sized blobs, which might be too big for a resource-constrained client?.

In the former case, the channel could check and reject attestations that exceed a reasonable size. 

In the latter case, these are usually files in the tens of Kb. Is that still a concern?

facutuesca:
> A big gap I see with this proposal is how this should be handled in lock-files. Given that `.sigs` is not immutable should we store the attestation sha256 in the lock-file as well. But what happens if on the server/mirror the .sigs are updated? Should we store the original .sigs in the lock-file?

IMO lockfiles should lock the trusted identities, not the hashes of the attestations or `.sigs` responses. With regards to the gap, this proposal started as a spec to serve the attestations. Maybe lockfile changes could be done in a separate one?

cep-XXXX.md
wolfv: Both are fine with me. I think JSON array is slightly easier to parse with standard JSON decoders.

wolfv:
@conda/steering-council, the RFC period has officially opened for this CEP. It will last two weeks, so please place your comments **before 2026-09-02 EOD AoE**. If there are no blocking concerns, a vote will start immediately after the RFC period.

cep-XXXX.md
wolfv: yeah I think that is a different problem that we need to tackle in other ways. It also depends heavily on the channel. 

For example if the channel is `my-company` then I could state that I expect all publisher attestations to come from `github.com/my-company` 

At that point it becomes a lot more difficult for a mirror to fake this as they would have had to take over `my-company`. A mirror could totally omit the signatures though. We'd rely on the client configuration to enforce this trust.

cep-XXXX.md
wolfv: I kept it a JSON array ... I don't really mind either way but JSONL is IMO not _as_ easy to parse (still very reasonably easy though).

cep-XXXX.md
wolfv: I removed all this but we'll keep it in mind for the client implementation(s).

cep-XXXX.md
wolfv: Removed size again, changed the key to `attestations_sha256`.

cep-XXXX.md
wolfv: Done.

cep-XXXX.md
wolfv: I think this was copied from the template :)

Callek:
I’m wondering whether coordination of attestation updates is intentionally left out of scope here.

For a static channel, a package and its CEP 27 attestation may be handled in separate steps. `conda-index` generates the package’s repodata record from the package metadata, but does not currently discover a neighboring `.sigs` file or populate `attestations_sha256`. Something outside the normal indexing process therefore needs to create the sidecar and ensure that the corresponding field is included in repodata.

There is also a concurrency question when multiple attestations are added. Because the sidecar is an array and `attestations_sha256` identifies the entire serialized sidecar, each addition produces a new aggregate sidecar. If two additions are processed concurrently from the same existing sidecar, they could produce competing successors, with whichever update wins potentially omitting the other attestation. The immutable URLs prevent cache races for published revisions, but do not by themselves prevent this kind of lost update.

The CEP specifies the order in which a completed replacement sidecar and its repodata update are published, but not how additions are serialized or merged before that point. Is that deliberately left entirely to channel implementations? If so, would it be worth stating that explicitly? Alternatively, does the specification need to define any consistency guarantee—for example, that concurrently accepted attestations must not be lost—without prescribing a particular upload API or storage implementation?

wolfv:
Indeed, it makes sense to clarify a serialization order to keep `sha256` hashes stable. We could order them by logIndex, timestamp, or some other property.

Concurrency issues / attestation append issues are indeed left to the indexing implementation. E.g. for rattler index we would probably suggest to have a `sigs` folder next to the packages, where attestations are stored with `<package-name>.N.sig.json` which are then used in the indexing step (`N` for the index of the attestation).

However, on a real server this will probably look a bit different.

wolfv:
@conda/steering-council

The RFC period has ended, and I’m opening the vote on **Distribution of Sigstore Attestations for Conda Packages**.

This vote falls under the **[Enhancement Proposal Approval](https://github.com/conda/governance#enhancement-proposal-approval)** policy. It requires participation from at least **8 of the 13** active Steering Council members (60% quorum), with at least **60% of non-abstaining votes** voting `yes` to pass. Abstentions count toward quorum.

To vote, please select exactly one checkbox under your name below, or leave a comment with `yes`, `no`, or `abstain`. Questions and review comments are welcome.

**This vote will end on 2026-09-16, End of Day, [Anywhere on Earth](https://time.is/Anywhere_on_Earth) (AoE).**

## Votes

@baszalmstra (Bas Zalmstra)
- [x]  yes
- [ ]  no
- [ ]  abstain

@chenghlee (Cheng H. Lee)
- [x]  yes
- [ ]  no
- [ ]  abstain

@ocefpaf (Filipe Fernandes)
- [x]  yes
- [ ]  no
- [ ]  abstain

@Hind-M (Hind Montassif)
- [x]  yes
- [ ]  no
- [ ]  abstain

- [ ]  yes
- [ ]  no
- [x]  abstain

@jezdez (Jannis Leidel)
- [x]  yes
- [ ]  no
- [ ]  abstain

@jakirkham (John Kirkham)
- [ ]  yes
- [ ]  no
- [ ]  abstain

- [x]  yes
- [ ]  no
- [ ]  abstain

- [x]  yes
- [ ]  no
- [ ]  abstain

@msarahan (Michael Sarahan)
- [ ]  yes
- [ ]  no
- [ ]  abstain

@pavelzw (Pavel Zwerschke)
- [x]  yes
- [ ]  no
- [ ]  abstain

@xhochy (Uwe Korn)
- [x]  yes
- [ ]  no
- [ ]  abstain

@wolfv (Wolf Vollprecht)
- [x]  yes
- [ ]  no
- [ ]  abstain

travishathaway (approved):

wolfv:
Final reminder @msarahan @jakirkham

jaimergp:

* SC size: 13
* Votes: 11 (84.6% of SC)
  * Yes: 10 (100% of votes)
  * No: 0
  * Abstain: 1
  
@conda/steering-council, this will be minted as CEP 50.

jaimergp (approved):

PR #143, opened 2025-12-19
Adopt API Conventions in Conda Code, Exempt non-API Code From Deprecation Schedule

dholth:
## Checklist for submitter

- [x] I am submitting a new CEP: [Changes to Conda Deprecation Schedule](https://github.com/dholth/ceps/blob/change-deprecation/cep-0099.md).

### DIFF
--- a/cep-0099.md
+++ b/cep-0099.md
@@ -2,7 +2,7 @@
 
 <table>
 <tr><td> Title </td><td> Adopt API Conventions in Conda Code, Exempt non-API Code From Deprecation Schedule </td>
-<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Status </td><td> Proposed </td></tr>
 <tr><td> Author(s) </td><td> Daniel Holth &lt;dholth@anaconda.com&gt;</td></tr>
 <tr><td> Created </td><td> Dec 16, 2025</td></tr>
 <tr><td> Updated </td><td> Feb 25, 2025</td></tr>
--- a/cep-0099.md
+++ b/cep-0099.md
@@ -34,7 +34,7 @@ details in conda.
 
 ## Specification
 
-This CEP proposes that conda adopt the [Python Library
+This CEP proposes adopting the [Python Library
 Interface](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols)
 convention for new code.
 
@@ -99,6 +99,11 @@ from conda._internal.frobnicate import Frobnicate as Frobnicate
 Other programmers, and type checkers, know they can use the `Frobnicate` class
 but should not reference `_value` or `helper1`, `helper2`. In turn, the
 developer has more freedom to refactor.
+
+## Counterexample
+
+Conda should maintain backwards compatibility when doing so does not cause a
+significant maintenance burden.
 
 ## Alternatives
 
--- a/cep-0099.md
+++ b/cep-0099.md
@@ -17,7 +17,7 @@ schedule, demarcated by following Python library conventions.
 
 ## Motivation
 
-[CEP-9](https://github.com/conda/ceps/blob/main/cep-0009.md) describes a
+[CEP 9](https://github.com/conda/ceps/blob/main/cep-0009.md) describes a
 deprecation schedule to warn about removals from the codebase but does not
 describe any part of conda's codebase that is exempt from the deprecation
 schedule. This 10-15 month deprecation cycle, applied to all symbols in the
@@ -29,7 +29,7 @@ for new code. Instead, most new code is not called from outside conda and should
 be allowed to change as needed between conda releases.
 
 By applying Python type-checker API conventions to codebases that fall under
-CEP-9, this proposal aims to create a place for experiments and implementation
+CEP 9, this proposal aims to create a place for experiments and implementation
 details in conda.
 
 ## Specification
--- a/cep-0099.md
+++ b/cep-0099.md
@@ -112,6 +112,11 @@ significant maintenance burden.
 * __Release fast-moving code as libraries__. A good idea but not suitable for
   code that is tightly coupled to conda.
 
+## Clarification
+
+Although non-Python codebases may have their own deprecation and API policies,
+CEP-9 and this CEP shall only apply to Python codebases.
+
 ## Prior Art
 
 `conda` has a `conda.api` module. The module provides a limited set of wrapper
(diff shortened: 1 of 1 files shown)


cep-0099.md
```diff
@@ -0,0 +1,77 @@
...
+convention for new code. In this convention underscore-prefixed symbols are
+considered private and are not part of the library's interface but `__all__`
+overrides these rules. Symbols may be re-exported into the public interface with
+forms like `import X as X`. Python type checkers understand these conventions
+and can let developers know if they are using conda's public interface.
+
+Newly added private code would not be subject to the deprecation policy. This
+code could change as necessary instead of after a deprecation cycle.
+
+Private portions of newly-added public modules, i.e. underscore-prefixed symbols
```
soapy1: Is there a convention for when (and at what level) we might want to introduce `__all__` into a module or not?

cep-0099.md
```diff
@@ -0,0 +1,77 @@
...
+details in conda.
+
+## Specification
+
+This CEP proposes that conda adopt the [Python Library
+Interface](https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols)
+convention for new code. In this convention underscore-prefixed symbols are
+considered private and are not part of the library's interface but `__all__`
+overrides these rules. Symbols may be re-exported into the public interface with
+forms like `import X as X`. Python type checkers understand these conventions
```
soapy1: I think this can be more clearly communicated as a list of rules oppose to a couple long sentences. Would you be open to quoting the rules specified in the link:


```suggestion
convention for new code. 

> * Symbols whose names begin with an underscore (but are not dunder names) are considered private.
> * Imported symbols are considered private by default. A fixed set of [import forms](https://typing.python.org/en/latest/spec/distributing.html#import-conventions) re-export imported> symbols.
> * A module can expose an __all__ symbol at the module level that provides a list of names that are considered part of the interface. This overrides all other rules above, allowing imported symbols or symbols whose names begin with an underscore to be included in the interface.
> * Local variables within a function (including nested functions) are always considered private.

Python type checkers understand these conventions
```

soapy1 (commented):
I like this idea!

cep-0099.md
dholth: I would audit existing code especially projects under https://github.com/conda, add those, and also provide deliberate API exports for new code but not necessarily worry about `__all__` when conda imports from itself.

travishathaway (approved):
I think this is a good idea and will help us more clearly delineate parts of code that are meant to be public and parts that are meant to be private.

I only wish that the conventions around API design in the Python community were a bit a stronger so documents like this wouldn't be necessary. Regardless, I think this will be not only useful for conda itself but all projects in the conda organization in GitHub.

dholth:
@conda/steering-council please vote on this CEP

beckermr (dismissed):
We need to couple this PR with PRs on CEP-9 covered projects that make sure the conventions here don't move things which clearly should be public to private.

Also, I find the symbol reexport rules confusing. I'd rather have a simpler local convention where a symbol anywhere is private if any part of its full import path to its original definition starts with a single `_` and is public otherwise. However, this last comment is something to take up with a PEP, not here.

cep-0099.md
```diff
@@ -0,0 +1,140 @@
...
+  supported by type checkers.
+* __Release fast-moving code as libraries__. A good idea but not suitable for
+  code that is tightly coupled to conda.
+* [Retract CEP 9](https://github.com/conda/ceps/pull/152), move policy decisions
+  to conda developer team.
+
+## Clarification
+
+Although non-Python codebases may have their own deprecation and API policies,
+CEP 9 and this CEP shall only apply to Python codebases.
```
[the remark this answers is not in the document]
dholth: The first half of the sentence "non-Python codebases may have their own deprecation and API policies" is meant to cover that case; Rust has strong mechanisms for separating interface (API) from implementation and Python underscores, symbol re-exports are not meaningful there.
This CEP is meant to be limited in scope for simplicity.

ryanskeith (approved):
I like this CEP.  It will allow developers to iterate on the internals without the mental tax of deprecation in a lot of cases.

cep-0099.md
```diff
@@ -0,0 +1,140 @@
...
+[CEP 9](https://github.com/conda/ceps/blob/main/cep-0009.md) describes a
+deprecation schedule to warn about removals from the codebase but does not
+describe any part of conda's codebase that is exempt from the deprecation
+schedule. This 10-15 month deprecation cycle, applied to all symbols in the
+codebase, is an unnecessary barrier to change.
+
+At the time it was not unreasonable to assume that any part of conda could have
+been used as a library by unknown dependents, but this is not a good assumption
+for new code. Instead, most new code is not called from outside conda and should
+be allowed to change as needed between conda releases.
```
jezdez: This section is ignoring [Hyrum's law](https://www.hyrumslaw.com/) and assumes that a declaration of "private APIs" prevents future adoption of these APIs in implicit contracts. It resets conda back to a state in which downstream users have to carefully read the symbol names and module structure to understand what to rely on, instead of being informed by it via the deprecation system.

cep-0099.md
```diff
@@ -0,0 +1,140 @@
...
+The developer thinks the new class should be an API, and creates a module
+`conda.frobnicate` with
+
+```python
+from conda._internal.frobnicate import Frobnicate as Frobnicate
+```
+
+Other programmers, and type checkers, know they can use the `Frobnicate` class
+but should not reference `_value` or `helper1`, `helper2`. In turn, the
+developer has more freedom to refactor.
```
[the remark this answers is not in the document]
dholth: Conda API users cannot violate the CEP. If they find an internal feature that should be API, then they can submit a PR that re-exports that feature.

cep-0099.md
```diff
@@ -0,0 +1,140 @@
...
+and can let developers know if they are using conda's public interface.
+
+Newly added private code would not be subject to the deprecation policy. This
+code could change as necessary instead of after a deprecation cycle.
+
+Private portions of newly-added public modules, i.e. underscore-prefixed symbols
+and class members or anything not in an `__all__` list, would also be permitted
+to change at any time.
+
+This proposal allows conda maintainers to add a top-level `_conda` for code that
```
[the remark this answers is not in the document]
jezdez: I don't understand, that's not a pattern I've seen elsewhere, you're saying there could be a `conda` and a `_conda` package in the repo?

jaimergp:
Thanks everyone. Five votes, all of them abstentions. The CEP has not passed. I'll close here.

There was some additional discussion about the result in last week's meeting. [See notes](https://github.com/conda/governance/pull/363).

PR #145, opened 2025-12-23
Add CEPs for Repodata Wheel Support

danyeaw:
Updated version to replace #144, developed with @travishathaway.

This CEP outlines how native support for pure Python wheel packages could be achieved by adding support for them in repodata. When implemented, conda clients will be able to seamlessly install conda packages and pure Python wheels from enabled channels.

## Checklist for submitter

- [X] I am submitting a new CEP: [Repodata Wheel Support](https://github.com/danyeaw/ceps/blob/wheel-repodata/cep-XXXX.md).

### DIFF
--- a/cep-XXX1.md
+++ b/cep-XXX1.md
@@ -24,12 +24,12 @@ described in [RFC2119][RFC2119] when, and only when, they appear in all capitals
 This CEP specifies how channel operators publish pure Python wheel packages in conda's package index (repodata): the `whl` member under `v{revision}`, record fields, publisher-side dependency conversion from wheel `METADATA`, and validation rules.
 It does not define conda client solver, download, or install-time behavior; those are specified in [CEP XXX2 – Wheel conda client support][cep-xxx2].
 Motivation, ecosystem history, and rejected alternatives are in [CEP XXX0 – Wheel support in conda (overview)][cep-xxx0].
-We limit the scope to pure Python wheels to avoid platform-specific binary compatibility issues; even so, publishing wheels in repodata greatly increases the packages channels can expose without building conda recipes for every case.
+We limit the scope to pure Python wheels to avoid platform-specific binary compatibility issues. Even so, publishing wheels in repodata greatly increases the packages channels can expose without building conda recipes for every case.
 
 ## Relationship to companion CEPs
 
 - **[CEP XXX0 – Wheel support in conda (overview)][cep-xxx0]** — Why native wheel support, historical context, and discussion of rejected ideas (informative).
-- **[CEP XXX2 – Wheel conda client support][cep-xxx2]** — Normative client behavior: loading `whl` into the solve, exclusivity and preference, download, and installation into the prefix.
+- **[CEP XXX2 – Wheel conda client support][cep-xxx2]** — Normative client behavior: loading `whl` into the solve with other repodata records, download, and installation.
 
 ## Specification
 
@@ -122,7 +122,9 @@ Wheels that fail any of these checks MUST NOT be added to `whl`.
 
 ### Dependency conversion
 
-Wheel dependencies (from METADATA file's Requires-Dist entries) MUST be converted to conda format following these rules:
+Channel operators MUST convert Python dependency declarations—`Requires-Dist`, `Requires-Python`, and extras—into conda `depends`, **`extra_depends`** when applicable, and related fields, following the rules below.
+
+Those declarations MAY be read from the wheel's embedded **`METADATA`** and/or from the [**PyPI JSON API**][pypi-json-api] release payload for the same **project** and **version** as the wheel being indexed. Using the API can simplify batch indexing and mirrors. Regardless of source, the same conversion rules and interoperable minimum in [PEP 508 marker translation guidance](#pep-508-marker-translation-guidance) apply.
 
 - **Package names:** Names per [CEP 26][cep-26] and match existing conda-forge package names where they exist
 - **Version specifiers:** Map PEP 440 [version specifiers][version-specifiers] to conda format:
@@ -161,7 +163,7 @@ Resulting conda record:
 
 ### PEP 508 marker translation guidance
 
-This subsection layers **interoperable expectations** for the wire format, **non-normative** reference material for richer mappings, and **lossy translation** policy. It does not duplicate a full PEP 508 variable matrix in this CEP; maintaining that table is left to living documentation and tooling (see below).
+This subsection layers interoperable expectations for the repodata format, non-normative reference material for richer mappings, and a general translation policy.
 
 #### Interoperable minimum
 
@@ -180,7 +182,7 @@ Per-variable translation of other PEP 508 markers (for example `sys_platform`, `
 #### Lossy translation
 
 When a channel cannot encode a PEP 508 dimension in repodata (for example no virtual package for an architecture string, or unsupported boolean branches), the operator MAY **omit** the condition, **simplify** the requirement, or **widen** the dependency (for example listing a package in `depends` without `when=` so it applies in more environments than PEP 508 would).
-Channel operators SHOULD document their policy for such cases (including any conservative “include unconditionally” behavior). They MUST NOT claim that published repodata reproduces **full** PEP 508 semantics when a lossy transform was applied.
+Channel operators SHOULD document their policy for such cases (including any conservative “include unconditionally” behavior).
 
 ### Handling conditional dependencies and extras
 
@@ -188,9 +190,9 @@ Wheel `depends` entries that encode PEP 508 environment markers MUST use conditi
 
 PyPI **extras** (optional dependency groups) MUST be represented on the wheel repodata record using an **`extra_depends`** object (a mapping from extra name to lists of dependency strings) as specified in [PR 111][pr-111]. A default install uses only `depends` and MUST NOT union in `extra_depends` entries unless the user selects optional groups (for example with `extras=` in `MatchSpec` as described in [PR 111][pr-111]).
 
-### Solver behavior and package preference
+### Solver behavior
 
-Normative rules for how conda clients treat `whl` records during dependency resolution (candidate pool, exclusivity, default preference for conda vs wheel, and user overrides) are defined in [CEP XXX2 – Wheel conda client support][cep-xxx2].
+Normative rules for how conda clients treat `whl` records during dependency resolution (same candidate pool as other repodata records; no CEP-mandated default between conda builds and wheels) are defined in [CEP XXX2 – Wheel conda client support][cep-xxx2].
 This CEP only requires that published `whl` records are valid [repodata records][repodata-record-schema] so that clients can consume them alongside `packages` and `packages.conda`.
 
 ### Limitations
@@ -201,8 +203,9 @@ This CEP has the following known limitations:
 2. **Environment markers (publisher defaults):** Default conversion rules in this CEP focus on Python-version markers on `Requires-Dist`, mapped to `when=` per [PR 111][pr-111]. Other markers (for example `sys_platform`) are out of scope for those default publisher rules and are not required to be converted into repodata here; see [PEP 508 marker translation guidance](#pep-508-marker-translation-guidance).
 How conda clients evaluate `when=` and optional groups at solve time (including environment context) is specified in [CEP XXX2][cep-xxx2] together with [PR 111][pr-111].
 3. **Conditionals and extras:** Normative syntax and record fields for `when=` on `depends` and for optional groups in **`extra_depends`** are specified in [PR 111][pr-111], on which this repodata CEP and [CEP XXX2][cep-xxx2] rely for PyPI-aligned conditionals and extras in published records and at client solve time, respectively.
-4. **Repodata size:** Supporting a significant portion of pure Python packages from PyPI (potentially hundreds of thousands of packages with multiple versions each) will substantially increase repodata size. Channels adopting wheel support at scale will need to implement sharded repodata ([CEP 16][cep-16]) to maintain acceptable performance. Alternatively, channels may choose to curate a subset of popular or requested packages rather than mirroring all of PyPI.
-5. **PyPI package deletion:** PyPI allows package maintainers to delete packages (as opposed to just yanking them), which can break locked environments that reference those packages. Channels using external PyPI URLs directly are subject to this risk. For production use and reproducible environments, channels MAY mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs.
+4. **Repodata size:** Supporting a significant portion of pure Python packages from PyPI (potentially hundreds of thousands of packages with multiple versions each) will substantially increase repodata size. Channels adopting wheel support at scale SHOULD implement sharded repodata ([CEP 16][cep-16]) to maintain acceptable performance.
+5. **PyPI package deletion:** PyPI allows package maintainers to delete releases (as opposed to only yanking them). Releases may also be removed when classified as **malicious** or otherwise pulled by PyPI administrators. Any removal may break locked environments that reference those artifacts.
+Channels using external PyPI URLs directly are subject to this risk. For production use and reproducible environments, channels MAY mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs.
 
 ## Rationale
 
@@ -226,13 +229,14 @@ Several factors can cause wheel names to differ from conda-style names:
 
 ### For conda clients
 
-Client behavior (loading `whl` into the solve, preference and exclusivity, download, and installation into the environment prefix) is specified in [CEP XXX2 – Wheel conda client support][cep-xxx2]. Channel operators need not implement that CEP; they only publish repodata that conforms to this document.
+Client behavior (loading `whl` into the solve with other repodata records, download, and installation into the environment prefix) is specified in [CEP XXX2 – Wheel conda client support][cep-xxx2]. Channel operators need not implement that CEP; they only publish repodata that conforms to this document.
 
 ### For channel operators
 
 Channel operators adding wheel support SHOULD:
 
 - Implement validation to ensure only pure Python wheels are included
+- When using the [PyPI JSON API][pypi-json-api] for dependency metadata, consider comparing against the wheel's `METADATA` and fall back (or merge) when the API response is incomplete relative to the artifact
 - Ensure that dependencies are solvable, including that compiled dependencies exist on the conda channel declared as `info.channel_relations.base` when wheels depend on conda packages from that stack
 - Maintain a mapping of PyPI to conda-style names for their channel, or declare `channel_relations.base` so clients load a base channel that defines those mappings (see [Naming standard and channel mapping](#naming-standard-and-channel-mapping) and [Channel relations in repodata][cep-channel-relations])
 - Consider automation to keep the repodata up to date with newer releases on PyPI
@@ -424,6 +428,7 @@ The package record expresses the conditional with `when=` so `typing_extensions`
 - [PR 151 – URL field for package records][pr-151]
 - [PR 111 – Conditional dependencies and optional groups][pr-111]
 - [conda-pupa][conda-pupa]
+- [PyPI JSON API documentation][pypi-json-api]
 
 ## Copyright
 
@@ -443,5 +448,6 @@ All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdo
 [pr-151]: https://github.com/conda/ceps/pull/151
 [pr-111]: https://github.com/conda/ceps/pull/111
 [conda-pupa]: https://github.com/dholth/conda-pupa
+[pypi-json-api]: https://docs.pypi.org/api/json/
 [cep-xxx0]: cep-XXX0.md
 [cep-xxx2]: cep-XXX2.md
--- a/cep-XXX1.md
+++ b/cep-XXX1.md
@@ -8,10 +8,11 @@
   Travis Hathaway &lt;travis.j.hathaway@gmail.com&gt;
 </td></tr>
 <tr><td> Created </td><td> Dec 23, 2025</td></tr>
-<tr><td> Updated </td><td> Apr 16, 2026</td></tr>
+<tr><td> Updated </td><td> Apr 24, 2026</td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/145 </td></tr>
 <tr><td> Implementation </td><td> TBD </td></tr>
-<tr><td> Requires </td><td>https://github.com/conda/ceps/pull/151 https://github.com/conda/ceps/pull/146 https://github.com/conda/ceps/pull/155 https://github.com/conda/ceps/pull/111</td></tr>
+<tr><td> Requires </td><td> https://github.com/conda/ceps/pull/151 https://github.com/conda/ceps/pull/146 https://github.com/conda/ceps/pull/155 https://github.com/conda/ceps/pull/111 [CEP XXX2 (Wheel conda client support)](cep-XXX2.md) </td></tr>
+<tr><td> See also </td><td> [CEP XXX0 – Wheel support in conda (overview)](cep-XXX0.md) </td></tr>
 </table>
 
 > The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
@@ -20,27 +21,15 @@ described in [RFC2119][RFC2119] when, and only when, they appear in all capitals
 
 ## Abstract
 
-This CEP outlines how native support for pure Python wheel packages will be achieved by adding support for them to conda's package index (repodata). When implemented, conda clients will be able to seamlessly install conda packages and pure Python wheels from enabled channels.
-We explicitly limit the scope of this CEP to pure Python wheels to avoid platform-specific binary compatibility issues, but even with this limitation, we dramatically expand the number of available packages for users in the conda ecosystem.
+This CEP specifies how channel operators publish pure Python wheel packages in conda's package index (repodata): the `whl` member under `v{revision}`, record fields, publisher-side dependency conversion from wheel `METADATA`, and validation rules.
+It does not define conda client solver, download, or install-time behavior; those are specified in [CEP XXX2 – Wheel conda client support][cep-xxx2].
+Motivation, ecosystem history, and rejected alternatives are in [CEP XXX0 – Wheel support in conda (overview)][cep-xxx0].
+We limit the scope to pure Python wheels to avoid platform-specific binary compatibility issues; even so, publishing wheels in repodata greatly increases the packages channels can expose without building conda recipes for every case.
 
-## Motivation
+## Relationship to companion CEPs
 
-While conda remains a language-agnostic packaging distribution, installing packages for use with Python represents the majority of conda usage. Users frequently encounter packages only available as wheels on PyPI, forcing them to use hybrid workflows that mix conda and PyPI installations. This approach creates several problems:
-
-- Mixing of conda and pip or uv can result in overwritten files and broken environments
-- Users must understand two package managers, their interaction points, and which of their dependencies are available from which  ecosystem
-- Where there is support for mixing environments, it requires multiple solves which reduces overall package installation performance
-
-By adding native support for pure Python wheels to repodata, conda clients can:
-
-- Eliminate the cognitive burden of managing two package managers
-- Provide users with transparent access to the broader Python ecosystem
-- Maintain environment consistency and reproducibility
-- Resolve dependencies across conda and PyPI packages in a single solve
-- Fill gaps in conda package availability for simpler pure-Python packages
-- Reduce the maintenance burden for straightforward pure-Python packages that don't require metadata modifications
-
-Note that this CEP does not eliminate the need for conda recipes entirely. Many packages will still require proper conda feedstocks due to metadata differences, dependency corrections, or other ecosystem-specific needs. This CEP provides an additional tool for handling simpler pure-Python cases more efficiently.
+- **[CEP XXX0 – Wheel support in conda (overview)][cep-xxx0]** — Why native wheel support, historical context, and discussion of rejected ideas (informative).
+- **[CEP XXX2 – Wheel conda client support][cep-xxx2]** — Normative client behavior: loading `whl` into the solve, exclusivity and preference, download, and installation into the prefix.
 
 ## Specification
 
@@ -145,8 +134,6 @@ Wheel dependencies (from METADATA file's Requires-Dist entries) MUST be converte
 - **Python version requirements:** Convert Requires-Python to explicit python dependency
 - **Environment markers:** Map Python-version markers on `Requires-Dist` to conditional `MatchSpec` `when=` as specified in [PR 111][pr-111]. Other markers are out of scope for the default conversion rules in this CEP (see [Limitations](#limitations)).
 
-For a non-normative, per-variable description of how **conda-pypi** maps PEP 508 markers to `when=` fragments and **`extra_depends`** for optional groups, see [PEP 508 marker conversion][conda-pypi-marker-conversion] in the conda-pypi documentation.
-
 Example conversion:
 
 ```text
@@ -172,6 +159,29 @@ Resulting conda record:
 }
 ```
 
+### PEP 508 marker translation guidance
+
+This subsection layers **interoperable expectations** for the wire format, **non-normative** reference material for richer mappings, and **lossy translation** policy. It does not duplicate a full PEP 508 variable matrix in this CEP; maintaining that table is left to living documentation and tooling (see below).
+
+#### Interoperable minimum
+
+For published `whl` records, the following conversions are the **normative floor** aligned with [Dependency conversion](#dependency-conversion) and [PR 111][pr-111]:
+
+- **`Requires-Python`:** MUST appear as an explicit `python` dependency in `depends`.
+- **Python-version conditions on `Requires-Dist`:** MUST be represented with conditional `MatchSpec` `when=` on the relevant dependency strings per [PR 111][pr-111].
+- **`extra` (optional groups):** MUST be represented with **`extra_depends`** and optional-group selection per [PR 111][pr-111], not only as opaque marker text on `depends`.
+
+Channels and tools MAY apply additional, stricter mappings where they can express them in repodata.
+
+#### Non-normative extended mappings
+
+Per-variable translation of other PEP 508 markers (for example `sys_platform`, `platform_system`) to `when=` fragments, virtual packages, or omissions is **not** fully specified in this CEP. For a non-normative, per-variable description—including how **conda-pypi** maps markers to `when=` and **`extra_depends`**—see [PEP 508 marker conversion][conda-pypi-marker-conversion] in the conda-pypi documentation.
+
+#### Lossy translation
+
+When a channel cannot encode a PEP 508 dimension in repodata (for example no virtual package for an architecture string, or unsupported boolean branches), the operator MAY **omit** the condition, **simplify** the requirement, or **widen** the dependency (for example listing a package in `depends` without `when=` so it applies in more environments than PEP 508 would).
+Channel operators SHOULD document their policy for such cases (including any conservative “include unconditionally” behavior). They MUST NOT claim that published repodata reproduces **full** PEP 508 semantics when a lossy transform was applied.
+
 ### Handling conditional dependencies and extras
 
 Wheel `depends` entries that encode PEP 508 environment markers MUST use conditional `MatchSpec` syntax with `when=` as specified in [PR 111][pr-111].
@@ -180,29 +190,17 @@ PyPI **extras** (optional dependency groups) MUST be represented on the wheel re
 
 ### Solver behavior and package preference
 
-#### Dependency resolution
-
-Solvers MUST treat pure Python wheels as valid package candidates during dependency resolution with these constraints:
-
-- **Exclusivity:** Solvers MUST NOT install both a conda package and wheel for the same package name.
-- **Dependency satisfaction:** When a wheel is selected, its `depends` list MUST be satisfied like any conda package. Dependencies in the `depends` list MAY be satisfied by either wheels or conda packages.
-- **Platform matching:** Since all wheels in `whl` are pure Python (noarch), no platform filtering is needed.
-
-#### User control of precedence
-
-By default, channels with conda packages MUST prefer conda packages when they are available. Users MAY override default precedence through:
-
-- Channel priority configuration (prefer channels with wheels).
-- Explicit wheel requests through `<channel name>::<package name>` syntax.
-- Explicit configuration in the client itself (e.g. `prefer_conda` or `prefer_wheel`)
+Normative rules for how conda clients treat `whl` records during dependency resolution (candidate pool, exclusivity, default preference for conda vs wheel, and user overrides) are defined in [CEP XXX2 – Wheel conda client support][cep-xxx2].
+This CEP only requires that published `whl` records are valid [repodata records][repodata-record-schema] so that clients can consume them alongside `packages` and `packages.conda`.
 
 ### Limitations
 
 This CEP has the following known limitations:
 
 1. **Pure Python only:** This CEP explicitly does not address wheels with binary extensions, which require platform-specific compatibility guarantees beyond the current scope. Conda’s strength is binary compatibility, so using conda packages may be the optimal solution.
-2. **Environment markers:** This CEP’s default conversion rules focus on Python-version markers mapped to `when=` per [PR 111][pr-111]. Other markers (for example `sys_platform`) are out of scope for those default rules and are not converted here, though some MAY be expressible with `when=` once clients implement [PR 111][pr-111].
-3. **Conditionals and extras:** Normative syntax and record fields for `when=` on `depends` and for optional groups in **`extra_depends`** are specified in [PR 111][pr-111], which this wheel CEP depends on for PyPI-aligned conditionals and extras.
+2. **Environment markers (publisher defaults):** Default conversion rules in this CEP focus on Python-version markers on `Requires-Dist`, mapped to `when=` per [PR 111][pr-111]. Other markers (for example `sys_platform`) are out of scope for those default publisher rules and are not required to be converted into repodata here; see [PEP 508 marker translation guidance](#pep-508-marker-translation-guidance).
+How conda clients evaluate `when=` and optional groups at solve time (including environment context) is specified in [CEP XXX2][cep-xxx2] together with [PR 111][pr-111].
+3. **Conditionals and extras:** Normative syntax and record fields for `when=` on `depends` and for optional groups in **`extra_depends`** are specified in [PR 111][pr-111], on which this repodata CEP and [CEP XXX2][cep-xxx2] rely for PyPI-aligned conditionals and extras in published records and at client solve time, respectively.
 4. **Repodata size:** Supporting a significant portion of pure Python packages from PyPI (potentially hundreds of thousands of packages with multiple versions each) will substantially increase repodata size. Channels adopting wheel support at scale will need to implement sharded repodata ([CEP 16][cep-16]) to maintain acceptable performance. Alternatively, channels may choose to curate a subset of popular or requested packages rather than mirroring all of PyPI.
 5. **PyPI package deletion:** PyPI allows package maintainers to delete packages (as opposed to just yanking them), which can break locked environments that reference those packages. Channels using external PyPI URLs directly are subject to this risk. For production use and reproducible environments, channels MAY mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs.
 
@@ -228,16 +226,7 @@ Several factors can cause wheel names to differ from conda-style names:
 
 ### For conda clients
 
-Clients implementing this CEP SHOULD:
-
-- Parse the `whl` mapping inside each supported `v{revision}` object alongside existing `packages` /
-  `packages.conda` sections
-- Apply the same filtering and preference logic used for conda packages
-- Extract wheel metadata during solving to populate dependency information
-- Evaluate `when=` on wheel `depends` entries per [PR 111][pr-111]
-- Honor **`extra_depends`** when the user selects optional groups (for example with `extras=` in `MatchSpec`) per [PR 111][pr-111]
-- Treat wheel records that use `when=` on `depends` or **`extra_depends`** metadata as part of the same `v{revision}` repodata payload as other new-syntax records, per the [backwards-compatible repodata update strategy](https://github.com/conda/ceps/pull/146)
-- Provide the ability to natively install wheels or on-the-fly convert wheels to conda packages for installation
+Client behavior (loading `whl` into the solve, preference and exclusivity, download, and installation into the environment prefix) is specified in [CEP XXX2 – Wheel conda client support][cep-xxx2]. Channel operators need not implement that CEP; they only publish repodata that conforms to this document.
 
 ### For channel operators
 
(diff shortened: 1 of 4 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,561 @@
...
+
+Prefix.dev starts a barebones pip implemented in Rust to resolve and install PyPI dependencies with Pixi.
+
+### Pixi Integrates with uv (Jan 2024)
+
+Pixi changes course to use uv directly instead of rip, which unlocks features like editable installations, and git and path dependencies.
+
+### conda-pupa creates on-the-fly conversion plugin (July 2024)
+
+Daniel Holth creates a conda plugin that supports on-the-fly conversion of conda packages to wheels in a local channel.
```
jjhelmus: This seems to be leaving out some history

cep-XXXX.md
```diff
@@ -0,0 +1,561 @@
...
+
+Jaime Rodríguez-Guerra starts development on the conda-pypi plugin aimed to improve conda and PyPI interoperability.
+
+### Rip development starts (2023)
+
+Prefix.dev starts a barebones pip implemented in Rust to resolve and install PyPI dependencies with Pixi.
+
+### Pixi Integrates with uv (Jan 2024)
+
+Pixi changes course to use uv directly instead of rip, which unlocks features like editable installations, and git and path dependencies.
```
pavelzw: these are all now available for conda-only workflows through pixi-build.

danyeaw: Hey @pavelzw, thanks so much for the feedback! Do you think we are missing a milestone in our brief history section? This pixi build feature is more about building path/git for conda packages than installing wheels isn't it?

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+level keys:
+
+- repodata_version
+- info
+- packages
+- packages.conda
+- removed
+- signatures
+
+This CEP proposes the addition of a new `packages.whl` section to account for the wheel format. This key points to a mapping that MUST contain [repodata record][repodata-record-schema] objects. The key of this mapping MUST be the wheel file name, and the contents of the package record MUST contain metadata related to that wheel.
```
travishathaway:
```suggestion
This CEP proposes the addition of a new `packages.whl` section to account for the wheel format. This key points to a mapping that MUST contain [repodata record][repodata-record-schema] objects. The key of this mapping MUST follow the specification described below in "Key naming requirements".
```
Just making sure this part of the specification agrees with the rest.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+
+### Record structure and naming convention
+
+To support sparse repodata processing and maintain compatibility with conda's existing infrastructure, wheel records MUST use conda-style naming conventions rather than wheel filenames as keys. This allows solvers to efficiently filter packages by name without parsing the full package record.
+
+### Key naming requirements
+
+The key for each entry in `packages.whl` must follow this format:
```
travishathaway:
```suggestion
The key for each entry in `packages.whl` MUST follow this format:
```
Style consistency.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+- Channel operators adding wheels to repodata MUST determine the appropriate conda-style name by:
+  - Checking if the package already exists in conda-forge and using that exact name
+  - If not in conda-forge, applying conda naming conventions per [CEP 26 - Identifying Packages and Channels in the conda Ecosystem][cep-26]
+  - Documenting any naming mappings for their channel
+
+### Wheel download URLs
+
+This CEP introduces a new optional `artifact_url` field in package records to specify download locations for individual packages.
+
+> Note for this draft: The `artifact_url` field could also be added as a separate CEP to allow it for other record types.
```
travishathaway: I think that would actually be a good idea to avoid asymmetries in package record specifications. Either that or we explicitly mention this new field is for all package record types.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+- Mixed sources within the same repodata file
+
+### Wheel-Specific Record Values
+
+When populating repodata records for pure Python wheels:
+
+- `build: MUST be py3_0
+- `build_number`: MUST be 0 for the initial addition of a wheel version. MAY be incremented for subsequent rebuilds of the same - wheel version (e.g., to correct dependencies or metadata)
+- `fn`: MUST be the wheel filename (e.g., package-1.0.0-py3-none-any.whl)
+- `subdir`: MUST be noarch
```
travishathaway:
```suggestion
- `subdir`: MUST be "noarch"
```
Consistency

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+Several factors can cause wheel names to differ from conda-style names:
+
+1. Name normalization: PyPI normalizes dashes to underscores (PEP 503)
+  - Example: `lazy-loader` (conda-forge) vs `lazy_loader` (PyPI wheel)
+2. Python-specific clarification: PyPI packages are implicitly Python libraries
+  - Example: `authzed-py` (conda-forge) vs `authzed` (PyPI)
+3. Variant differences: Conda may offer multiple variants with different dependencies
+  - Example: `seaborn-base` (conda-forge) vs `seaborn` (PyPI)
+4. Cross-channel naming: Different conda channels may use different names
+  - Example: `pyperformance` (conda-forge) vs `performance` (main)
```
travishathaway:
```suggestion
1. **Name normalization:** PyPI normalizes dashes to underscores (PEP 503)
  - Example: `lazy-loader` (conda-forge) vs `lazy_loader` (PyPI wheel)
2. **Python-specific clarification:** PyPI packages are implicitly Python libraries
  - Example: `authzed-py` (conda-forge) vs `authzed` (PyPI)
3. **Variant differences:** Conda may offer multiple variants with different dependencies
  - Example: `seaborn-base` (conda-forge) vs `seaborn` (PyPI)
4. **Cross-channel naming:** Different conda channels may use different names
  - Example: `pyperformance` (conda-forge) vs `performance` (main)
```
Style edits to stay consistent with the rest of the document.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+2. Pure Python only: This CEP explicitly does not address wheels with binary extensions, which require platform-specific compatibility guarantees beyond the current scope. Conda’s strength is binary compatibility, so using conda packages may be the optimal solution.
+3. Environment markers: Only Python version markers are converted to dependencies. Other environment markers (OS, platform, etc.) are ignored based on the pure Python assumption.
+4. Conditionals and Extras: Conditional dependencies and extras are specified by a separate CEP.
+5. Repodata size: Supporting a significant portion of pure Python packages from PyPI (potentially hundreds of thousands of packages with multiple versions each) will substantially increase repodata size. Channels adopting wheel support at scale will need to implement sharded repodata ([CEP 16][cep-16]) to maintain acceptable performance. Alternatively, channels may choose to curate a subset of popular or requested packages rather than mirroring all of PyPI.
+
+## Implementation Notes
+
+### For conda clients
+
+Clients implementing this CEP should:
```
travishathaway:
```suggestion
Clients implementing this CEP SHOULD:
```
Consistency.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+```
+conda create -n pip-environment python pip
+conda activate pip-environment
+pip install <package>
+```
+Despite its safety for environment management, relying solely on conda for this purpose prevents leveraging the package distribution capabilities of the conda ecosystem.
+
+### Editable installs with conda for dependencies only
+
+Conda provides all the dependencies of a given package. Then that package is installed on top in editable mode, without addressing dependencies to make sure we don’t accidentally overwrite conda files:
```
travishathaway:
```suggestion
Conda provides all the dependencies of a given package. Then that package is installed on top in editable mode, without addressing dependencies to make sure conda files aren't accidentally overwritten:
```
Writing style nitpick.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+
+### Add more conda packages
+
+Create and maintain new conda packages for each PyPI dependency needed. Tools like [Grayskull] exist to make this easier to convert. However, this is a significant workload for the community, with over half of all conda-forge packages being pure Python. Even with more dedicated resources, creating recipes for over 400 thousand pure Python packages is not achievable.
+
+### Add interoperability to tools through pip dependency scanning
+
+The original version of the conda-pypi plugin called `pip` with the `--dry-run` option to analyze the solution to install a package. With a list of all the dependencies needed, the plugin installed everything available in conda channels first and what was left over would be installed with `pip install --no-deps`. Disadvantages of this approach include:
+
+- Users still have to know which packages they want from PyPI and then have to `conda pip install` them.
```
travishathaway:
```suggestion
- Users still have to know which packages they want from PyPI and then have to run `conda pip install` to install them.
```
To stay consistent with how it referenced to below.

cep-XXXX.md
```diff
@@ -0,0 +1,568 @@
...
+- Conda's SAT-based solvers (libsolv, resolvo) require complete package metadata upfront to solve in a single pass, while PyPI's API is designed for lazy evaluation with incremental fetching and backtracking (as pip and uv do). Adapting conda's solvers would require fundamental architectural changes.
+
+- On-demand fetching would also require hundreds of sequential HTTP requests during solving, making solves slow and network-dependent while breaking offline capability and reproducibility guarantees.
+
+### Automatic wheel to conda package conversion and hosting
+
+Another alternative would be establishing a build farm to automatically convert wheels to conda packages and host them on a channel, with conversion triggered by popularity, community requests, or on-demand.
+
+#### Advantages of automatic conversion
+- Unified format: All packages are conda packages, simplifying client implementation
```
travishathaway:
```suggestion
- All packages are conda packages, simplifying client implementation
```
Maybe remove this part to stay consistent with the rest of the list items?

beckermr:
I'm going to leave my opinions on the general goals/ideas/features of this CEP in an effort to help bring other perspectives to this CEP. 

TL;DR - As a conda-forge/core developer, I personally would NOT recommend folks use this feature, nor would I enable or offer support for this feature for conda-forge.

The CEP states

> By adding native support for pure Python wheels to repodata, conda clients can:
> 
> - Resolve dependencies across conda and PyPI packages in a single solve
> - Provide users with transparent access to the broader Python ecosystem
> - Maintain environment consistency and reproducibility
> - Eliminate the cognitive burden of managing two package managers
> - Fill gaps in conda package availability without requiring new conda builds
> - Reduce the maintenance burden by fully or partially eliminating the need to create and maintain conda recipes for pure-Python packages
> 

Here is a point-by-point explanation of why in my estimation this feature simply would not work for conda-forge. 

> Resolve dependencies across conda and PyPI packages in a single solve

In a non-trivial fraction of cases, even for pure-Python packages, the requirements in the conda-forge have subtle differences from the upstream requirements. These changes range from package renaming (e.g., `-` and `_` are identical for python packages, but not conda packages, matplotlib vs matplotlib-base, etc.) to more substantial changes that enable broader compatibility (e.g., upstream uses exact pins for dependencies, but conda-forge relaxes them because they are clearly not needed). I can imagine that when it comes to pure-Python packages that depend on a compiled backend package, things will also be funky in some cases. 

These requirement differences will likely lead to some funky solves that either the conda or conda-forge developers will hear about.

> Maintain environment consistency and reproducibility

Environment consistency is a tricky concept IMHO. For sure with this CEP one can in some cases create an environment where all constraints are satisfied. However, if the repodata is wrong, due to the issues outlined above, the formal consistency of the requirements doesn't really matter.

Reproducibility is an even trickier concept. For an environment to be reproducible, one needs to have the same solver with that solver run under the same conditions. Let's assume the conda version is fixed and the solver command is run on the same machine. Even then the constraint on having the same conditions combined with this CEP in effect means that both the upstream wheel metadata and the conda channel metadata have to be the same. Given that the most likely source of wheels is pypi, there is no way one can promise those same conditions. 

Even if we restrict ourselves to environments built from lock files, the combination of the conda channel with pypi as a source of wheels will also not always be reproducible. PyPI users can delete packages (as opposed to simply yanking packages) and those deletions will break even locked envs. We do not allow package deletions on conda-forge for this exact reason. Thus for conda-forge, we could not recommend using this feature for reproducible envs from lock files unless PyPI turns off the ability for users to delete files.

> Eliminate the cognitive burden of managing two package managers

The vast majority of the cognative burden is the differing and interacting repodata, not whether or not one types `pip install` or `conda install`. This CEP doesn't and cannot address that issue. Even consistently interpreting the repodata in a single solver doesn't address this issue.

> Fill gaps in conda package availability without requiring new conda builds
> Reduce the maintenance burden by fully or partially eliminating the need to create and maintain conda recipes for pure-Python packages

For the reasons stated above, I am personally skeptical injecting pure-Python wheel metadata into the repodata would consistently result in a correct-enough environments to eliminate the need for new conda builds or the need to repackage pure-python Packages in conda-forge. I am not saying this doesn't work some of the time. Instead I am saying that the solution proposed in this CEP is not so much better than the current "conda, then pip" solve that it can achieve the arguably difficult goals above. Stated another way, in a world where this feature existed instead of the feature to pip-install overtop of conda environments, conda-forge likely would still need/want to repackage everything.

### Other comments

> When there are naming differences between channels, wheel records MUST use the conda-forge package name as the standard.

This statement is a red-flag for me on this CEP. First, conda-forge itself doesn't have an authoritative mapping of its own packages back to python wheels. There are several approaches in the wild and none of them is standardized into an automated bit of repodata tools like conda/mamba/pixi/rattler can read and interact with. See the discussions of PURLS. Second, treating conda-forge as a special channel specifically in a CEP (as opposed to simply using it as a motivating usecase), is definitely IMHO the antithesis of what a CEP is supposed to be. conda is a set of tools and standards and should not be singling out any one purveyor of conda packages.

> !=X.Y.Z → Omit dependency (conda does not support version exclusions; see Limitations section)

I don't follow this comment. Can you clarify? For sure I have used this operator in the run section before (see, e.g., https://github.com/conda-forge/ngmix-feedstock/blob/main/recipe/meta.yaml#L30), At minimum any requirement that is `foo!=xyz` should be added into the constrains section of the repodata. If it is a the wheel's run section, one should add the package without any constraints in the run section of the repodata as well. Th combination of these two additions should achieve the same effect as "install foo, but not version xyz."

h-vetinari:
Not commenting on the whole CEP (I share much of @beckermr's reservations, but currently don't have a strong opinion), just one aspect that's important to get right IMO:

> I think having the wheel index own the mapping is still the right approach, what if we have the index declare which channel it is mapping names to. For example, we could add an optional field called `name_mapping_channel` to the info section like:

I think it would be a good idea to consider whether this can build on top of the proposed PEP [804](https://peps.python.org/pep-0804/) ([discourse](https://discuss.python.org/t/pep-804-an-external-dependency-registry-and-name-mapping-mechanism/103891)), which tries to standardize something useable around the whole name mapping issue.

It'd be certainly better if we can build on top of that (and help support that PEP) rather than inventing yet another scheme.

CC @jaimergp @rgommers @mgorny

danyeaw:
Hi @beckermr, I made the updates to the relevant sections to fix version exclusions, remove reliance on conda-forge for channel mapping, clarify that we need downstream patching ability, added an implementation options section, and added recommendations for protecting against PyPI packages being deleted. Thanks again for all of the great feedback.

Hi @h-vetinari
> I think it would be a good idea to consider whether this can build on top of the proposed PEP [804](https://peps.python.org/pep-0804/)

Thanks that's a great point that we should build on this! I added a call out to PEP-804 in the [Naming standard and channel mapping](https://github.com/danyeaw/ceps/blob/wheel-repodata/cep-XXXX.md#naming-standard-and-channel-mapping) section.

cep-XXXX.md
```diff
@@ -0,0 +1,633 @@
...
+3. **Conditionals and Extras:** Conditional dependencies and extras are specified by a separate CEP.
+4. **Repodata size:** Supporting a significant portion of pure Python packages from PyPI (potentially hundreds of thousands of packages with multiple versions each) will substantially increase repodata size. Channels adopting wheel support at scale will need to implement sharded repodata ([CEP 16][cep-16]) to maintain acceptable performance. Alternatively, channels may choose to curate a subset of popular or requested packages rather than mirroring all of PyPI.
+5. **PyPI package deletion:** PyPI allows package maintainers to delete packages (as opposed to just yanking them), which can break locked environments that reference those packages. Channels using external PyPI URLs directly are subject to this risk. For production use and reproducible environments, channels SHOULD mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs. This ensures that packages referenced in lock files remain available even if deleted from PyPI.
```
JeanChristopheMorinPerso: I agree that deletion is sometimes a problem, but I'm wondering if this could encourage unsafe practices. Maybe we should add some wording or warnings regarding security. Or we could also remove the recommendation altogether or replace
>For production use and reproducible environments, channels SHOULD mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs. This ensures that packages referenced in lock files remain available even if deleted from PyPI.

with
>For production use and reproducible environments, channels can mirror and store wheel artifacts locally rather than relying solely on external PyPI URLs.

(replace should with can, and remove the last sentence).

Opinions?

cep-XXXX.md
```diff
@@ -0,0 +1,633 @@
...
+}
+```
+
+When `name_mapping_channel` is present, channel operators SHOULD:
+
+- Use the declared channel's package names as the standard for name mapping
+- Document any naming mappings specific to their channel
+- Ensure consistency with the declared channel's naming conventions
+
+Channel operators MUST document any naming conventions and mappings specific to their channel, regardless of whether `name_mapping_channel` is declared.
```
JeanChristopheMorinPerso: Are we confortable with saying `MUST` here, considering the current state of affairs when it comes to mappings in conda-forge and main? (and to be fair with conda-forge, they are in a better state than main).

cep-XXXX.md
```diff
@@ -0,0 +1,633 @@
...
+
+This CEP introduces a new optional `artifact_url` field in package records to specify download locations for individual packages.
+
+> Note for this draft: The `artifact_url` field could also be added as a separate CEP to allow it for other record types.
+
+When present, the `artifact_url` field follows these semantics:
+
+- If `base_url` is defined in the repodata info object (per [CEP 15][cep-15]), `artifact_url` contains the path relative to `base_url`
+- If `base_url` is not defined, `artifact_url` contains the complete download URL
+- If `artifact_url` is not present (`null`), the download location is constructed from `base_url` and `fn` (existing behavior)
```
JeanChristopheMorinPerso: Minor: This is a bullet point under "When present, the `artifact_url` field follows these semantics". You could likely do this:
>When present, the `artifact_url` field follows these semantics:
>- If `base_url` is defined in the repodata info object (per [CEP 15][cep-15]), `artifact_url` contains the path relative to `base_url`
>- If `base_url` is not defined, `artifact_url` contains the complete download URL
>
>When not present, the download location is constructed from `base_url` and `fn` (existing behavior)

?

cep-XXXX.md
```diff
@@ -0,0 +1,633 @@
...
+- Extract wheel metadata during solving to populate dependency information
+- Provide the ability to natively install wheels or on-the-fly convert wheels to conda packages for installation
+
+### For channel operators
+
+Channel operators adding wheel support should:
+
+- Implement validation to ensure only pure Python wheels are included
+- Maintain a mapping of PyPI to conda-style names for their channel
+- Consider automation to keep wheel metadata synchronized with PyPI
```
JeanChristopheMorinPerso: I don't know if this is part of a standard or not, but files on PyPI are immutable. So it's impossible to modify the metadata of an existing wheel. Is it possible that you meant "automation to keep the repodata up-to-date with newer releases on PyPI"?

JeanChristopheMorinPerso (commented):
I left mostly questions

cep-XXXX.md
danyeaw: Thanks for the feedback, I like those wording clarifications. I went with "MAY mirror" to match the keywords.

cep-XXXX.md
danyeaw: Thanks so much for this feedback, I updated this based on your suggestion.

cep-XXXX.md
```diff
@@ -0,0 +1,633 @@
...
+- Channel priority configuration (prefer channels with wheels).
+- Explicit wheel requests through `<channel name>::<package name>` syntax.
+- Explicit configuration in the client itself (e.g. `prefer_conda` or `prefer_wheel`)
+
+### Limitations
+
+This CEP has the following known limitations:
+
+1. **Pure Python only:** This CEP explicitly does not address wheels with binary extensions, which require platform-specific compatibility guarantees beyond the current scope. Conda’s strength is binary compatibility, so using conda packages may be the optimal solution.
+2. **Environment markers**:** Only Python version markers are converted to dependencies. Other environment markers (OS, platform, etc.) are ignored based on the pure Python assumption.
```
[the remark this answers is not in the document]
danyeaw: Hi @JeanChristopheMorinPerso, since conditional dependencies is a separate CEP, what if we said that channel operators MAY patch and that they SHOULD support further conditional dependencies as they become available?

cep-XXXX.md
danyeaw: Thanks, I updated it from MUST to SHOULD.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+Wheel dependencies (from METADATA file's Requires-Dist entries) MUST be converted to conda format following these rules:
+
+- **Package names:** Names per [CEP 26][cep-26] and match existing conda-forge package names where they exist
+- **Version specifiers:** Map PEP 440 [version specifiers][version-specifiers] to conda format:
+  - `==X.Y.Z` → `X.Y.Z` (exact pin)
+  - `>=X.Y.Z` → `>=X.Y.Z`
+  - `<=X.Y.Z` → `<=X.Y.Z`
+  - `<X.Y.Z` → `<X.Y.Z`
+  - `>X.Y.Z` → `>X.Y.Z`
+  - `~=X.Y` → `>=X.Y,<X+1.0` (compatible release)
```
baszalmstra: Conda also has this specifier.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+    "click >=7.0",
+    "numpy >=1.20.0"
+  ],
+  "constrains": [
+    "numpy !=1.24.0"
+  ]
+}
+```
+
+> **Note:** The `importlib-metadata` dependency is omitted because the `Requires-Python: >=3.8 makes its environment marker always false because the package requires Python >=3.8, making the python_version < '3.8' marker always false.
```
baszalmstra: That's lucky in this case, maybe just remove this?

cep-XXXX.md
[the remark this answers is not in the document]
baszalmstra: I think its important that we do describe how to convert those markers when conditional dependencies do exist. Because not all markers are trivial to convert (e.g. `platform_machine`, `python_full_version` vs `python_version`). I think that should be in this CEP but I can also imagine that a following could work.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+- Leverages mature conda infrastructure and tooling without client changes
+- Enables metadata enrichment, quality control, and validation before publishing
+
+#### Disadvantages and why native wheel support was chosen
+
+Despite these advantages, this approach was rejected because:
+
+- **Infrastructure burden:** Requires significant storage and bandwidth to host and serve converted packages that duplicate PyPI's CDN infrastructure, plus ongoing maintenance of the conversion pipeline and synchronization automation
+- **Resource inefficiency:** Wheels are already an excellent format for pure Python packages; conversion adds no technical value and wastes resources
+- **Timeliness:** Creates lag between PyPI publication and availability, with curation decisions needed about which packages to convert
```
baszalmstra: This is not different from conversion of the wheels. A manual/automated step still needs to happen to convert the metadata. Converting the wheel data itself takes seconds at worst.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+- **Maintenance complexity:** Every wheel format change requires tooling updates; bugs affect all converted packages
+- **Ecosystem misalignment:** Converting away from wheels moves conda further from Python packaging standards
```
baszalmstra: Worst case we keep the status quo in my eyes. Direct conversion from a pypi wheel to a conda package already removes the in between step of requiring a recipe.

cep-XXXX.md
```diff
@@ -0,0 +1,636 @@
...
+
+To support sparse repodata processing and maintain compatibility with conda's existing infrastructure, wheel records MUST use conda-style naming conventions rather than wheel filenames as keys. This allows solvers to efficiently filter packages by name without parsing the full package record.
+
+### Key naming requirements
+
+The key for each entry in `packages.whl` MUST follow the format `{name}-{version}`.
+
+Examples:
+
+- `httpx-0.28.1`
```
wolfv: I am a little bit afraid that this convention kinda closes the door on other wheel variants (like non-noarch packages) in the future.  We could just align it closer with the actual filename to future-proof things?

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+### Wheel-Specific Record Values
+
+When populating repodata records for pure Python wheels:
+
+- `build`: MUST be py`PY_MAJOR_VERSION`_`build_number` (e.g. py3_0)
+- `build_number`: MUST be 0 for the initial addition of a wheel version. MAY be incremented for subsequent rebuilds of the same - wheel version (e.g., to correct dependencies or metadata)
+- `fn`: MUST be the wheel filename (e.g., package-1.0.0-py3-none-any.whl)
+- `subdir`: MUST be "noarch"
+- `noarch`: MUST be "python"
+- `artifact_url`: MUST be present and follow the semantics described above
```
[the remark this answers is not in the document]
danyeaw: Thanks, changed it to SHOULD.

cep-XXXX.md
danyeaw: Thanks, TIL! I updated the proposal to include the `~=`.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+Like in the example above of only requiring `importlib-metadata` for certain Python versions, conditional and extra dependencies MUST be supported to enable full interoperability between the ecosystems. This will be supported through a separate CEP: <https://github.com/conda/ceps/pull/111>.
+
+### Solver behavior and package preference
+
+#### Dependency resolution
+
+Solvers MUST treat pure Python wheels as valid package candidates during dependency resolution with these constraints:
+
+- **Exclusivity:** Solvers MUST NOT install both a conda package and wheel for the same package name.
+- **Dependency satisfaction:** When a wheel is selected, its `depends` list MUST be satisfied like any conda package.
```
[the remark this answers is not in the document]
danyeaw: Thanks, I added this clarification!

cep-XXXX.md
danyeaw: As you point out, wheel conversion could also be automated without curation. I agree that this isn't a major differentiator between the approaches and I removed it. Thanks!

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+## Implementation Notes
+
+### For conda clients
+
+Clients implementing this CEP SHOULD:
+
+- Parse the new `packages.whl` section alongside existing package sections
+- Apply the same filtering and preference logic used for conda packages
+- Extract wheel metadata during solving to populate dependency information
+- Provide the ability to natively install wheels or on-the-fly convert wheels to conda packages for installation
```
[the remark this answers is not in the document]
danyeaw: This a great point, we have been more focused on getting the repodata to the right format and less on the native installation because our first phase is with conda-pypi which converts wheels to conda packages on the fly. A future step would be to support them natively. I'll start to work on a section about how this will work.

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+This CEP outlines how native support for pure Python wheel packages will be achieved by adding support for them to conda's package index (repodata). When implemented, conda clients will be able to seamlessly install conda packages and pure Python wheels from enabled channels.
+We explicitly limit the scope of this CEP to pure Python wheels to avoid platform-specific binary compatibility issues, but even with this limitation, we dramatically expand the number of available packages for users in the conda ecosystem.
+
+## Motivation
+
+While conda remains a language-agnostic packaging distribution, installing packages for use with Python represents the majority of conda usage. Users frequently encounter packages only available as wheels on PyPI, forcing them to use hybrid workflows that mix conda and pip installations. This approach creates several problems:
```
```suggestion
While conda remains a language-agnostic packaging distribution, installing packages for use with Python represents the majority of conda usage. Users frequently encounter packages only available as wheels on PyPI, forcing them to use hybrid workflows that mix conda and PyPI installations. This approach creates several problems:
```

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+
+3. **Variant differences:** Conda may offer multiple variants with different dependencies
+    - Example: `seaborn-base` (conda-forge) vs `seaborn` (PyPI)
+
+4. **Cross-channel naming:** Different conda channels may use different names
+    - Example: `pyperformance` (conda-forge) vs `performance` (main)
+
+### Naming standard and channel mapping
+
+The key name for wheel records SHALL be constructed by combining the conda-style name (derived from METADATA and normalized per conda conventions) with the version, build string, and wheel tags (abi_tag-platform_tag) extracted from the filename. The python_tag is omitted as it is redundant with the build string. This ensures the name portion follows conda naming conventions, supports multiple rebuilds, and the tags portion aligns with the wheel filename structure.
```
jaimergp: I don't see how we can extend the name-version-build syntax with extra fields without hurting backwards compatibility. All the wheel tags should be condensed in the build string, or dropped from the key.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+
+### For channel operators
+
+Channel operators adding wheel support SHOULD:
+
+- Implement validation to ensure only pure Python wheels are included
+- Maintain a mapping of PyPI to conda-style names for their channel
+- Consider automation to keep the repodata up to date with newer releases on PyPI
+- Document any naming conventions specific to their channel
+- For production channels, consider mirroring wheel artifacts locally to ensure reproducibility and protect against PyPI deletions
```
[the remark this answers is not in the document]
jaimergp: I'm also thinking about this. We are adding a ton of metadata complexity to support the case where "maybe we don't want to copy this file over in a certain way". At that point, why not create a hybrid solver that knows how to traverse the PyPI index with certain rules (consider only pure Python wheels with dependencies already available, or only pure Python).

This proposal is a bit in the middle of nowhere: it requires preparsing all the wheel metadata and exposing it in a conda-friendly way, but then all the work is left for the client when it comes to installing the wheel. So we either go full conda and convert everything from the beginning, or stay 100% in PyPI world and integrate the solvers better (the mapping problem is present in all three solutions).

This proposal does one thing correctly though: it brings metadata control (by allowing repodata patching) without forcing anyone to copy any files (but in some cases it might be unavoidable).

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+```bash
+git clone https://github.com/owner/package.git
+conda create -n editable-install package --deps-only
+conda activate editable-install
+pip install -e . --no-deps
+```
+
+### Add more conda packages
+
+Create and maintain new conda packages for each PyPI dependency needed. Tools like [Grayskull] exist to make this easier to convert. However, this is a significant workload for the community, with over half of all conda-forge packages being pure Python. Even with more dedicated resources, creating recipes for over 400 thousand pure Python packages is not achievable.
```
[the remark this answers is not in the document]
jaimergp: I could see a `conda-forge/wheels-index` repository where the automation pipelines are maintained folks add their requests to:

1. Add a project to the watchlist so it is included in the index
2. Yank certain wheels
3. Repodata patch
4. Import and archive pure wheel feedstocks
5. etc

I don't know if the resulting repodata would be added to conda-forge proper, or maybe a separate conda-forge-wheels channel (to keep the production channel lighter).

jaimergp (commented):
I have read the CEP and added a few comments. I am supportive of having better PyPI support in the conda ecosystem so users don't have to try their luck with multiple overlapping tools, but I have reservations with the scope and direction of this proposal.

First, it doesn't tell the whole story. The current CEP focuses on pre-processing wheel metadata in a friendly way for the conda solver. It doesn't go into the details and nuance of name mapping (with the complications it brings!), or what to do with the solved records once a solution is found: how to install the wheel, how to cache it, how to populate its `conda-meta/*.json` metadata. Ideally, the CEP would cover the whole story, or at least contextualize where it sits in the pipeline and which other CEPs to consult to get the full story.

I have reservations about the metadata-only approach too, and I think this needs to be better discussed in the Rejected ideas section. _Why_ is this approach so desirable compared to the others should go in a Rationale section. IIUC, this is because:

- Mirroring wheels is expensive and maybe unnecessary
- Metadata patching is desirable
- Easier to implement from the solver side of things

Also (unrelated, but worth exploring), if we do go this way, the proposal doesn't need to limit the strategy to wheels only. If we wanted to offer an agnostic field, like `packages.*`, where any package format could be allowed, it wouldn't take much more work:

- `url` shows where to download the artifact from
- `fn` informs of the format, and the client would know how to deal with it
- `depends` and others need to conform to conda conventions and name mappings

(This is to say that a big part of the specification right now doesn't seem to be very wheel-specific).

---

That aside, there are some editorial changes that would need to be made, but let's get there once we have agreed on a direction better.

cep-XXXX.md
danyeaw: Hi @jaimergp, are you saying that we could limit client changes by making what was the "build" string a much longer set of wheel tags?

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+
+When there are naming differences between PyPI wheels and conda packages, channel operators MUST determine the appropriate conda-style name by applying conda naming conventions per [CEP 26 - Identifying Packages and Channels in the conda Ecosystem][cep-26].
+
+To help users understand which naming conventions are being used, channels MAY reference a parent channel that defines the naming conventions and mappings for wheel-to-conda name translation. Parent/child channel relationships will be defined in a separate CEP, which will specify how channels can declare a `parent_channel` field in the `info` section of repodata to inherit naming conventions and mappings from another channel.
+For example, a wheel channel could reference conda-forge as its parent channel, inheriting conda-forge's established naming conventions and package name mappings. This allows channels to avoid duplicating name mapping definitions and ensures consistency across related channels.
+
+Channel operators SHOULD document any naming conventions and mappings specific to their channel, including whether they reference a parent channel for naming conventions.
```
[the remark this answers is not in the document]
danyeaw: Hi @jaimergp, we need to map the names of all the packages and the names of all the depends, otherwise the solver won't be able to create a solution. The PyPI name is saved as part of the fn field for the filename.

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+This example also demonstrates conditional dependencies. The original `METADATA` file from the wheel has the following dependency information:
+
+```text
+Requires-Python: >=3.8
+Requires-Dist: typing-extensions>=4.0.0; python_version < '3.9'
+```
+
+The package record uses conditional dependency syntax (`; if python < 3.9`) to declare `typing_extensions` only when the Python version is < 3.9, matching the original wheel METADATA. The Python version constraint of >=3.8 is directly mapped.
+
+## Rejected ideas
```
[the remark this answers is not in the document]
danyeaw: Nice catch, I forgot about that one. I added it as a rejected approach, thanks!

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+- packages
+- packages.conda
+- removed
+- signatures
+
+This CEP proposes the addition of a new `packages.whl` section to account for the wheel format. This key points to a mapping that MUST contain [repodata record][repodata-record-schema] objects. The key of this mapping MUST follow the specification described below in [Key naming requirements](#key-naming-requirements).
+
+### Record structure and naming convention
+
+To support sparse repodata processing and maintain compatibility with conda's existing infrastructure, wheel records MUST use conda-style naming conventions rather than wheel filenames as keys. This allows solvers to efficiently filter packages by name without parsing the full package record.
```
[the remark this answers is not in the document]
danyeaw: Thanks, I updated this section.

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+    - Example: `lazy-loader` (METADATA name) vs `lazy_loader` (wheel filename)
+
+2. **Python-specific clarification:** PyPI packages are implicitly Python libraries
+    - Example: `authzed-py` (conda-forge) vs `authzed` (PyPI)
+
+3. **Variant differences:** Conda may offer multiple variants with different dependencies
+    - Example: `seaborn-base` (conda-forge) vs `seaborn` (PyPI)
+
+4. **Cross-channel naming:** Different conda channels may use different names
+    - Example: `pyperformance` (conda-forge) vs `performance` (main)
```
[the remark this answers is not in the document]
danyeaw: Thanks, I moved this to a rationale section.

cep-XXXX.md
```diff
@@ -0,0 +1,613 @@
...
+
+To help users understand which naming conventions are being used, channels MAY reference a parent channel that defines the naming conventions and mappings for wheel-to-conda name translation. Parent/child channel relationships will be defined in a separate CEP, which will specify how channels can declare a `parent_channel` field in the `info` section of repodata to inherit naming conventions and mappings from another channel.
+For example, a wheel channel could reference conda-forge as its parent channel, inheriting conda-forge's established naming conventions and package name mappings. This allows channels to avoid duplicating name mapping definitions and ensures consistency across related channels.
+
+Channel operators SHOULD document any naming conventions and mappings specific to their channel, including whether they reference a parent channel for naming conventions.
+
+### Wheel download URLs
+
+This CEP introduces a new optional `artifact_url` field in package records to specify download locations for individual packages.
```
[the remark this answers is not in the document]
travishathaway: I don't have any problems with that.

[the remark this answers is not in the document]
danyeaw: Thanks! I renamed this to `url`.

cep-XXXX.md
```diff
@@ -0,0 +1,634 @@
...
+  - `==X.Y.Z` → `X.Y.Z` (exact pin)
+  - `>=X.Y.Z` → `>=X.Y.Z`
+  - `<=X.Y.Z` → `<=X.Y.Z`
+  - `<X.Y.Z` → `<X.Y.Z`
+  - `>X.Y.Z` → `>X.Y.Z`
+  - `~=X.Y` → `>=X.Y,<X+1.0` (compatible release)
+  - `!=X.Y.Z` → Add to `constrains` field
+- **Multiple specifiers:** Combine with commas (e.g., >=1.0,<2.0)
+- **Python version requirements:** Convert Requires-Python to explicit python dependency
+- **Environment markers:** Ignore markers other than Python version (pure Python assumption)
```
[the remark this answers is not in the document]
danyeaw: I updated the CEP to allow for additional environment markers. I also referenced a non-normative conversion table from conda-pypi.

cep-XXXX.md
```diff
@@ -0,0 +1,636 @@
...
+- If `base_url` is not defined, `artifact_url` contains the complete download URL
+
+When not present (`null`), the download location is constructed from `base_url` and `fn` (existing behavior).
+
+This approach allows packages to be served from:
+
+- A shared relative or absolute `base_url` with all wheels in the same directory, by populating the `base_url` field and leaving the `artifact_url` field empty.
+- A manual PyPI repository with wheels in directories by the package name by populating the absolute URL in the `artifact_url` field, or the `base_url` and a relative path in the `artifact_url` field.
+- External PyPI mirrors or CDNs using absolute URLs by populating the `artifact_url` field, for example to <https://files.pythonhosted.org/packages/.../package-1.0.0-py3-none-any.whl>
+- Mixed sources within the same repodata file
```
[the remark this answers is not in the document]
danyeaw: This is now a separate CEP at https://github.com/conda/ceps/pull/151.

cep-XXXX.md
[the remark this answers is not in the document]
danyeaw: I added a non-normative reference to this for our conda-pypi conversion table.

cep-XXXX.md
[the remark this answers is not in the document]
danyeaw: Hey @baszalmstra and @jaimergp, I broke up the CEP in to 3:

1. Informational overview CEP about an overview, the history, and rejected ideas
2. CEP about repodata for wheels
3. CEP about client solving, mapping, and installation

Please let me know what you think.

danyeaw:
I updated the CEPs to not be reliant on remote URLs or redirects. Although the conda-pypi channel is using remote URLs currently, we could host wheels on a channel like any other conda package and also conda only channels could make use of remote URLs or redirects if we decide to standardize a solution around that. A proposed solution is in https://github.com/conda/ceps/pull/151, but it is still being discussed.

I also updated the links in the CEPs to reference the CEP 43 and 44 for conditionals and extras.

cep-XXX1.md
```diff
@@ -0,0 +1,377 @@
...
+### Solver behavior
+
+Normative rules for how conda clients treat `whl` records during dependency resolution (same candidate pool as other repodata records; no CEP-mandated default between conda builds and wheels) are defined in [CEP XXX2 – Wheel conda client support][cep-xxx2].
+This CEP only requires that published `whl` records are valid [repodata records][repodata-record-schema] so that clients can consume them alongside `packages` and `packages.conda`.
+
+### Limitations
+
+This CEP has the following known limitations:
+
+1. **Pure Python only:** This CEP explicitly does not address wheels with binary extensions, which require platform-specific compatibility guarantees beyond the current scope. Conda’s strength is binary compatibility, so using conda packages may be the optimal solution.
```
intentionally-left-nil, line 198: The actual repodata proposal itself is not `noarch` specific. The same schema would work just as fine in any subdir. Restricting this here now, without thinking through a future-proofing option risks more breaking changes later if/when subdir-specific repodata is supported.

From a quick analysis, there are almost **150,000** pure python wheels which contain platform specific dependencies. This needs solving now, even for the restricted target of `-none-any` wheels
```python
>>> q = """
... SELECT COUNT(*)
... FROM wheel w
... JOIN metadata_blob mb ON mb.sha256 = w.metadata_sha256
... WHERE w.filename LIKE '%-none-any.whl'
...   AND (
...     mb.parsed_json LIKE '%;%os_name%'
...     OR mb.parsed_json LIKE '%;%platform_system%'
...     OR mb.parsed_json LIKE '%;%sys_platform%'
...     OR mb.parsed_json LIKE '%;%platform_machine%'
...   );
... """
... wheel_count = con.execute(q).fetchone()[0]
... print(wheel_count)
...
147,250
```

cep-XXX1.md
```diff
@@ -0,0 +1,377 @@
...
+Before adding a wheel to `whl`, channel operators MUST verify:
+
+- The wheel's platform tag is `any` (i.e., wheel filenames ending in `-none-any.whl`, such as `requests-2.32.5-py3-none-any.whl`)
+- The wheel's ABI tag is `none`
+- The wheel contains no compiled extensions (`.so`, `.pyd`, `.dylib` files)
+- The wheel's `METADATA` file is present and valid
+
+Wheels that fail any of these checks MUST NOT be added to `whl`.
+
+### Dependency conversion
```
intentionally-left-nil, line 120: I think this section (and other related sections about how to generate the repodata) should be dropped from the formal language of the CEP.

There are many reasons why recipe authors, or channel maintainers might want to generate repodata for their wheels differently. As long as the repodata is valid, why should that conversion be out of conformance with this CEP?

For example, should `Provides-Extra` be honored? Pip will refuse to honor a `Requires-Dist` that references an extra not claimed in `Provides-Extra`. However, uv is happy to accept it.

If you formalize an answer either way, then this CEP is responsible for ossifying the support (or realistically limiting the support) for what is able to be accomplished. Any repodata created outside of these requirements is out-of-compliance with the spec.

There are many such oddities in the expansive set of requirement markers. I think that spending time on the core proposal, and leaving the conversion process to companies, teams, and particular use cases is the best way to drive innovation forwards

**I do think that examples of how the CEP allows automatic conversion of wheel metadata to repodata is important, because it shows that it could work, and also highlights potential shortcomings of the proposal.**

cep-XXX1.md
```diff
@@ -0,0 +1,377 @@
...
+
+Normative rules for how conda clients treat `whl` records during dependency resolution (same candidate pool as other repodata records; no CEP-mandated default between conda builds and wheels) are defined in [CEP XXX2 – Wheel conda client support][cep-xxx2].
+This CEP only requires that published `whl` records are valid [repodata records][repodata-record-schema] so that clients can consume them alongside `packages` and `packages.conda`.
+
+### Limitations
+
+This CEP has the following known limitations:
+
+1. **Pure Python only:** This CEP explicitly does not address wheels with binary extensions, which require platform-specific compatibility guarantees beyond the current scope. Conda’s strength is binary compatibility, so using conda packages may be the optimal solution.
+2. **Environment markers (publisher defaults):** Default conversion rules in this CEP focus on Python-version markers on `Requires-Dist`, mapped to `when=` per [CEP 43][cep-43]. Other markers (for example `sys_platform`) are out of scope for those default publisher rules and are not required to be converted into repodata here; see [PEP 508 marker translation guidance](#pep-508-marker-translation-guidance).
```
intentionally-left-nil, line 199: I think this limitation should be addressed in this spec, either by entirely removing the requirements for how markers are translated, or to properly handle the full set of markers we want to support, now or in the future.

Dropping the specifiers and making the dependency unconditional is probably not the right answer, or similarly to drop the conditional dependency and make it mandatory.

Supporting this translation now is not only good hygiene, but it is also required for any wheel which is not pure-python but we would want to support (now or in the future).

Even just for pure python wheels, a quick analysis finds that this would impact 150,000 wheels (see my earlier comment)

There are major question marks if software wants to choose to honor things like platform_system, os_name, or especially platform_machine.

For example, `os_name = "nt"` could conceivably be converted to `when=[__win]` but the same cannot be sad of `os_name != "nt"`.

`platform_machine` is even worse. You kind of can squint and use the `__archspec` virtual package to map some of the platforms, but you again run into the negation place.

Since all of these markers are valid for wheels in general, and valid for dependencies of `-none-any` "pure python" wheels, I think these markers should be discussed further.

With this discussion we can talk about whether we want to (or not) to support the spirit of "one repodata, with `when` (for things like __win etc), or if we want to map this to the traditional conda paradigm of os + arch as subdirs.

My personal opinion is that since subdirs form a core part of the conda packaging ecosystem, this proposal should map to that now from a repodata.json perspective. Whether software or channel maintainers use this approach, or if they choose to drop or modify markers should be left up to the community to innovate

intentionally-left-nil (changes requested):

cep-XXX1.md
```diff
@@ -0,0 +1,377 @@
...
+
+### `whl` dictionary structure
+
+The `whl` dictionary maps conda-like filenames to repodata records. The key MUST follow the format specified in [Key naming requirements](#key-naming-requirements). The value MUST be a repodata record object that conforms to the [repodata record schema][repodata-record-schema] with the following field specifications:
+
+- **`name`**: Taken from the wheel's METADATA `Name` field, normalized per [CEP 26][cep-26] and any name mappings aligned with the channel declared in `info.channel_relations.base` when present (see [Naming standard and channel mapping](#naming-standard-and-channel-mapping)).
+- **`version`**: Taken from the wheel's METADATA `Version` field, normalized per PEP 440.
```
intentionally-left-nil, line 53: This should be a CEP-26/CEP-29 compatible version, not a PEP-440 version.

This matters when using tooling to install wheels, but also to install wheel dependencies (which are matchspec defined, and thus not 100% PEP-440 compatible)

cep-XXX1.md
```diff
@@ -0,0 +1,377 @@
...
+- **`build`**: Format `py{PY_MAJOR_VERSION}_{abi_tag}_{platform_tag}_{build_number}` (e.g., `py3_none_any_0`), conforming to CEP 26 build string conventions and the repodata record schema pattern `^([a-z0-9_.]+_)?[0-9]+$`. The build number MUST be at the end of the build string. The `{abi_tag}` and `{platform_tag}` are extracted from the wheel filename.
```
intentionally-left-nil: I'm not sure if we should formally specify this (especially with conda packages not having such a formal syntax).

I do think having the ability to have an (optional) hash would go a long way to preventing future name collisions in the future, if we did want to specify things. Also, we should change `py{PY_MAJOR_VERSION}` to just `{python_tag}` IMO - I don't see a problem with having the minor version in the build string, and it makes the tooling simpler since it's just a mechanical transformation of the wheel.

This also would directly translate over if wheels are in other subdirs

cep-XXX1.md
danyeaw: Hey @intentionally-left-nil, the wheel metadata is already in PEP440. We aren't inventing some type of version conversion method here, we use the version that comes with the wheel. Conda still interprets that string with its own rules (CEP 26 charset, CEP 33 ordering, CEP 29 matching) when solving.

cep-XXX1.md
danyeaw: Hi @intentionally-left-nil, I am open to changing some parts of the CEP from MUST to MAY if you think there are things that are optional or should allow flexibility. Flexibility is useful if there is truly more than one way to do something. However, it seems like converting the major wheel dependency declarations to conda equivalent fields is required if you are going to have something that works. 

I don't think we should drop how to generate repodata generation in general from the CEP. Could you please provide more details on what you would like to see changed?

cep-XXX1.md
danyeaw: Hi @intentionally-left-nil, for something to be a standard now, it has to be something that is already designed and standardized on. We do not have a solution for how binary wheel packages would mix with conda packages on the same channel. If we come up with a solution for that, then we can create another CEP that includes that broader scope.

I don't understand how the 150K number is relevant. The pure-Python wheels live in the `noarch` subdir, whether or not they depend on other packages that are platform specific (and would live in other subdirs).

cep-XXX1.md
danyeaw: Hi @intentionally-left-nil, are you envisioning that we add something like https://conda.github.io/conda-pypi/developer/marker-conversion/#pep-508-variables in to the CEP for the ones that do have a direct conversion?

intentionally-left-nil:
Hello, I'd like to introduce reroll [github](https://github.com/intentionally-left-nil/reroll), [pypi](https://pypi.org/project/py-reroll/) 
as a potential reference implementation to this CEP, if spec changes are made

Reroll's headline is that it is able to generate repodata.json for 99.9% of all supported wheels on pypi (pure python or otherwise) from my local analysis. There's more details in the README, but a supported wheel is an approximate of whether uv or pip would install the wheel in a python 3.4+ environment. For this initial implementation, I further scope it to CPython, all primary subdirs supported today, and manylinux (as opposed to musl or old linux platform tags), but those restrictions could be revisited later.

Not only does reroll convert the vast bulk of wheels, but it attempts to do so at full specificity, fully preserving the pypi semantics. Reroll will emit an error, rather than guessing or loosening a requirement it cannot translate.

Besides the actual code, I think this approach is important from a CEP perspective for several reasons.

* Ecosystem-wide analysis helps us understand if the proposals we are making will apply to the ecosystem, and how many edge cases or exceptions we would need to deal with
* Validating that a fully mechanical/deterministic approach can provide full fidelity, without the need to loosen or throw away restrictions to capture the semantic meaning from the wheel side
* Prove that repodata generation tools can fully handle converting all aspects of a wheel from PEP/pypi standards to conda, allowing this translation to be a requirement of the repodata itself. This allows solvers and other clients to just treat things as normal conda matchspec and related entities, greatly simplifying codepaths

# Changes needed
Reroll mostly conforms to the draft CEP, but differs in a few key areas.

1. Remove noarch-specific restrictions to the CEP, especially on the repodata side. All repodata should support any subdir, just like conda packages today. Instead, if the community is concerned about compiled wheel compatibility and other rabbit holes, I suggest adding text like: "Clients MAY choose to refuse to solve, or install a wheel which is not pure python. A client should detect this via the "fn" key to determine the original filename, for py*-none-any tags. A client which does choose to install a subdir-specific wheel does so without specific guidance from this CEP"
2. Add consistent wording that all fields of the v3/wheel directory are standard fields as according to the relevant conda CEP's. The wheel filename is in accordance to CEP-26. Build strings are also in accordance to CEP-26. Versions are according to CEP-29 and other related standards. All package names (the package and its dependencies) must be the correct conda-ecosystem name. Dependecies must be in accordance to matchspec CEP-29, and conda conventions (the python_abi metapackage, etc)
3. Subpoint: Remove any requirement on the build tag. Existing solvers (rattler) depend on the build tag uniqueness to properly de-dupe sharded repodata and others, but the current scheme cannot handle all edge cases. Leave it up to package authors to come up with a build tagging scheme that works for them, and instruct tooling to use the `fn` key if it needs the original wheel name
4. Remove, or vastly increase the specificity of the `### Dependency conversion` section. Reroll makes several decisions about dependency conversion which are aimed at its goals, and assumptions. For example, it does not follow the exact logic around operator conversion, but has its own logic. I would prefer to have this section be removed as a normative section of the CEP, and have the CEP instead focus on the outcome (valid matchspec), as opposed to the exact contents

## Why arch-specific repodata
Fundamentally, conda expresses its affinity to a particular platform via subdir. The modern `when` syntax, combined with virtual packages (such as `__win`) help approximate another path, but it is very easy to see where this falls down. For example, how does one translate the PEP-508 string of `packageA ; os_name != "nt"` There is no matchspec `[when="nt != ???"]` because matchspec here means this package has this version. Instead, conda expresses these things as subdirs, and wheels should as well.

Reroll shows how you can use partial solving to reduce PEP-508 markers into simpler conditions, which can then be expressed using the `when` syntax.

The other elephant in the room is for compiled wheels. One hill I will die on is that our users don't know or care about pure python wheels, `noarch` or other subdirs. They just want to conda install a package and have it work. Encoding "pure python" (which also doesn't even have a straightforward definition) means that our users will always have to understand "oh you can't get this wheel from conda for `reasons`". I think this is just very shortsighted. I understand all the complexities that come with this, but there are also plenty of wheels (rust, others) which link quite cleanly into environments and could be used. Instead, I think the decision of what types of wheels belong in the hands of channel administrators, and clients (to properly install the wheels).

I also don't think this is a decision that can be deferred until later. Properly handling "pure python" wheels with platform-specific dependencies requires at least discussing this topic. Although some dependencies can be solved with `when=[__win]` this is not comprehensive and quickly falls apart with complex requirements

## Why conda-specific fields
Clients don't need to understand PEP-440 version strings today. They don't need to understand PEP-508 markers, or pypi naming conventions. If we bleed any of this into the wheel-repodata it means that rattler, conda, and every other implementation will start needing to carry around (and agree upon) the proper parsing of these fields. Some examples: Version range parsing is slightly different between conda and pypi. This is why conda recipes usually have `python >=3.12,<3.13.0a0`. Also since matchspec doesn't support PEP-440 exactly, there would be cases where the version of your package can't be referenced by a dependency specifier. It's best just to treat everything as a normal conda package for the most part. In the same vein, conda currently places no real restrictions on the bulid tag. We should refrain from doing so here as well. Leave it up to packages and channels to decide the proper mechanism for their build tags. If people want to know the originating wheel for a repodata, the new `fn` tag already provides a mechanism to understand the source (and potentially take decisions based off it). Doing so this way will preserve flexibility in the future (for custom build tag matching) the same way that conda has afforded normal conda packages

The build tag mechanism as currently described would have the same build tag for these real wheels in pypi: `2Keys-0.3.4-py2.py3-none-any.whl` and `2Keys-0.3.4-py3-none-any.whl` (along with other conflicts that exist today)

# More information
If you're curious about the decisions behind reroll, please take a look at the [docs](https://github.com/intentionally-left-nil/reroll/tree/main/docs) section. I have spent as much time hand-writing and documenting my decisions as any other part of the repository and it contains a wealth of information about what and why. I also may have made mistakes, and it's much easier to point out mistakes in my documentation, so I can correct the code to match the spec.

If you want to know more about how I am analyzing reroll and conda-pypi across the pypi ecosystem, please look at [reroll-data](https://github.com/intentionally-left-nil/reroll-data). Please note the coding standards are much lower in that repository, and it's just my scratch space to explore and understand the ecosystem. I've been saving my (AI-assisted) analysis in the [notebooks](https://github.com/intentionally-left-nil/reroll-data/tree/main/notebooks) folder

jezdez:
For the record, @intentionally-left-nil's project `reroll` has not been reviewed by the conda maintainers team, and was developed independently. It should be considered a 3rd-party proposal at the moment for the CEP. I already communicated to Anil that I would have appreciated a little more heads-up before posting a claimed 99.9% support of supported wheels and that it seems to ignore the conda-pypi project.

cep-XXX1.md
intentionally-left-nil: An actual example where this is problematic is:
`2Keys-0.3.4-py2.py3-none-any.whl` and `2Keys-0.3.4-py3-none-any.whl`
Both of these wheels exist on pypi. Both of these wheels would reduce to the same build tag according to this mechanism.

Instead of trying to solve this, I would leave it up to channel authors & recipe builders to decide on a build hash that works for them, allows for partial matching as desired, while remaining unique. Anything that needs to know the original wheel name can use the `fn` key

danyeaw: Hi @intentionally-left-nil, ok, I updated this to `{python_tag}`, thanks for the feedback!

PR #146, opened 2026-01-12
CEP 48: A backwards-compatible repodata update strategy for repodata `v3`

jaimergp:
## Checklist for submitter

- [X] I am submitting a new CEP: [A backwards-compatible repodata update strategy for repodata `v3`](https://github.com/jaimergp/ceps/blob/repodata-updates/cep-XXXX.md).

### DIFF
--- a/cep-0048.md
+++ b/cep-0048.md
@@ -1,8 +1,8 @@
-# CEP XXXX - A backwards-compatible update strategy for repodata `v3`
+# CEP 48 - A backwards-compatible update strategy for repodata `v3`
 
 <table>
 <tr><td> Title </td><td> A backwards-compatible update strategy for repodata <code>v3</code> </td>
-<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Status </td><td> Accepted </td></tr>
 <tr><td> Author(s) </td>
   <td>
     Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;,
@@ -11,7 +11,7 @@
   </td>
 </tr>
 <tr><td> Created </td><td> Jan 12, 2026 </td></tr>
-<tr><td> Updated </td><td> Jul 26, 2026 </td></tr>
+<tr><td> Updated </td><td> Aug 21, 2026 </td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/146 </td></tr>
 <tr><td> Implementation </td><td> N/A </td></tr>
 </table>
@@ -252,6 +252,10 @@ For example, for this input `index.json`:
 ```
 
 This would involve bigger changes in the implementations, delaying the adoption of `v3`.
+
+## Changelog
+
+- 2026-08-21: CEP was accepted.
 
 ## Copyright
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -25,7 +25,7 @@ This document addresses the challenges of updating the specification of `repodat
 
 `repodata.json` files are central to the conda ecosystem. They are the main source of packaging metadata and inform solvers about the catalog of available packages and their dependency constraints. As such, innovation work often refrains from modifying it, and the format itself has seen very few changes over its lifetime. However, a few ongoing efforts will inevitably result in `repodata.json` modifications (conditional dependencies, optional dependency groups, non-conda dependencies, etc).
 
-The main problem is the introduction of backwards incompatible changes. The obvious solution is to bump the `repodata_version` field (like it was done with [CEP 15](./cep-0015.md)). However, this is not desirable for existing channels, since it immediately prevents non-compatible clients from interacting with the channel. Since most clients would update via a new version available in the channel, it creates a chichen-and-egg problem that would significantly delay the introduction of new features and hinder adoption.
+The main problem is the introduction of backwards incompatible changes. The obvious solution is to bump the `repodata_version` field (like it was done with [CEP 15](./cep-0015.md)). However, this is not desirable for existing channels, since it immediately prevents non-compatible clients from interacting with the channel. Since most clients would update via a new version available in the channel, it creates a chicken-and-egg problem that would significantly delay the introduction of new features and hinder adoption.
 
 There must be a strategy to introduce backwards incompatible changes without breaking existing channels. This CEP centralizes the discussion for the update strategy and consolidates that feedback into a concrete proposal.
 
@@ -58,7 +58,7 @@ Adding new fields is backwards compatible and does not break older clients, whic
 
 Bumps in this number should only result in backwards incompatible changes that would anyway prevent a channel from operating completely. While `repodata_version: 2` exists (as per CEP 15), its implementations are not sufficiently old to guarantee that the majority of existing conda clients would support it:
 
-- `rattler` supports it [v0.9.0](https://github.com/conda/rattler/blob/main/CHANGELOG.md#090---2023-09-22) (released on 2023-09-22), which means that `pixi` supports it since [v0.4.0](https://github.com/prefix-dev/pixi/blob/d8d2d8a3e8e1ce99707885aa1437e3768614456b/Cargo.toml#L38) (released on 2023-09-22 too).
+- `rattler` supports it since [v0.9.0](https://github.com/conda/rattler/blob/main/CHANGELOG.md#090---2023-09-22) (released on 2023-09-22), which means that `pixi` supports it since [v0.4.0](https://github.com/prefix-dev/pixi/blob/d8d2d8a3e8e1ce99707885aa1437e3768614456b/Cargo.toml#L38) (released on 2023-09-22 too).
 - `conda` only supports it as of [v24.5.0](https://github.com/conda/conda/blob/main/CHANGELOG.md#2450-2024-05-08) (released on 2024-05-08)
 - `mamba` started supporting it in [v2.0](https://github.com/mamba-org/mamba/blob/main/CHANGELOG.md#20240925) (released on 2024-09-25).
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -11,7 +11,7 @@
   </td>
 </tr>
 <tr><td> Created </td><td> Jan 12, 2026 </td></tr>
-<tr><td> Updated </td><td> Mar 9, 2026 </td></tr>
+<tr><td> Updated </td><td> Mar 18, 2026 </td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/146 </td></tr>
 <tr><td> Implementation </td><td> N/A </td></tr>
 <tr><td> Requires </td><td> N/A </td></tr>
@@ -23,9 +23,8 @@ This document addresses the challenges of updating the specification of `repodat
 
 * to provide a way to extend repodata with new variants, _without_ incrementing `repodata_version`
 * to encode new versions of the repodata in parallel with existing one, behind a `vN` key where `N` is the repodata revision
-* to indicate the latest available repodata revision as part of `info`
-* to provide a timestamp indicating the point in time when the index started providing the most recent repodata revision ("migration timestamp")
-* to use the migration timestamp for client messaging, e.g. when an installer is too old to see repodata of a given revision.
+* to indicate the latest available repodata revision as part of the `info` top-level dictionary
+* to provide useful metadata for end-user error reporting in case of lack of support in the client (oldest and newest timestamps, number of packages available)
 
 ## Motivation
 
@@ -41,8 +40,10 @@ A repodata revision introduces backwards incompatible features in a way that doe
 
 A new item MUST be added to the `info.repodata_revisions` array, that MUST list the revisions found in the repodata file as dictionaries with the following schema:
 
-* `revision: int`: The integer identifying the revision.
-* `migrated_at: int`: The timestamp (in milliseconds) that signals when the channel started indexing packages in the new revision; i.e. the timestamp of the first record published in the new metadata.
+* `revision: int`: Required. The integer identifying the revision.
+* `n_packages: int`: Required. The number of packages available in this revision.
+* `oldest: int | None`: Optional. The timestamp (in milliseconds) of the oldest record published in this revision. If set to `None` or missing, the timestamp information is not available.
+* `newest: int | None`: Optional. The timestamp (in milliseconds) of the newest record published in this revision. If set to `None` or missing, the timestamp information is not available.
 
 A new top-level field identified by the syntax `v{revision}` (where `revision` comes from `info.repodata_revisions[*].revision`) MUST map to a dictionary whose schema is presented in the relevant CEP.
 
@@ -54,7 +55,7 @@ The `repodata_version` MUST be `1` or, if [CEP 15](./cep-0015.md) applies, `2`.
 
 ### Version metadata in `info`
 
-Adding a new field in the `info` dictionary is backwards compatible, and can be used by clients to parse the necessary keys directly without having to traverse the whole document. The migration timestamp is useful for client messaging, like "the client is not recent enough to see all records in this channel, please update to ensure you can obtain access to all packages". They are not added as a top-level field to stop polluting the global namespace.
+Adding a new field in the `info` dictionary is backwards compatible, and can be used by clients to parse the necessary keys directly without having to traverse the whole document. The `oldest`, `newest`, and `n_packages` fields are useful for client messaging, like "the client is not recent enough to see all records in this channel, please update to ensure you can obtain access to {n_packages} additional packages published between {oldest} and {newest}". They are not added as a top-level field to stop polluting the global namespace.
 
 ### Using top-level fields for new metadata schemas
 
@@ -82,7 +83,9 @@ A hypothetical new repodata revision `3` would need to present the following `in
     "repodata_revisions": [
       {
         "revision": 3,
-        "migrated_at": 1768249989751,  // 2026-01-12 20:33 UTC
+        "n_packages": 2,
+        "oldest": 1768249989851,
+        "newest": 1773851561010,
       }
     ]
   },
@@ -102,28 +105,29 @@ A hypothetical new repodata revision `3` would need to present the following `in
     }
   },
   "packages.conda": {
-    "example-1.0.0-0.conda": {
+    "package-1.0.0-0.conda": {
       "build": "0",
       "build_number": 0,
       "depends": [],
-      "md5": "82ecc40f09b9c44483e6b70cad2545d7",
-      "name": "example",
+      "md5": "4483e6b70c82ecc40f09b9c4ad2545d7",
+      "name": "package",
       "noarch": "generic",
-      "sha256": "eb65e866067865793b981c2ba74485f75bef441842b5998badc4ec66717685c7",
+      "sha256": "4485f75bef441842b59eb65e866067865793b981c2ba798badc4ec66717685c7",
       "size": 1234,
       "subdir": "noarch",
-      "timestamp": 1689209309623,
+      "timestamp": 1689209359623,
       "version": "1.0.0"
     }
   },
   "v3": {
+    // This is a hypothetical v3, not a real proposal
     "tar.bz2": {},
     "conda": {
-      "example-1.0.0-0": {  // key does not have the extension anymore
+      "example-2.0.0-0": {  // key does not have the extension anymore
         "build": "0",
         "build_number": 0,
         "depends": [
-          "package[version=3,build_number=0,when=__unix]"  // bracket syntax, w/ conditional
+          "package[version=1,build_number=0,when=__unix]"  // bracket syntax, w/ conditional
         ],
         "md5": "82ecc40f09b9c44483e6b70cad2545d7",
         "name": "example",
@@ -132,7 +136,23 @@ A hypothetical new repodata revision `3` would need to present the following `in
         "size": 1234,
         "subdir": "noarch",
         "timestamp": 1768249989851,
-        "version": "1.0.0",
+        "version": "2.0.0",
+        "new_field": "CRITICAL"  // new field
+      },
+      "example-3.0.0-0": {  // key does not have the extension anymore
+        "build": "0",
+        "build_number": 0,
+        "depends": [
+          "package[version=3,build_number=0,when=__unix]"  // bracket syntax, w/ conditional
+        ],
+        "md5": "6b70cad2545d782ecc40f09b9c44483e",
+        "name": "example",
+        "noarch": "generic",
+        "sha256": "74485f75bef441842b5998badc4ec66717685c7eb65e866067865793b981c2ba",
+        "size": 2345,
+        "subdir": "noarch",
+        "timestamp": 1773851561010,
+        "version": "3.0.0",
         "new_field": "CRITICAL"  // new field
       }
     }
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,161 @@
+# CEP XXXX - A backwards-compatible repodata update strategy
+
+<table>
+<tr><td> Title </td><td> A backwards-compatible repodata update strategy </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td>
+  <td>
+    Jaime Rodríguez-Guerra &lt;jaime.rogue@gmail.com&gt;,
+    Bas Zalmstra &lt;bas@prefix.dev&gt;,
+    Wolf Vollprecht &lt;w.vollprecht@gmail.com&gt;
+  </td>
+</tr>
+<tr><td> Created </td><td> Jan 12, 2026 </td></tr>
+<tr><td> Updated </td><td> Jan 12, 2026  </td></tr>
+<tr><td> Discussion </td><td> ... </td></tr>
+<tr><td> Implementation </td><td> N/A </td></tr>
+<tr><td> Requires </td><td> N/A </td></tr>
+</table>
+
+## Abstract
+
+This document addresses the challenges of updating the specification of `repodata.json` files (and its sharded derivatives) by proposing a strategy based on new top-level fields instead of `repodata_version` bumps.
+
+## Motivation
+
+`repodata.json` files are central to the conda ecosystem. They are the main source of packaging metadata and inform solvers about the catalog of available packages and their dependency constraints. As such, innovation efforts often refrain from modifying it, and the format itself has seen very few changes over its lifetime. However, a few ongoing efforts will inevitably result in `repodata.json` modifications (conditional dependencies, optional dependency groups, non-conda dependencies, etc).
+
+The main problem is the introduction of backwards incompatible changes. The obvious solution is to bump the `repodata_version` field (like it was done with [CEP 15](./cep-0015.md)). However, this is not desirable for existing channels, since it immediately prevents non-compatible clients from interacting with the channel. Since most clients would update via a new version available in the channel, it creates a chichen-and-egg problem that would significantly delay the introduction of new features and hinder adoption.
+
+There must be a strategy to introduce backwards incompatible changes without breaking existing channels. This CEP centralizes the discussion for the update strategy and consolidates that feedback into a concrete proposal.
+
+## Rationale
+
+### Version metadata in `info`
+
+Adding a new field in the `info` dictionary is backwards compatible, and can be used by clients to parse the necessary keys directly without having to traverse the whole dictionary. The migration timestamp is useful for client messaging, like "the client is not recent enough to see all records in this channel, please update to ensure you can obtain access to all packages". They are not added as a top-level field to stop polluting the global namespace.
+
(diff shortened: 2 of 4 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Using top-level fields for new metadata schemas
+
+Adding new fields is backwards compatible and does not break older clients, which will simply ignore those and continue operating as usual.
+
+### Freezing `repodata_version` to `1`
+
+Bumps in this number should only result in backwards incompatible changes that would anyway prevent a channel from operating completely. While `repodata_version: 2` exists (as per CEP 15), its implementations are not sufficiently old to guarantee that the majority of existing conda clients would support it:
+
+- `rattler` supports it [v0.9.0](https://github.com/conda/rattler/blob/main/CHANGELOG.md#090---2023-09-22) (released on 2023-09-22), which means that `pixi` supports it since [v0.4.0](https://github.com/prefix-dev/pixi/blob/d8d2d8a3e8e1ce99707885aa1437e3768614456b/Cargo.toml#L38) (released on 2023-09-22 too).
```
h-vetinari: missing "since"

h-vetinari (commented):
I'm not sure if "migrated" / `migrated_at` is the best naming. Aside from a question of what purpose the timestamp serves (see below), I think we should probably rename this to something simpler like `since:`. If we have to choose a verb, I'd rather we take something like "provide" rather than "migrate"; the latter is IMO not a good description for providing multiple repodata revisions simultaneously for an extended period of time.

For argument's sake, I think the below would be pretty self-explanatory

```json
    "repodata_revisions": [
      {
        "revision": 3,
        "since": 1768249989751,     // 2026-01-12 20:33 UTC
        "dropped": 2234563200000,   // 2040-10-23 00:00 UTC
      },
      {
        "revision": 4,
        "since": 1842998400000,     // 2028-05-27 00:00 UTC
      },
      {
        "revision": 5,
        "since": 2053641600000,     // 2035-01-29 00:00 UTC
      }
    ]
```

cep-XXXX.md
```diff
@@ -0,0 +1,149 @@
...
+
+The main problem is the introduction of backwards incompatible changes. The obvious solution is to bump the `repodata_version` field (like it was done with [CEP 15](./cep-0015.md)). However, this is not desirable for existing channels, since it immediately prevents non-compatible clients from interacting with the channel. Since most clients would update via a new version available in the channel, it creates a chichen-and-egg problem that would significantly delay the introduction of new features and hinder adoption.
+
+There must be a strategy to introduce backwards incompatible changes without breaking existing channels. This CEP centralizes the discussion for the update strategy and consolidates that feedback into a concrete proposal.
```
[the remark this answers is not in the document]
jaimergp: I think the Rationale and Specification sections can be changed in order here, and then it's clearer. I'll use this summary as the beginning of the specification though.

jaimergp:
Looks like we have no further comments, nice 😎

cep-XXXX.md
```diff
@@ -0,0 +1,175 @@
...
+## Specification
+
+A repodata revision introduces backwards incompatible features in a way that does not disrupt existing metadata. To do so, a new CEP MUST be proposed, following these guidelines below.
+
+A new item MUST be added to the `info.repodata_revisions` array, that MUST list the revisions found in the repodata file as dictionaries with the following schema:
+
+* `revision: int`: Required. The integer identifying the revision.
+* `n_packages: int`: Required. The number of packages available in this revision.
+* `oldest: int | None`: Optional. The timestamp (in milliseconds) of the oldest record published in this revision. If set to `None` or missing, the timestamp information is not available.
+* `newest: int | None`: Optional. The timestamp (in milliseconds) of the newest record published in this revision. If set to `None` or missing, the timestamp information is not available.
```
chenghlee: We could be a little clearer and say: "The [Unix time](https://en.wikipedia.org/wiki/Unix_time) in milliseconds" (a la CEP 34).

cep-XXXX.md
```diff
@@ -0,0 +1,175 @@
...
+
+A new top-level field identified by the syntax `v{revision}` (where `revision` comes from `info.repodata_revisions[*].revision`) MUST map to a dictionary whose schema is presented in the relevant CEP.
+
+The CEP MUST also specify how to identify whether a given record belongs in the newer version, or can be added to the previous ones (e.g. a new field extending CEP 20's `info/index.json`)
```
chenghlee: Does "The CEP" here mean _this_ CEP, or later "implementation" CEPs making repodata revisions under this CEP?

cep-XXXX.md
```diff
@@ -0,0 +1,175 @@
...
+
+The main problem is the introduction of backwards incompatible changes. The obvious solution is to bump the `repodata_version` field (like it was done with [CEP 15](./cep-0015.md)). However, this is not desirable for existing channels, since it immediately prevents non-compatible clients from interacting with the channel. Since most clients would update via a new version available in the channel, it creates a chicken-and-egg problem that would significantly delay the introduction of new features and hinder adoption.
+
+There must be a strategy to introduce backwards incompatible changes without breaking existing channels. This CEP centralizes the discussion for the update strategy and consolidates that feedback into a concrete proposal.
+
+## Specification
+
+A repodata revision introduces backwards incompatible features in a way that does not disrupt existing metadata. To do so, a new CEP MUST be proposed, following these guidelines below.
+
+A new item MUST be added to the `info.repodata_revisions` array, that MUST list the revisions found in the repodata file as dictionaries with the following schema:
```
chenghlee: Is there a reason to make this an array rather than a dict-of-dicts, with `vN` as the key and the other fields specified below as sub-dicts; e.g.,
```
"info": {
   ...
   "repodata_revisions": {
      "v3": {
        "n_packages": 42,
        "oldest": ....,
        "newest": ....
      },
      "v4": {
        "n_packages": 42,
        "oldest": ....,
        "newest": ....
      },
   }
   ...
}
```

A big issue I can see with `repodata_revisions` an array is that it (by itself) imposes no uniqueness constraint on `revision`, which leads us to have to answer, "what should tools do if they see multiple `{"revision": 3, ...}` in the array?"

cep-XXXX.md
```diff
@@ -0,0 +1,175 @@
...
+
+* `revision: int`: Required. The integer identifying the revision.
+* `n_packages: int`: Required. The number of packages available in this revision.
```
chenghlee: Since this is an required element, we should specify if tools MUST do anything particular with this data (e.g., verify the number of packages in the `vN` field), or is this field only exists for tools to use as they see fit (e.g., making package count verification optional).

cep-XXXX.md
chenghlee: Also, not sure if we need to explicitly say these here, but:

* I assume timestamp ranges for various repodata revisions are allowed to overlap? E.g., the existing (v1/v2) `packages.conda` section can continue to publish packages, independent of any v3 date range.
* If newest and oldest are both present and not None, newest must be greater than or equal to oldest.

dholth: I'm unsure about tracking these numbers and the counts and propagating them back from individual shards generation to the shards index which has the info section. If we imagine a shards-only repodata, then the client would have to download all shards to double check package counts and timestamps.
Allowing conda-index to say "just pull all newer timestamps than X to the v3 section" would be the sort of implementation that I would be interested in, where conda-index would stay simple.
As a client, a boolean "client should be upgraded because there's repodata we don't understand?" is the most interesting information.

baszalmstra: I agree with @dholth here.

cep-XXXX.md
```diff
@@ -0,0 +1,155 @@
...
+      "sha256": "eb65e866067865793b981c2ba74485f75bef441842b5998badc4ec66717685c7",
+      "size": 1234,
+      "subdir": "noarch",
+      "timestamp": 1689209309623,
+      "version": "1.0.0"
+    }
+  },
+  "v3": {
+    "tar.bz2": {},
+    "conda": {
```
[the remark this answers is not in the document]
wolfv: I don't think this matters, and shards are not going to go away. For parsers, it will be easier to uniformly deal with the different version keys, then a new format, I feel.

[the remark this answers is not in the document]
dholth: I've gotten used to the v3 -> format -> package without extension and I kindof like it.

cep-XXXX.md
wolfv: makes sense

cep-XXXX.md
wolfv: maybe we should make it optional? I think it's purely informational

wolfv:
There are two comments that would make this CEP better by @chenghlee and @dholth. 

Using a dictionary for the repodata revisions would encode uniqueness into the revision scheme. This sounds straight forward and like a good change. We will loose the "order" but we can order by the following timestamps.

I propose to remove the metadata values that require "global" indexing state. Instead we should add one configurable timestamp:

```
accepted_since: Timestamp // default value hard coded in indexing tool, can be overwritten
```

I propose a second timestamp to indicate when a given revision was made the _default_ (ie. all packages after this timestamp are published into the revision). 

```
defaulted_at: Option<Timestamp>
```

However, when using the `defaulted_at` timestamp, previous revisions MUST also carry a `defaulted_at` timestamp, so that the index can be recreated deterministically from the information contained within (the time zones between two `defaulted_at` timestamps will create the "revision default window").

cep-XXXX.md
jaimergp: See [e65a5c5](https://github.com/conda/ceps/pull/146/commits/e65a5c5903829332c08e71f106420a85bb8640fa)

cep-XXXX.md
jaimergp: See [e65a5c5](https://github.com/conda/ceps/pull/146/commits/e65a5c5903829332c08e71f106420a85bb8640fa)

cep-XXXX.md
jaimergp: I think that's implied by the commonly understood meaning of "oldest" and "newest". Still, I reworded this part to ensure that it matches actual metadata in the file.

jaimergp:
> We will loose the "order" but we can order by the following timestamps.

I don't think the order is important. I defined `vX` as a sequence  `v[3..]` so you can always sort them for pretty reporting.

The problem I see with `accepted_since` and `defaulted_at` is that it's unclear who makes that decision: the folks who publish packages? the maintainers of the channel infrastructure that govern the indexing tools and their configuration?

What's the intended usefulness of these fields?

I also changed the text so `oldest` / `newest` refer to the file itself, and not the global subdir state (for shards). The client can the aggregate across the consulted shards and report `min(oldest), max(newest)`, which is again just informative for the error reporting and not some sort of universal truth needed for some important process.

cep-XXXX.md
jaimergp: This is hopefully clearer now, PTAL 🙏

cep-XXXX.md
jaimergp: I made these values local to the file only. See [this comment](https://github.com/conda/ceps/pull/146#issuecomment-4383521254) too.

cep-XXXX.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Encoding version information in the filename
+
+One alternative would be to create new `repodata.json` filenames (e.g. `repodata.v4.json`) for each new incompatible bump. However, this was rejected by the authors as it introduces complexity in other areas:
+
+- It would require more HTTP calls to retrieve the latest version served by the channel.
+- It would introduce duplication across `repodata.json` versions and their shards.
+- Indexing tools would need to maintain the different versions of `repodata.json` in sync.
+
+### Nested `name`-`version`-`build_string` dictionaries
```
jaimergp, line 178: I honestly think this is a good idea and so far I don't have any solid arguments to say no. Too much nesting? Taste?

cep-XXXX.md
```diff
@@ -0,0 +1,260 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+This would simplify parsing of records, but would also complicate the enumeration of existing records.
+
+### Expose `MatchSpec` entries in records as dictionaries
```
jaimergp: I also like this one by @baszalmstra a lot; why aren't we doing this?

cep-XXXX.md
baszalmstra: It makes sparsely parsing the repodata a bit less efficient due to the nested structure. And what is the benefit? This file is not for human consumption anyway.

cep-XXXX.md
baszalmstra: We are designing this for pixi as well, but with toml. We are struggling to find a good way to define the recursive boolean expression logic that can exist in when statements nicely.

cep-XXXX.md
jaimergp: @dholth hinted that it would be better suited to find packages by name, instead of parsing strings from the flattened level. But no measurements were provided. If it hinders sparse-parsing, then I guess that's a good argument to include.

cep-XXXX.md
jaimergp: Yea I guess that `when` would remain as a string in this case.

cep-XXXX.md
dholth: I am currently anti-nesting.

wolfv:
I had some thoughts on the v3 metadata: instead of n_package, oldest / newest, how about we let the implementer define a "message" (free form string) to display to the user?

And how about we define two timestamps:

- one for the "first" v3 package on the channel (equivalent to oldest). This could also be a hard-coded timestamp from when the indexing tool started accepting v3 or when the v3 CEP was passed
- one for "hard cutover" time (this is important as that would be a configuration variable that others would have to know about if we want to ever allow publishing v2 packages into v3)

The other option to publish a v2 package to v3 would be to force v3 in the build tooling. In that case we should probably specify that the index tool _SHOULD ALWAYS_ respect the version indicated by the package.

My motivation would be that we should be able to **deterministically rebuild the repodata**. 

Then lastly maybe another question would be if we allow a repodata patch to move a package from v2 -> v3

adam-browning:
Hi @jaimergp — thanks for the v3 rollout strategy.

I want to flag an interaction with CEP #151 that affects organizations
consuming conda through an artifact proxy (Artifactory, Nexus,
Cloudsmith, Anaconda Server, internal mirrors, air-gapped relays).

#146's backwards-compat guarantee protects *older clients* during a v3
rollout. It does not protect the *proxy layer* in between. Once a
client adopts v3, it starts honoring the absolute `url` field
(already required in `packages.whl`, and extended by #151). Those
fetches route directly to external CDNs and bypass every proxy
control — caching, vulnerability/license scanning, egress
allow-listing, audit, and tenant quotas. A compromised or typosquat
channel can weaponize this against any consumer behind a proxy they
trust for the channel host.

Two requests:

1. Strengthen the spec: REQUIRE that any artifact referenced by an
   absolute `url` in a v3 record also remain reachable at the v1/v2
   flat path (`base_url + fn`) for the same record. This lets a proxy
   strip absolute URLs and keep both old and new clients working.

2. Add a "Mirror / Proxy Compatibility" section that names the threat
   model and recommends the above. Today the CEP's compatibility
   story stops at the client and leaves the proxy ecosystem exposed.

Thanks in advance

danyeaw:
Hi @adam-browning, I think any restrictions on usage of `URL` fields should be with that CEP and shouldn't be captured by this backwards compatibility update strategy one. This one just describes how we will revise repodata going forward.

jaimergp:
As per the discussions in [today's community meeting](https://github.com/conda/governance/pull/395), the security implications of `url` inclusion in this CEP need further consideration and would delay the implementation of `v3`. Consensus in the call was to defer `url` to the next revision. Wheel-support implementers can still experiment with `url` fields before they are generalized to all repodata records.

Also, it was suggested to add a free form field to the revision metadata so channel operators can include custom messages per-revision. This is similar to the (unstandardized) notices.json mechanism, so that's worth discussing too.

cep-XXXX.md
```diff
@@ -0,0 +1,262 @@
...
+Hence, we suggest to stick to `repodata_version: 1` and _only_ use `repodata_version: 2` when a new channel needs a global `base_url` for all the entries in the `packages` and `packages.conda` fields.
+
+## Example
+
+```js
+{
+  "repodata_version": 1,
+  "info": {
+    "subdir": "noarch",
+    "repodata_revisions": {
```
danyeaw, line 101: I noticed that py-rattler 0.24.0 now implements this as a list of objects this an integer revision field like:
`"repodata_revisions": [{"revision": 3, "n_packages": 2, "oldest": ..., "newest": ...}]`

baszalmstra: Yeah I think this was not updated yet to follow the latest changes in the CEP. Are you currently relying on this? Happy to implement this ASAP.

baszalmstra: The previous version of py-rattler didnt parse this at all.

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+- Each value MUST be a dictionary of type `dict[str, dict]` where:
+  - Each subkey MUST be a non-empty string representing the artifact filename without its extension.
+  - Each subvalue MUST be a valid [CEP 36](./cep-0036.md) "package record metadata" dictionary,  including these changes:
+    - The `indexed_timestamp` field introduced by CEP PR [#154](https://github.com/conda/ceps/pull/154) SHOULD be set.
+    - The `extra_depends` field introduced by [44](./cep-0045.md) MAY be present.
+    - The MatchSpec strings mentioned in the fields `depends`, `constrains` and the lists of strings within `extra_depends` groups:
+      - MUST set the `name` field to an exact string (no globbing allowed).
+      - MAY set the fields: `version`, `build`, `build_number`, CEP 43's `when`, CEP 44 `extra`, CEP 45 `flags`.
+      - MUST NOT set any other fields.
+      - MUST be represented with the `name` + square-brackets form (e.g. `name[version="1.2.*",build_number=0]`)
```
jaimergp: Reviewers, please pay attention to this part, since it restricts which `MatchSpec` strings can be used.

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+
+### The `v3` top-level key
+
+This key MUST map to a dictionary of type `dict[str, dict]`:
+
+- Each key MUST be a non-empty string. The key SHOULD represent the file extension (without the leading period) of the included artifacts (usually `tar.bz2` and `conda`).
+- Each value MUST be a dictionary of type `dict[str, dict]` where:
+  - Each subkey MUST be a non-empty string representing the artifact filename without its extension.
+  - Each subvalue MUST be a valid [CEP 36](./cep-0036.md) "package record metadata" dictionary,  including these changes:
+    - The `indexed_timestamp` field introduced by CEP PR [#154](https://github.com/conda/ceps/pull/154) SHOULD be set.
```
jaimergp: Question for the audience (and particularly @jezdez): `SHOULD` or `MUST` here?

danyeaw: CEP 47 says that the field is optional and does not rely on this CEP for backwards compatibility. I think for consistency we should keep it as `SHOULD`.

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+    - The `extra_depends` field introduced by [44](./cep-0045.md) MAY be present.
+    - The MatchSpec strings mentioned in the fields `depends`, `constrains` and the lists of strings within `extra_depends` groups:
```
danyeaw:
```suggestion
    - The `extra-depends` field introduced by [44](./cep-0045.md) MAY be present.
    - The MatchSpec strings mentioned in the fields `depends`, `constrains` and the lists of strings within `extra-depends` groups:
```

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+      - MUST set the `name` field to an exact string (no globbing allowed).
+      - MAY set the fields: `version`, `build`, `build_number`, CEP 43's `when`, CEP 44 `extra`, CEP 45 `flags`.
```
danyeaw:
```suggestion
      - MAY set the fields: `version`, `build`, `build_number`, CEP 43 `when`, CEP 44 `extras`, CEP 45 `flags`.
```

cep-XXXX.md
danyeaw:
```suggestion
      - If only `name` is set, the MatchSpec MUST be the bare name string (e.g. `pip`). Empty brackets (e.g. `pip[]`) MUST NOT be used.
      - If any field other than `name` is set, the MatchSpec MUST use the `name` + square-brackets form (e.g. `name[version="1.2.*",build_number=0]`).
```

Does this help clarify the intent instead of a catch-all bracket bullet?

danyeaw: Right now with v3 repodata we are generating with conda-index, version doesn't appear in square brackets. I understand that this CEP as worded would require `pytest[version=">=2.7.2"]` instead of `pytest>=2.7.2`. Is that a correct understanding and can you help me understand the motivation to make the syntax more restrictive? Thanks!

cep-XXXX.md
jaimergp: Yea, we need to get back to this. I think we need to amend that in CEP 45. Other fields in the record (e.g. `build_number`, `track_features`) use underscores as separators, so `extra-depends` is a clear outlier with that dash.

danyeaw: ya, I agree that amending CEP 45 would work as well. We are actually generating repodata on the conda-pypi channel with `extra_depends`, so changing it in the CEP would align the standard with what is being used in practice as well.

cep-XXXX.md
wolfv: I think this can be resolved now :)

cep-XXXX.md
wolfv: Makes sense to me.

baszalmstra: Also curious about this!

jaimergp: Applied, thanks.

jaimergp: Non-square brackets syntax can be very ambiguous (especially when build strings are also added). I want to move away from that syntax in stored repodata (and leave it as syntactic sugar for user input only). Users can still add whatever in the recipes (that bit is not specified here), but repodata creation tools (indexing) must canonicalize the representation in the repodata for clarity.

Is that a good enough reason?

jaimergp: cc @baszalmstra @wolfv

jaimergp:
@conda/steering-council, the vote for this CEP will start soon, like in 24-48h. If you have any comments, this is your last chance. Thanks!

jaimergp:
Dear @conda/steering-council, 

The vote for this CEP has started. It will be open for two weeks, until **August 20th, 2026, 23:59 Anywhere on Earth**. This time period has been chosen to make it eligible for time-out rules. As an Enhancement Proposal vote, it requires 60% affirmative votes to pass. 

- Uwe Korn (@xhochy) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Cheng H. Lee (@chenghlee) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Filipe Fernandes (@ocefpaf) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [ ] yes 
    - [ ] no 
    - [x] abstain 
- Michael Sarahan (@msarahan) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- John Kirkham (@jakirkham) 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- Jannis Leidel (@jezdez) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Wolf Vollprecht (@wolfv) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Bas Zalmstra (@baszalmstra) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Hind Montassif (@Hind-M) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Pavel Zwerschke (@pavelzw) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain

wolfv:
Thanks to everyone who chimed in! Really really excited to see this moving forward!

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+<tr><td> Implementation </td><td> N/A </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+[RFC2119]: https://datatracker.ietf.org/doc/html/rfc2119
+
+## Abstract
+
+This document proposes a set of updates to `repodata.json` files and its derivatives (sharded, subsets) to include the improvements introduced in CEP [43](./cep-0043.md), [44](./cep-0044.md), [45](./cep-0045.md), and [47](./cep-0047.md). To do so in a backwards compatible manner, it also proposes a revision system as a way to extend repodata with new additions, _without_ incrementing `repodata_version`.
```
Hind-M:
```suggestion
This document proposes a set of updates to `repodata.json` files and its derivatives (sharded, subsets) to include the improvements introduced in CEPs [43](./cep-0043.md), [44](./cep-0044.md), [45](./cep-0045.md), and [47](./cep-0047.md). To do so in a backwards compatible manner, it also proposes a revision system as a way to extend repodata with new additions, _without_ incrementing `repodata_version`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+### The `v3` top-level key
+
+This key MUST map to a dictionary of type `dict[str, dict]`:
+
+- Each key MUST be a non-empty string. The key SHOULD represent the file extension (without the leading period) of the included artifacts (usually `tar.bz2` and `conda`).
+- Each value MUST be a dictionary of type `dict[str, dict]` where:
+  - Each subkey MUST be a non-empty string representing the artifact filename without its extension.
+  - Each subvalue MUST be a valid [CEP 36](./cep-0036.md) "package record metadata" dictionary, including these changes:
+    - The `indexed_timestamp` field introduced by [CEP 47](./cep-0047.md) SHOULD be set.
+    - The `extra_depends` field introduced by [44](./cep-0044.md) MAY be present.
```
Hind-M:
```suggestion
    - The `extra_depends` field introduced by [CEP 44](./cep-0044.md) MAY be present.
```

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+    - The MatchSpec strings mentioned in the fields `depends`, `constrains` and the lists of strings within `extra_depends` groups:
+      - MUST set the `name` field to an exact string (no globbing allowed).
+      - MAY set the fields: `version`, `build`, `build_number`, CEP 43 `when`, CEP 44 `extras`, CEP 45 `flags`.
```
Hind-M:
```suggestion
      - MAY set the fields: `version`, `build`, `build_number`, [CEP 43](./cep-0043.md) `when`, [CEP 44](./cep-0044.md) `extras`, [CEP 45](./cep-0045.md) `flags`.
```

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+## Example
+
+```js
+{
+  "repodata_version": 1,
+  "info": {
+    "subdir": "noarch",
+    "repodata_revisions": {
+      "v3": {
+        "n_packages": 2,
```
Hind-M: Maybe I'm missing something but shouldn't `n_packages` be `1` as there is only `example-3.0.0-0` under `v3`?

jaimergp: Ooops, good catch!

```suggestion
        "n_packages": 1,
```

cep-XXXX.md
```diff
@@ -0,0 +1,257 @@
...
+        "oldest": 1768249989851,
+        "newest": 1773851561010
```
```suggestion
        "oldest": 1773851561010,
        "newest": 1773851561010
```

cep-0048.md
```diff
@@ -0,0 +1,257 @@
...
+
+## Example
+
+```js
+{
+  "repodata_version": 1,
+  "info": {
+    "subdir": "noarch",
+    "repodata_revisions": {
+      "v3": {
```
jaimergp, line 102:
```suggestion
      "v3": {
        "message": "This is just an example v3 revision",
```

jaimergp:
@mariusvniekerk, @marcelotrevisani, @msarahan, @jakirkham: gentle reminder <3

jaimergp:
@marcelotrevisani, @jakirkham: last reminder, vote ends tomorrow.

jaimergp:

* SC size: 13
* Votes: 11 (84.6% of SC)
  * Yes: 11 (100% of votes)
  * No: 0
  * Abstain: 0
  
@conda/steering-council, this will be minted as CEP 48.

wolfv:
Awesome!

wolfv (approved):

PR #151, opened 2026-02-02
Add CEP for a URL field for package records

danyeaw:
## Checklist for submitter

- [X] I am submitting a new CEP: [URL field for package records](https://github.com/danyeaw/ceps/blob/url-fields/cep-XXXX.md).

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -19,7 +19,7 @@ described in [RFC2119][RFC2119] when, and only when, they appear in all capitals
 
 ## Abstract
 
-This CEP introduces a new mandatory `url` field in package records to specify download locations for individual packages.
+This CEP introduces a new optional `url` field in package records to specify download locations for individual packages.
 
 ## Motivation
 
@@ -32,7 +32,7 @@ Currently, the download location for a package is constructed by combining the `
 
 Adding a `url` field enables several use cases that are not possible with the current approach:
 
-- **Per-package directories**: Common in PyPI indexes where each package has its own subdirectory
+- **Per-package directories**: Common in [PyPI simple repositories](https://packaging.python.org/en/latest/guides/hosting-your-own-index/#manual-repository) where each package has its own subdirectory
 - **CDN distribution**: Packages can be served from different CDNs with hash-based URLs
 - **Mixed sources**: Different packages in the same repodata can be hosted on different servers
 - **Backward compatibility**: Traditional flat directory structures continue to work by setting `url` to the filename
@@ -41,12 +41,25 @@ Although wheels were the primary motivation, this change provides general flexib
 
 ## Specification
 
-Package repodata records SHALL contain a `url` field, and the value SHALL be set to either an absolute or relative URL.
+Package repodata records MAY contain a `url` field. When present, the value SHALL be set to either an absolute or relative URL.
 
-Conda clients SHALL parse the `url` field as follows:
+Conda clients SHALL construct the download URL as follows:
 
-- If `url` is an absolute URL, use it as-is
-- If `url` is a relative URL, resolve it relative to the `base_url`
+1. Determine the base URL:
+   - If the repodata's `info` object contains a `base_url` field, use that value
+   - Otherwise, use an empty string (`""`)
+
+2. Determine the package path:
+   - If the package record contains a `url` field, use that value
+   - Otherwise, use the package filename (the dictionary key)
+
+3. Resolve the download URL by combining the base URL with the package path following [RFC 3986 URL resolution semantics](https://datatracker.ietf.org/doc/html/rfc3986#section-5) (equivalent to Python's `urllib.parse.urljoin(base_url, package_path)`).
+
+This resolution means:
+
+- **Absolute URLs**: If the package path is an absolute URL (e.g., `https://cdn.example.com/package.conda`), it is used as-is, ignoring the base URL
+- **Relative URLs with base_url**: If the package path is relative (e.g., `subdir/package.conda`) and `base_url` is set (e.g., `https://repo.example.com/`), the result is `https://repo.example.com/subdir/package.conda`
+- **Relative URLs without base_url**: If the package path is relative and no `base_url` is specified, the package path remains relative and will be resolved by the HTTP client relative to the repodata's location
 
 ## Examples
 
@@ -116,15 +129,19 @@ The client would resolve the relative `url` against the `base_url` to get: `http
 
 ## Backward Compatibility
 
-This CEP introduces a new mandatory `url` field to package records, which is a backwards-incompatible change. To maintain backward compatibility, this CEP follows the strategy outlined in [CEP XXXX - A backwards-compatible repodata update strategy](https://github.com/conda/ceps/pull/146).
+This CEP introduces a new optional `url` field to package records. Since the field is optional, this change is backwards-compatible:
 
-The `url` field will be introduced in a new repodata revision using a new top-level field (e.g., `v3`). Existing `packages` and `packages.conda` fields will remain unchanged, allowing older clients to continue functioning while newer clients can use the `url` field for more flexible package hosting.
+- Older clients that don't recognize the `url` field will continue to construct download URLs using the existing method (combining `base_url` with the package filename)
+- Newer clients will use the `url` field when present, falling back to the traditional method when absent
+- Existing repodata without `url` fields will continue to work without modification
+
+This approach allows for gradual adoption, where channels can add `url` fields only for packages that need non-standard locations (such as wheels in subdirectories), while keeping traditional flat-structure packages unchanged.
 
 ## Rejected ideas
 
 ### Using the dictionary key for paths
 
-The directory path could be embedded in the package's dictionary key (e.g., `packaging/packaging-25.0-pyh29332c3_1.conda`). However, this would break existing assumptions that the key is a filename for the `packages` and `packages.conda` sections of the repodata.
+The directory path could be embedded in the package's dictionary key (e.g., `packaging/packaging-25.0-pyh29332c3_1.conda`). While some clients may already support this, an explicit `url` field is clearer and maintains the convention that dictionary keys are filenames, not paths.
 
 ### Adding an `fn` field alongside `url`
 
@@ -133,7 +150,8 @@ A separate `fn` field could specify a different filename for saving locally. How
 ## References
 
 - [CEP 15 - Hosting repodata.json and packages separately by adding a `base_url` property](https://conda.org/learn/ceps/cep-0015): Introduced the `base_url` field for repodata
-- [PyPI Simple Repository API](https://peps.python.org/pep-0503/): Example of package repositories with per-package directories
+- [PyPI Simple Repository API](https://peps.python.org/pep-0503/): Specification for simple package repositories
+- [Hosting your own simple repository](https://packaging.python.org/en/latest/guides/hosting-your-own-index/#manual-repository): Example of package repositories with per-package subdirectories
 
 ## Copyright
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,143 @@
+# CEP XXXX - URL field for package records
+
+<table>
+<tr><td> Title </td><td> URL field for package records</td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td>
+  Dan Yeaw &lt;dyeaw@anaconda.com&gt;
+</td></tr>
+<tr><td> Created </td><td> Feb 2, 2026</td></tr>
+<tr><td> Updated </td><td> Feb 2, 2026</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/111 </td></tr>
+<tr><td> Implementation </td><td> TBD </td></tr>
+<tr><td> Requires </td><td>https://github.com/conda/ceps/pull/135 https://github.com/conda/ceps/pull/146</td></tr>
+</table>
+
+> The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+This CEP introduces a new mandatory `url` field in package records to specify download locations for individual packages.
+
+## Motivation
+
+Currently, the download location for a package is constructed by combining the `base_url` field (defined in [CEP 16](https://conda.org/learn/ceps/cep-0016)) with the package filename (which serves as the dictionary key in repodata). This has a couple of limitations:
+
+1. Packages must be stored in the same directory
+2. It is not possible to serve the files from different servers or Content Delivery Networks (CDNs)
+
+## Rationale
+
+Adding a `url` field enables several use cases that are not possible with the current approach:
+
+- **Per-package directories**: Common in PyPI indexes where each package has its own subdirectory
+- **CDN distribution**: Packages can be served from different CDNs with hash-based URLs
+- **Mixed sources**: Different packages in the same repodata can be hosted on different servers
+- **Backward compatibility**: Traditional flat directory structures continue to work by setting `url` to the filename
+
+Although wheels were the primary motivation, this change provides general flexibility for package hosting in the conda ecosystem.
+
+## Specification
+
+Package repodata records SHALL contain a `url` field, and the value SHALL be set to either an absolute or relative URL.
+
+Conda clients SHALL parse the `url` field as follows:
+
+- If `url` is an absolute URL, use it as-is
+- If `url` is a relative URL, resolve it relative to the `base_url`
+
+## Examples
+
+### Absolute URL
+
+In this example, the package is served from a CDN with an absolute URL:
+
+```json
+{
+  "packages.conda": {
+    "numpy-2.4.2-py314hd4f4903_0.conda": {
+      "build": "py314hd4f4903_0",
+      "build_number": 0,
+      "constrains": [],
+      "depends": [
+        "libblas >=3.9.0,<4.0a0",
+        "libcblas >=3.9.0,<4.0a0",
+        "liblapack >=3.9.0,<4.0a0",
+        "python >=3.14,<3.15.0a0"
+      ],
+      "license": "BSD-3-Clause",
+      "md5": "c15ea513263c9a15d504ab6b087c6d81",
+      "name": "numpy",
+      "sha256": "1f39bde67c7d252f079260f871c1d6e67b9757e94953369081f5764423fb5e01",
+      "size": 8970626,
+      "subdir": "linux-64",
+      "timestamp": 1770020169383,
+      "url": "https://cdn.example.com/packages/numpy/numpy-2.4.2-py314hd4f4903_0.conda",
+      "version": "2.4.2"
+    }
+  }
+}
+```
+
+### Relative URL with base_url
+
+In this example, packages are organized in subdirectories by package name, using a relative URL combined with `base_url`. The repodata file is located at `https://repo.example.com/conda/linux-64/repodata.json`, where `linux-64` is the platform-specific subdirectory:
+
+```json
+{
+  "info": {
+    "base_url": "https://repo.example.com/conda/linux-64/"
+  },
+  "packages.conda": {
+    "packaging-25.0-pyh29332c3_1.conda": {
+      "build": "pyh29332c3_1",
+      "build_number": 1,
+      "depends": [
+        "python >=3.8"
+      ],
+      "license": "Apache-2.0 OR BSD-2-Clause",
+      "md5": "8da6e3f6a14a8f7b8e43f0e4b3e6b5c9",
+      "name": "packaging",
+      "noarch": "python",
+      "sha256": "a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a9b0c1d2e3f4a5b6c7d8e9f0a1b2",
+      "size": 51234,
+      "subdir": "noarch",
+      "timestamp": 1735689600000,
+      "url": "packaging/packaging-25.0-pyh29332c3_1.conda",
+      "version": "25.0"
+    }
+  }
+}
+```
+
+The client would resolve the relative `url` against the `base_url` to get: `https://repo.example.com/conda/linux-64/packaging/packaging-25.0-pyh29332c3_1.conda`
+
+## Backward Compatibility
+
+This CEP introduces a new mandatory `url` field to package records, which is a backwards-incompatible change. To maintain backward compatibility, this CEP follows the strategy outlined in [CEP XXXX - A backwards-compatible repodata update strategy](https://github.com/conda/ceps/pull/146).
+
+The `url` field will be introduced in a new repodata revision using a new top-level field (e.g., `v3`). Existing `packages` and `packages.conda` fields will remain unchanged, allowing older clients to continue functioning while newer clients can use the `url` field for more flexible package hosting.
+
+## Rejected ideas
+
+### Using the dictionary key for paths
+
+The directory path could be embedded in the package's dictionary key (e.g., `packaging/packaging-25.0-pyh29332c3_1.conda`). However, this would break existing assumptions that the key is a filename for the `packages` and `packages.conda` sections of the repodata.
+
+### Adding an `fn` field alongside `url`
+
+A separate `fn` field could specify a different filename for saving locally. However, packages should not be renamed after downloading, and the filename can be obtained from either the `url` basename or the dictionary key. This would add no value while increasing repodata size.
+
+## References
+
+- [CEP 15 - Hosting repodata.json and packages separately by adding a `base_url` property](https://conda.org/learn/ceps/cep-0015): Introduced the `base_url` field for repodata
+- [PyPI Simple Repository API](https://peps.python.org/pep-0503/): Example of package repositories with per-package directories
+
+## Copyright
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).
+
+<!-- links -->
+[RFC2119]: https://datatracker.ietf.org/doc/html/rfc2119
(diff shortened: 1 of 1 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,143 @@
...
+- **Per-package directories**: Common in PyPI indexes where each package has its own subdirectory
+- **CDN distribution**: Packages can be served from different CDNs with hash-based URLs
+- **Mixed sources**: Different packages in the same repodata can be hosted on different servers
+- **Backward compatibility**: Traditional flat directory structures continue to work by setting `url` to the filename
+
+Although wheels were the primary motivation, this change provides general flexibility for package hosting in the conda ecosystem.
+
+## Specification
+
+Package repodata records SHALL contain a `url` field, and the value SHALL be set to either an absolute or relative URL.
```
baszalmstra: Wouldnt it be easier to _optionally_ add the url? And fallback to previous behavior if its not present?

danyeaw:
Hi @jaimergp, I would to please move this to the RFC stage, thanks!

jaimergp:
Dear @conda/steering-council, this CEP is officially entering its Request For Comments (RFC) period. This will last two weeks, until **May 12th EOD AoE**.

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+<tr><td> Title </td><td> URL field for package records</td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td>
+  Dan Yeaw &lt;dyeaw@anaconda.com&gt;
+</td></tr>
+<tr><td> Created </td><td> Feb 2, 2026</td></tr>
+<tr><td> Updated </td><td> Feb 2, 2026</td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/111 </td></tr>
+<tr><td> Implementation </td><td> TBD </td></tr>
+<tr><td> Requires </td><td><a href="https://github.com/conda/ceps/pull/135">conda/ceps#135</a> (CEP 36 — package metadata files served by conda channels)<br><a href="https://github.com/conda/ceps/pull/146">conda/ceps#146</a> (backwards-compatible repodata update strategy)</td></tr>
```
```suggestion
<tr><td> Requires </td><td> <a href="https://github.com/conda/ceps/pull/146">conda/ceps#146</a> (backwards-compatible repodata update strategy)</td></tr>
```

No need to annotate approved CEPs here. Also, if we require 146, that means they have to be voted together. Are we doing that? Or do you think this CEP can be voted separately?

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Abstract
+
+This CEP introduces a new optional `url` field in package records to specify download locations for individual packages.
+
+## Motivation
+
+Currently, the download location for a package is constructed by combining the `base_url` field (defined in [CEP 16](https://conda.org/learn/ceps/cep-0016)) with the package filename (which serves as the dictionary key in repodata). This has a couple of limitations:
+
+1. Packages must be stored in the same directory
```
```suggestion
1. Packages must be stored in the same directory as `repodata.json`
```

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+- **Per-package directories**: Common in [PyPI simple repositories](https://packaging.python.org/en/latest/guides/hosting-your-own-index/#manual-repository) where each package has its own subdirectory
+- **CDN distribution**: Packages can be served from different CDNs with hash-based URLs
+- **Mixed sources**: Different packages in the same repodata can be hosted on different servers
+- **Backward compatibility**: Traditional flat directory structures continue to work by setting `url` to the filename
+
+Although wheels were the primary motivation, this change provides general flexibility for package hosting in the conda ecosystem.
+
+## Specification
+
+Package repodata records MAY contain a `url` field. When present, the value SHALL be set to either an absolute or relative URL.
```
jaimergp: Who's adding that `url` field? I assume it does NOT come from `info/index.json` files in records; instead it's added by the indexing tool because they are processing remote files, correct?

jaimergp: What's a "relative URL"? Do you mean a "relative path"?
```suggestion
Package repodata records MAY contain a `url` field. When present, the value MUST be set to either a full URL or a relative POSIX path.
```

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+
+2. Determine the package path:
+   - If the package record contains a `url` field, use that value
+   - Otherwise, use the package filename (the dictionary key)
+
+3. Resolve the download URL by combining the base URL with the package path following [RFC 3986 URL resolution semantics](https://datatracker.ietf.org/doc/html/rfc3986#section-5) (equivalent to Python's `urllib.parse.urljoin(base_url, package_path)`).
+
+This resolution means:
+
+- **Absolute URLs**: If the package path is an absolute URL (e.g., `https://cdn.example.com/package.conda`), it is used as-is, ignoring the base URL
```
```suggestion
- **Full URLs**: If the package path is a full URL (e.g., `https://cdn.example.com/package.conda`), it is used as-is, ignoring the base URL
```

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+- **Relative URLs with base_url**: If the package path is relative (e.g., `subdir/package.conda`) and `base_url` is set (e.g., `https://repo.example.com/`), the result is `https://repo.example.com/subdir/package.conda`
```
```suggestion
- **Relative path with `base_url`**: If the package path is relative (e.g., `subdir/package.conda`) and `base_url` is set (e.g., `https://repo.example.com/`), the result is `https://repo.example.com/subdir/package.conda`
```

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+- **Relative URLs without base_url**: If the package path is relative and no `base_url` is specified, the package path remains relative and will be resolved by the HTTP client relative to the repodata's location
```
```suggestion
- **Relative path without `base_url`**: If the package path is relative and no `base_url` is specified, the package path remains relative and will be resolved by the HTTP client relative to the repodata's location
```

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Examples
+
+### Absolute URL
+
+In this example, the package is served from a CDN with an absolute URL:
```
```suggestion
In this example, the package is served from a CDN with a full URL:
```

cep-XXXX.md
jaimergp: Can relative paths traverse parents with `../` or is that forbidden? If not, should we forbid that for security purposes?

cep-XXXX.md
```diff
@@ -0,0 +1,164 @@
...
+}
+```
+
+The client would resolve the relative `url` against the `base_url` to get: `https://repo.example.com/conda/linux-64/packaging/packaging-25.0-pyh29332c3_1.conda`
+
+## Backward Compatibility
+
+This CEP introduces a new optional `url` field to package records. Since the field is optional, this change is backwards-compatible:
+
+- Older clients that don't recognize the `url` field will continue to construct download URLs using the existing method (combining `base_url` with the package filename)
```
jaimergp: Won't this mean that these old clients will get 404s for records that define `url`?

cep-XXXX.md
jaimergp: Are there any rules where the filename in the `url` field must match the filename in the repodata key? Should they? Which filename is it used for the local cache if they don't match?

danyeaw: Hi @jaimergp, thanks for the feedback! I updated the CEP to exclude path traversal to parents. I think it isn't great for security and other methods of declaring base_urls are already flexible enough without it.

danyeaw: Hi @jaimergp, thanks! I updated this to make it more clear that it is the index tool that adds the url.

cep-XXXX.md
danyeaw: We will use the backwards compatible repodata update strategy, I made this more clear in the CEP.

cep-XXXX.md
danyeaw: Thanks, I added language about the cache key needing to be the basename of the file. This will support both current files and future ones like wheels.

baszalmstra:
One question I have is what we do with lock-files. Let me explain.

Currently, lockfiles store the packages by their url. But now we have two urls. We have the "channel url" (`<channel>/<subdir>/<filename>`) and the "download url" as proposed by this CEP. I was first thinking we could just use the "download url" as the url for the package in the lockfile, however, thinking about it more, this poses a small problem. Two metadata entries in repodata can have different metadata but the same download url. 

I think we should still use the "channel url" in lock files to uniquely identify the package in the channel/repodata, and also encode the "download url" separately.

I think it would be good to include this information in the CEP as a recommendation. WDYT?

danyeaw:
Hi @baszalmstra, this is a great point, thanks for bringing it up! I agree on using the channel URL for packages on the channel. But the recommendation needs to handle the case where a record's url is an absolute external URL, for example how we are using it for wheels linked to a CDN. In that case the artifact doesn't live on the channel at all, so the download URL is the primary artifact identifier. What if I update the CEP to say:

- Absolute url: use the download URL as the primary lockfile identifier
- Relative url or no url: use the channel URL as the primary lockfile identifier

Would that be clear and unambiguous?

baszalmstra:
I think it doesn't pose a problem as long as we keep both urls in the lockfile. We will never download from the "channel url" if a "download url" is present. So in the current `pixi.lock` it could look like (simplified):

```yaml
- conda: https://conda.anaconda.org/pypi-test/noarch/foobar-0.1.0-a.whl
  url: files.pypihosted.com/myexamplehash/cp310-cp310-macosx_10_9_x86_64.whl
- conda: https://conda.anaconda.org/pypi-test/noarch/foobar-0.1.0-b.whl
  url: files.pypihosted.com/myexamplehash/cp310-cp310-macosx_10_9_x86_64.whl
```

Note that the download url is the same but the unique identifier from the channel isnt.

danyeaw:
Hey @baszalmstra, I could definitely be wrong and might be missing something, but to your example, `https://conda.anaconda.org/pypi-test/noarch/foobar-0.1.0-a.whl` doesn't exist as a file. The only place that the file would exist is at the url (`files.pypihosted.com/myexamplehash/cp310-cp310-macosx_10_9_x86_64.whl`). Wouldn't the lockfile look like:

```
environments:
  default:
    channels:
      - url: conda-pypi
      - url: conda-forge
    packages:
      osx-arm64:
        - conda: https://files.pypihosted.com/myexamplehash/annotated_types-0.7.0-py3-none-any.whl
        - conda: https://files.pypihosted.com/myexamplehash/annotated_doc-0.0.4-py3-none-any.whl
```

baszalmstra:
Let me try to clarify with a problem statement:

What happens if you have two wheel channels that have different metadata for the same url? Both could end up in the lock-file, but if we use the "download url" we cant uniquely identify them.

jaimergp:
The way I read it, Bas wants a unique identifier of a package record entry in the repodata. So far, that happened to be the download URL as well, so that's what we used. This CEP breaks that happy accident and now we don't have a guaranteed unique identifier. Bas wants to keep using that URL as a URI (identifier instead of location).

I wonder though if we should elevate PURLs (with the changes proposed in Cheng's #159) as those unique identifiers.

baszalmstra:
Exactly. 👍

danyeaw:
Hi @baszalmstra and @jaimergp, I added a Lockfiles considerations section, please let me know if you think that captures the concern and the right proposal for how to deal with it going forward.

cep-XXXX.md
```diff
@@ -141,6 +141,28 @@ However, older clients that don't recognize the `url` field will continue to con
...
+```
+
+The resolved `url` is therefore no longer a guaranteed unique identifier for a package record.
+
+### Recommendation
+
+Lockfile implementations SHOULD adopt Package URLs (PURLs), as defined in [conda/ceps#159](https://github.com/conda/ceps/pull/159) (draft), as the canonical unique identifier for package records. A conda PURL encodes a package's identity (channel, name, version, build string, and subdir) in a standardized, location-independent form, for example:
+
+```text
+pkg:conda/conda-forge/numpy@2.4.2?build=py314hd4f4903_0&subdir=linux-64
```
baszalmstra, line 161: This is not quite enough because it doesnt include the full url of the channel itself. A lockfile can easily contain numpy from conda-forge but from different sources. I dont know what the canonical form for that is?

wolfv: Agreed, and would argue that we should either stick to the "conda URL" or do something like:

`{conda_url}@{resolved_url}`

Otherwise, indirection via conda_url -> PackageRecord (in lockfile) -> URL sounds like it should work fine.

hunger: As I read conda/ceps#159, this should resolve to `https://conda.anaconda.org/conda-forge/noarch/repodata.json`. There is a repodata in that place, assuming numpy resolves in there with the data provided, this should specify the package OK.

baszalmstra: Yes but if the lockfile contains packages from `http://prefix.dev/conda-forge` and `http://conda.anaconda.org/conda-forge` we need to be able to distinguish them.

adam-browning:
Hi @danyeaw — thanks for driving this CEP.

Before the vote, I want to flag a critical security gap that I think
warrants explicit treatment in the spec: organizations that consume
conda channels through an artifact proxy (Artifactory, Nexus,
Cloudsmith, Anaconda Server, internal mirrors, air-gapped relays)
rely on the channel URL being the single source of truth for every
artifact. They use that boundary for caching, vulnerability/license
scanning, egress allow-listing, and audit.

CEP #151's absolute `url` field breaks that boundary. A compromised
channel maintainer (or typosquat channel) can re-point a single
record at an attacker-controlled host. The artifact bypasses every
proxy and scanner the consumer trusts for the channel host. Hashes
don't help — the attacker controls them.

Two requests:
1. Add a Security Considerations section to the CEP that names this
   threat model and the recommended mitigations.
2. Require client implementations to expose a policy knob (e.g.
   `allow` / `relative_only` / `disabled`) so admins in regulated
   environments can keep fetches inside the trusted channel
   boundary. Without this, enterprise users have no opt-out.

Without an explicit opt-out on the client side, our mutual customers will expose their organizations to downloads circumventing their security gates.

Thanks in advance,

danyeaw:
Hey @adam-browning, thanks for the feedback. I added a new security considerations section. I think a compromised channel maintainer normally could also modify packages on the channel, but I do mention that this adds a narrower attack vector if someone somehow only got access to change the `repodata.json`. I think adding a security knob on the client side is good. Do you think proxies should also be able to restrict usage of the `URL` field? Please let me know what you think.

beckermr:
I want to make sure we track any security implications for url fields added to the index.json blobs at package build time. We don't want folks injecting random urls into artifacts this way. I think these should be stripped, if ever found, and never kept in the indexing process.

wolfv:
I see some alternative possibilities for the mirroring use case:

- we could have a field similar to `base_url` that can be added into the metadata section of repodata json as the URL forwarder (`url_base_url` or something). There would be a API that we'd ping with the URL and then follow the redirect (e.g. `https://foo.com/?url=bla...` -> pypi.org. If the field is not set, then we just go to the URL indicated.
- Alternatively, we could scrap the proposal and Anaconda can run a reverse proxy themselves that maps from the conda URL to the PyPI wheel URL and forwards to it with a 301. I think that could be the best solution IMO.

danyeaw:
If a channel operator wanted to do something malicious today, they already can. They control the whole channel and all the packages. Is there some other threat scenario that I'm missing? If a channel could reverse proxy a package URL today, that seems like it is even worse than if a channel could explicitly set a malicious channel URL with this CEP. At least the explicit case is auditable.

The only genuinely new thing security related about this CEP is that a proxy/mirror now needs to understand and rewrite the url field to keep downloads within its perimeter. This is no different than a PyPI mirror has to do today.

I agree that packages shouldn't be setting their own URLs, and that should be stripped if they do.

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+"RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+This CEP introduces a new optional `url` field in package records to specify download locations for individual packages.
+
+## Motivation
+
+Currently, the download location for a package is constructed by combining the `base_url` field (defined in [CEP 16](https://conda.org/learn/ceps/cep-0016)) with the package filename (which serves as the dictionary key in repodata). This has a couple of limitations:
```
chenghlee, line 26: The notion of a `base_url` is actually defined in CEP 15. CEP 16 inherits that definition and extends it to support `shards_base_url`.

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. Packages must be stored in the same directory as `repodata.json`
```
chenghlee, line 28: This claim is a little confusing; [CEP 15](https://conda.org/learn/ceps/cep-0015/) defines `base_url` for the explicit purpose of allowing `repodata.json` to live in a separate location than channel base URL for the packages artifacts themselves. What CEP 15 does _not_ provide for each individual package artifact to live in a separate location, which is what this CEP seemingly wants to achieve.

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+2. It is not possible to serve the files from different servers or Content Delivery Networks (CDNs)
+
+## Rationale
+
+Adding a `url` field enables several use cases that are not possible with the current approach:
```
chenghlee, line 33: Would replacing

> Adding a `url` field enables...

with:

> Adding an optional `url` field to individual package repodata records enables...

help clarify the motivation, especially in regards to the scope of the suggested changes?

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+- **Per-package directories**: Common in [PyPI simple repositories](https://packaging.python.org/en/latest/guides/hosting-your-own-index/#manual-repository) where each package has its own subdirectory
+- **CDN distribution**: Packages can be served from different CDNs with hash-based URLs
+- **Mixed sources**: Different packages in the same repodata can be hosted on different servers
+- **Backward compatibility**: Traditional flat directory structures continue to work by setting `url` to the filename
+
+Although wheels were the primary motivation, this change provides general flexibility for package hosting in the conda ecosystem.
```
chenghlee, line 40: In particular, this CEP would allow people to construct their own channel by combining existing ones, beyond the flexibility provided by CEP 42.

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Specification
+
+Package repodata records MAY contain a `url` field. The `url` field is NOT part of the package's `info/index.json`. It is added by the index generation tool when building repodata. When present, the value MUST be set to either a full URL or a relative POSIX path. Relative paths MUST NOT contain `..` path components.
```
chenghlee, line 44: I'd consider replacing:

> relative POSIX path

with

> URI relative path, as defined in [RFC 3986](https://www.rfc-editor.org/info/rfc3986/#section-3.3)

This better aligns with how other parts of this and other CEPs, particularly CEPs 15 and 26, describe how to locate package artifacts and avoids any potential confusion since other CEPs don't strictly require that conda package artifact hosting systems support POSIX filesystem semantics.

(Okay, those CEPs also do not strictly require that such hosting URIs support any notion of a path hierarchy, such that relative paths wouldn't make sense in the first place; but from a practical perspective, that's not a problem I'm terribly worried about. 😅)

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+Conda clients SHALL construct the download URL as follows:
+
+1. Determine the base URL:
+   - If the repodata's `info` object contains a `base_url` field, use that value
+   - Otherwise, use an empty string (`""`)
```
chenghlee: I think this can be collapsed to:

> 1. Determine the base URL as described in [the "Specification" section of CEP 15](https://github.com/conda/ceps/blob/main/cep-0015.md#specification).

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+2. Determine the package path:
+   - If the package record contains a `url` field, use that value
+   - Otherwise, use the package filename (the dictionary key)
+
+3. Resolve the download URL by combining the base URL with the package path following [RFC 3986 URL resolution semantics](https://datatracker.ietf.org/doc/html/rfc3986#section-5) (equivalent to Python's `urllib.parse.urljoin(base_url, package_path)`).
+
+4. Determine the local cache filename:
+   - The basename of the resolved download URL SHALL be used as the local cache filename.
```
chenghlee, line 59: What does `basename` mean in this context?

If it means the last part of the resolved download URL, that could be potentially problematic as it means two channels that serve packages with the same filename but different contents (e.g., `mypkg-4.0-0.conda`) would collide in the cache. Though, that's not a problem specific to this CEP.

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+This resolution means:
+
+- **Full URLs**: If the package path is a full URL (e.g., `https://cdn.example.com/package.conda`), it is used as-is, ignoring the base URL
+- **Relative path with `base_url`**: If the package path is relative (e.g., `subdir/package.conda`) and `base_url` is set (e.g., `https://repo.example.com/`), the result is `https://repo.example.com/subdir/package.conda`
+- **Relative path without `base_url`**: If the package path is relative and no `base_url` is specified, the package path remains relative and will be resolved by the HTTP client relative to the repodata's location
```
chenghlee, line 65: Can these two bullet point collapsed to "use the behavior described in CEP 15, replacing `package filename` with the relative path provided by the value of `url` "?

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+- `relative_only`: absolute URLs in `url` fields are ignored; the client falls back to constructing the download URL from `base_url` and the package filename. RECOMMENDED for environments using artifact proxies or with egress restrictions.
+- `disabled`: the `url` field is ignored entirely; behavior is equivalent to a client that does not implement this CEP.
+
+Channel operators using absolute URLs for CDN distribution SHOULD document the set of external hosts used so that consumers can permit them in egress allow-lists.
+
+## Rejected ideas
+
+### Using the dictionary key for paths
+
+The directory path could be embedded in the package's dictionary key (e.g., `packaging/packaging-25.0-pyh29332c3_1.conda`). While some clients may already support this, an explicit `url` field is clearer and maintains the convention that dictionary keys are filenames, not paths.
```
chenghlee, line 184: Also, [CEP 36's "Schema" section](https://github.com/conda/ceps/blob/main/cep-0036.md#schema) explicitly requires packages' key

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. Determine the base URL:
+   - If the repodata's `info` object contains a `base_url` field, use that value
+   - Otherwise, use an empty string (`""`)
+
+2. Determine the package path:
+   - If the package record contains a `url` field, use that value
+   - Otherwise, use the package filename (the dictionary key)
+
+3. Resolve the download URL by combining the base URL with the package path following [RFC 3986 URL resolution semantics](https://datatracker.ietf.org/doc/html/rfc3986#section-5) (equivalent to Python's `urllib.parse.urljoin(base_url, package_path)`).
```
chenghlee, line 56: I'm wondering if this part can be simplified down to something like:

> 1. If the `url` field is an absolute URL, then its value MUST be accepted of the download URL of the package artifact.
> 2. If the `url` field is a relative URL, then the download URL of the package artifact MUST be constructed according to [CEP 15](https://github.com/conda/ceps/blob/main/cep-0015.md#specification), with the value of `url` used in place of "package filename" within that specification.

That way, we can ensure reasonably consistent behavior as to how `base_url` or its absence should be handled, especially if we ever make future adjustments to CEP 15.

chenghlee: As an aside, I would avoid using Python's `urllib.parse.urljoin(...)` as a reference for how to join URL components, as its behavior is (IMO, at least) a bit quirky:

```
>>> urllib.parse.urljoin('https://example.com/my-channel/linux-64', './mypkg-1.0-0.conda')
'https://example.com/my-channel/mypkg-1.0-0.conda'
```

cep-XXXX.md
chenghlee, line 65: Non-blocking stylistic note: while HTTP is by far the most common protocol used by conda and friends, it's generally not safe to assume that it's the _only_ protocol being used. (E.g., S3 and (S)FTP are definitely not out of the question.)

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+The client would resolve the relative `url` against the `base_url` to get: `https://repo.example.com/conda/linux-64/packaging/packaging-25.0-pyh29332c3_1.conda`
+
+## Backward Compatibility
+
+This CEP introduces a new optional `url` field to package records. Since the field is optional, this change is backwards-compatible:
+
+- Newer clients will use the `url` field when present, falling back to the traditional method when absent
+- Existing repodata without `url` fields will continue to work without modification
+
+However, older clients that don't recognize the `url` field will continue to construct download URLs using the existing method (combining `base_url` with the package filename). If a package is stored at a non-flat path (e.g., `packaging/packaging-25.0-pyh29332c3_1.conda`) and the file does not also exist at the traditional flat path, older clients will receive a 404 error.
```
chenghlee, line 140: Consider replacing

> ...will receive a 404 error.

with

> ...might receive an HTTP 404 or other missing file error.

Though this does raise a (possibly minor) security concern, as in principle, a hostile package repository could supply a compromised package to an older client, and if it that client does not very the checksum of the downloaded artifact, ...

cep-XXXX.md
```diff
@@ -0,0 +1,203 @@
...
+
+```text
+pkg:conda/conda-forge/numpy@2.4.2?build=py314hd4f4903_0&subdir=linux-64
+```
+
+PURLs separate identity from location: two channels that independently index the same external CDN artifact will produce distinct PURLs. The resolved `url` field SHOULD be stored separately in the lockfile for fetching purposes.
+
+## Security Considerations
+
+A package record's `url` field is authored by the channel. An attacker who can write to `repodata.json` can set an absolute `url` pointing to an attacker-controlled host and update the `md5`/`sha256` hashes to match. The client fetches the malicious artifact directly from that host, and could bypass any artifact proxy that the user may rely on for vulnerability scanning, egress control, and audit. Hash verification passes because the attacker controls both the artifact and the expected hashes.
```
chenghlee, line 168: Built into this threat model is the assumption that all conda-compatible clients perform checksum verification of the downloaded artifacts, which I'm 100% sure they all do (especially the older ones).

cep-XXXX.md
chenghlee, line 184: A strict reading of CEP 36 entirely prohibits this as a possibility, as it explicitly requires the keys to be filenames.

chenghlee (changes requested):

PR #154, opened 2026-03-05
CEP 47: Index timestamp in package record metadata

jezdez:
## Checklist for submitter

- [x] I am submitting a new CEP: [Index timestamp in package record metadata](https://github.com/jezdez/ceps/blob/cep-upload-timestamp/cep-XXXX.md).

### DIFF
--- a/cep-0047.md
+++ b/cep-0047.md
@@ -1,11 +1,11 @@
-# CEP XXXX - Index timestamp in package record metadata
+# CEP 47 - Index timestamp in package record metadata
 
 <table>
 <tr><td> Title </td><td> Index timestamp in package record metadata </td></tr>
-<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Status </td><td> Accepted </td></tr>
 <tr><td> Author(s) </td><td> Jannis Leidel </td></tr>
 <tr><td> Created </td><td> Mar 4, 2026 </td></tr>
-<tr><td> Updated </td><td> Jun 9, 2026 </td></tr>
+<tr><td> Updated </td><td> Jun 30, 2026 </td></tr>
 <tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/154 </td></tr>
 <tr><td> Implementation </td><td> <a href="https://github.com/conda/conda/issues/15759">conda/conda#15759</a> </td></tr>
 </table>
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,124 @@
+# CEP XXXX - Upload timestamp in package record metadata
+
+<table>
+<tr><td> Title </td><td> Upload timestamp in package record metadata </td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jannis Leidel </td></tr>
+<tr><td> Created </td><td> Mar 4, 2026 </td></tr>
+<tr><td> Updated </td><td> Mar 4, 2026 </td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+<tr><td> Requires </td><td> CEP 36 </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+
+[RFC2119]: https://datatracker.ietf.org/doc/html/rfc2119
+
+## Abstract
+
+This CEP proposes adding an optional `upload_timestamp` field to per-record metadata in `repodata.json`, as specified by [CEP 36](./cep-0036.md). Unlike the existing `timestamp` field (which records build time and is set by the package builder), `upload_timestamp` is set by the channel server when an artifact is indexed, providing a trustworthy, server-controlled publication time.
+
+## Motivation
+
+The conda ecosystem currently lacks a reliable, per-artifact indicator of when a package was made available on a channel. This gap has practical consequences for two emerging use cases: supply chain security and environment reproducibility.
+
+### The `timestamp` field is builder-controlled
+
+[CEP 34](./cep-0034.md) defines `timestamp` in `info/index.json` as "Starting time of the package build", expressed as Unix time in milliseconds. This value propagates into `repodata.json` as specified by [CEP 36](./cep-0036.md). Because it is set by the build tool (`conda-build`, `rattler-build`), a package author or malicious actor can set it to any value. The `common-1` schema at schemas.conda.org also marks this field as `patchable: true`.
+
+### `channeldata.json` is too coarse
+
+[CEP 38](./cep-0038.md) specifies a `timestamp` field in `channeldata.json`, defined as "Upload date of the most recently published artifact". While this is server-controlled, it is per-package-name, not per-artifact. It only reflects the most recent upload and cannot be used to determine when a specific version or build was published.
+
+### Ecosystem precedent
+
+PyPI provides server-controlled `upload_time` and `upload_time_iso_8601` fields per release file in its JSON API. These are set by the PyPI server at upload time and cannot be modified by package authors. This is the foundation for reliable tools like:
+
+- `uv --exclude-newer` and `pip --exclude-newer` for reproducible resolution
+- Dependency cooldown features that delay installation of recently published packages as a supply chain security measure
+
+The conda ecosystem would benefit from an equivalent signal.
+
+### Use cases
+
+1. **Dependency cooldowns**: A configurable time window that excludes packages published too recently from being installed, giving security vendors time to flag malicious packages. See conda/conda#15759 and [William Woodruff's analysis](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) of supply chain attack windows.
+
+2. **Reproducible environments**: Resolving an environment as it would have looked at a specific point in time (similar to `uv --exclude-newer`), using the publication date rather than the build date.
+
+3. **Auditing and forensics**: Determining exactly when a given artifact appeared on a channel, independent of when it was built.
+
+## Specification
+
+### Package record metadata
+
+[CEP 36](./cep-0036.md) defines the schema for entries in `packages` and `packages.conda` within `repodata.json`. This CEP adds one optional field to the package record metadata:
+
+- `upload_timestamp: int`. Optional. Unix time in milliseconds when the artifact was added to the channel index. It MUST be set by the channel server or indexing tool, not by the build tool. If absent, clients MAY fall back to `timestamp` for time-based operations.
+
+### Channel server requirements
+
+Channel servers and indexing tools (such as `conda-index`, `quetz`, or Anaconda's infrastructure) SHOULD set `upload_timestamp` on each package record when the artifact is first indexed into the channel. The value MUST reflect the actual time the artifact became available in the channel, not the build time or any value from the artifact itself.
+
+If a channel re-indexes existing artifacts without a prior `upload_timestamp`, it MAY use the file modification time or another server-side signal as a best-effort approximation. It MUST NOT copy the value from the artifact's `timestamp` field.
+
+### Client behavior
+
+Conda clients that implement time-based filtering (e.g., dependency cooldowns, exclude-newer) SHOULD prefer `upload_timestamp` when present. If `upload_timestamp` is absent, clients MAY fall back to `timestamp` with appropriate documentation that the value is builder-controlled.
+
+### Schema
+
+The `repodata-record-1.schema.json` at schemas.conda.org SHOULD be updated to include:
+
+```json
+"upload_timestamp": {
+    "type": "integer",
+    "description": "Unix time in milliseconds when the artifact was added to the channel index. Set by the channel server, not the build tool.",
+    "patchable": false,
+    "minimum": 0
+}
+```
+
+Note that `patchable` is `false`: unlike `timestamp`, this field SHOULD NOT be modified by repodata patching, since it represents a server-side fact.
+
+## Rationale
+
+### Why per-record, not per-package
+
+`channeldata.json` (CEP 38) already tracks an upload date, but at the package-name level. Different versions and builds of the same package are published at different times, so a per-record field is necessary for meaningful time-based filtering.
+
+### Why server-controlled
+
+The `timestamp` field is builder-controlled and has known issues: it can be set to any value by the packager, and it reflects when the build started rather than when the package became available to users. A server-controlled field closes this gap.
+
+### Why optional
+
+Not all channel servers may be able to provide this field immediately. Making it optional allows gradual adoption while clients can fall back to `timestamp`.
+
+### Backwards compatibility
+
+Adding an optional field to existing package records in `packages` and `packages.conda` is a purely additive change. [CEP 36](./cep-0036.md) states that additional keys "SHOULD NOT be present and SHOULD be ignored", meaning older clients will silently ignore `upload_timestamp`. This does not require the repodata revision mechanism proposed in [conda/ceps#146](https://github.com/conda/ceps/pull/146), though it is consistent with that proposal's philosophy of backwards-compatible evolution.
+
+This follows the same pattern as the `url` field proposed in [conda/ceps#151](https://github.com/conda/ceps/pull/151), which also adds an optional per-record field to the existing schema.
+
+### Why milliseconds
+
+For consistency with the existing `timestamp` field (CEP 34) and `channeldata.json` timestamp (CEP 38), both of which use Unix time in milliseconds.
+
+## References
+
+- [CEP 34 - Contents of conda packages](./cep-0034.md)
+- [CEP 36 - Package metadata files served by conda channels](./cep-0036.md)
+- [CEP 38 - Channel-wide metadata files served by conda channels](./cep-0038.md)
+- [conda/ceps#146 - A backwards-compatible repodata update strategy](https://github.com/conda/ceps/pull/146)
+- [conda/ceps#151 - URL field for package records](https://github.com/conda/ceps/pull/151)
+- [schemas.conda.org - repodata-record-1.schema.json](https://schemas.conda.org/repodata-record-1.schema.json)
+- [schemas.conda.org - common-1.schema.json](https://schemas.conda.org/common-1.schema.json)
+- [PyPI JSON API - upload_time field](https://warehouse.pypa.io/api-reference/json.html)
+- [conda/conda#15759 - Dependency cooldowns](https://github.com/conda/conda/issues/15759)
+- [William Woodruff - We should all be using dependency cooldowns](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns)
+- [Andrew Nesbitt - Package managers need to cool down](https://nesbitt.io/2026/03/04/package-managers-need-to-cool-down.html)
+
+---
+
+All CEPs are explicitly [CC0 1.0 Universal](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -6,7 +6,7 @@
 <tr><td> Author(s) </td><td> Jannis Leidel </td></tr>
 <tr><td> Created </td><td> Mar 4, 2026 </td></tr>
 <tr><td> Updated </td><td> Mar 4, 2026 </td></tr>
-<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/154 </td></tr>
 <tr><td> Implementation </td><td> NA </td></tr>
 <tr><td> Requires </td><td> CEP 36 </td></tr>
 </table>
@@ -42,7 +42,7 @@ The conda ecosystem would benefit from an equivalent signal.
 
 ### Use cases
 
-1. **Dependency cooldowns**: A configurable time window that excludes packages published too recently from being installed, giving security vendors time to flag malicious packages. See conda/conda#15759 and [William Woodruff's analysis](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) of supply chain attack windows.
+1. **Dependency cooldowns**: A configurable time window that excludes packages published too recently from being installed, giving security vendors time to flag malicious packages. See [conda/conda#15759](https://github.com/conda/conda/issues/15759) and [William Woodruff's analysis](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns) of supply chain attack windows.
 
 2. **Reproducible environments**: Resolving an environment as it would have looked at a specific point in time (similar to `uv --exclude-newer`), using the publication date rather than the build date.
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -64,7 +64,9 @@ The value of a server-controlled index timestamp varies by channel model. Curate
 
 Channel servers and indexing tools (such as `conda-index`, `quetz`, or Anaconda's infrastructure) SHOULD set `indexed_timestamp` on each package record when the artifact is first indexed into the channel. The value MUST reflect the actual time the artifact became available in the channel, not the build time or any value from the artifact itself. The `repodata-record-1` schema at schemas.conda.org SHOULD be updated accordingly.
 
-For artifacts that predate this CEP and lack an `indexed_timestamp`, channel servers MAY seed the value from the artifact's `timestamp` field or from another server-side signal such as file modification time. This is acceptable because the security benefits of `indexed_timestamp` (dependency cooldowns, exclude-newer) protect against the initial publication window, which has long closed for existing artifacts. For artifacts first indexed after this CEP is approved, the value MUST be set by the server at indexing time.
+For artifacts that predate this CEP and lack an `indexed_timestamp`, channel servers MAY seed the value from the artifact's `timestamp` field or from another server-side signal such as file modification time. This is acceptable because the security benefits of `indexed_timestamp` (dependency cooldowns, exclude-newer) protect against the initial publication window, which has long closed for existing artifacts.
+
+For artifacts first indexed after this CEP is approved, the value MUST be set by the server at indexing time.
 
 ### Timestamp validation
 
@@ -86,7 +88,9 @@ The `timestamp` field is builder-controlled and has known issues: it can be set 
 
 ### Why `indexed_timestamp`, not `upload_timestamp`
 
-The field measures when the artifact appeared in the channel index, not when it was uploaded. In many conda workflows, these are distinct events. For example, conda-forge uploads built artifacts to a staging channel (`cf-staging`), validates them, and only then copies them to the production `conda-forge` channel where they are indexed into `repodata.json`. The `conda-index` maintainer also describes this as tracking when a package was "first indexed" rather than when it was uploaded. The name `indexed_timestamp` accurately reflects this, and fits naturally alongside the existing `timestamp` field in the package record schema.
+The field measures when the artifact appeared in the channel index, not when it was uploaded. In many conda workflows, these are distinct events. For example, conda-forge uploads built artifacts to a staging channel (`cf-staging`), validates them, and only then copies them to the production `conda-forge` channel where they are indexed into `repodata.json`. The `conda-index` maintainer also describes this as tracking when a package was "first indexed" rather than when it was uploaded.
+
+The name `indexed_timestamp` accurately reflects this, and fits naturally alongside the existing `timestamp` field in the package record schema.
 
 ### Not subject to repodata patching
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -109,8 +109,22 @@ This follows the same pattern as the `url` field proposed in [conda/ceps#151](ht
 
 For consistency with the existing `timestamp` field (CEP 34) and `channeldata.json` timestamp (CEP 38), both of which use Unix time in milliseconds.
 
+## Future work
+
+`indexed_timestamp` as specified is a server-controlled integer. It is more trustworthy than the builder-controlled `timestamp` field for the use cases this CEP targets, but it is still a single party's claim. A compromised channel server, a backup restore, or a re-indexing operation could silently change the value, and clients have no way to independently verify that the timestamp is authentic. Closing this gap is out of scope for this CEP, but a few directions are worth noting for future work:
+
+- **RFC 3161 trusted timestamps** ([RFC 3161]). A channel server could request a signed Timestamp Response (TSR) from a Trusted Third Party Timestamp Authority (TSA) when indexing a package, and store the TSR alongside `indexed_timestamp` in repodata. Public TSAs include Sigstore's `timestamp.sigstore.dev` (RFC 3628 / RFC 5816), FreeTSA, and DigiCert.
+
+    The Sigstore bundle format already carries RFC 3161 timestamps via its `TimestampVerificationData` field, which means an attestation-aware indexer (see [CEP 27](./cep-0027.md)) could combine the two signals.
+- **Build-time TSA timestamps**. Build tools such as `rattler-build` and `conda-build` could embed a TSA-signed timestamp over the package contents inside the artifact itself, providing a verifiable lower bound on when the package was built. This is orthogonal to `indexed_timestamp` and would more naturally amend [CEP 34](./cep-0034.md).
+
+These directions would require their own CEPs to specify field names, encoding, verification semantics, and migration story.
+
+[RFC 3161]: https://datatracker.ietf.org/doc/html/rfc3161
+
 ## References
 
+- [CEP 27 - Standardizing a publish attestation for the conda ecosystem](./cep-0027.md)
 - [CEP 34 - Contents of conda packages](./cep-0034.md)
 - [CEP 36 - Package metadata files served by conda channels](./cep-0036.md)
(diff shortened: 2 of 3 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
+# CEP XXXX - Upload timestamp in package record metadata
+
+<table>
+<tr><td> Title </td><td> Upload timestamp in package record metadata </td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Jannis Leidel </td></tr>
+<tr><td> Created </td><td> Mar 4, 2026 </td></tr>
+<tr><td> Updated </td><td> Mar 4, 2026 </td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
```
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/154 </td></tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+<tr><td> Implementation </td><td> NA </td></tr>
+<tr><td> Requires </td><td> CEP 36 </td></tr>
```
jaimergp: I need to think what to do about this field. I think it should only be for other PRs that are open at the same time, not necessarily the ones that have been merged already. I know I replaced these URLs for 29-38 but maybe it was not a good idea.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Package record metadata
+
+[CEP 36](./cep-0036.md) defines the schema for entries in `packages` and `packages.conda` within `repodata.json`. This CEP adds one optional field to the package record metadata:
+
+- `upload_timestamp: int`. Optional. Unix time in milliseconds when the artifact was added to the channel index. It MUST be set by the channel server or indexing tool, not by the build tool. If absent, clients MAY fall back to `timestamp` for time-based operations.
+
+### Channel server requirements
+
+Channel servers and indexing tools (such as `conda-index`, `quetz`, or Anaconda's infrastructure) SHOULD set `upload_timestamp` on each package record when the artifact is first indexed into the channel. The value MUST reflect the actual time the artifact became available in the channel, not the build time or any value from the artifact itself.
```
jaimergp: Given that the timestamp is going to be the time of indexation, maybe we should call it `indexed_timestamp` or `indexed_at`.

Alternatively, we could ask the indexers to peek at the file creation time on disk, but that's flaky and filesystem dependent so let's focus on "first time the package appeared on repodata".

jaimergp (commented):
Thank you for bringing this up, I don't think we've discussed the exploitability of `timestamp` before, and this CEP brings up a few good points. Let's use this opportunity to also secure the acceptable values for `timestamp`.

pavelzw (commented):
cool!

baszalmstra:
@jezdez How can we move this forward? Anything I can do to help?

baszalmstra:
@jezdez Gentle ping! I'd like to put this to a vote ASAP but there are still a few open questions.

jezdez:
@baszalmstra Apologies, this dropped off my todo list in the past weeks, I'll take a look!

cep-XXXX.md
jezdez: I think `Requires` is appropriate here since this CEP adds a field to the package record schema defined by CEP 36. Whether it is already merged or co-pending, the dependency relationship is the same.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+
+If a channel re-indexes existing artifacts without a prior `upload_timestamp`, it MAY use the file modification time or another server-side signal as a best-effort approximation. It MUST NOT copy the value from the artifact's `timestamp` field.
+
+### Client behavior
+
+Conda clients that implement time-based filtering (e.g., dependency cooldowns, exclude-newer) SHOULD prefer `upload_timestamp` when present. If `upload_timestamp` is absent, clients MAY fall back to `timestamp` with appropriate documentation that the value is builder-controlled.
+
+### Schema
```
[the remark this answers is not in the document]
jezdez: Agreed, I have removed the schema section. The CEP now just notes that schemas.conda.org should be updated accordingly. Schemas are downstream from the CEP.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+```json
+"upload_timestamp": {
+    "type": "integer",
+    "description": "Unix time in milliseconds when the artifact was added to the channel index. Set by the channel server, not the build tool.",
+    "patchable": false,
+    "minimum": 0
+}
+```
+
+Note that `patchable` is `false`: unlike `timestamp`, this field SHOULD NOT be modified by repodata patching, since it represents a server-side fact.
```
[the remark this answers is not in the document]
jezdez: With the schema section removed, I have added a brief note in the rationale that `indexed_timestamp` is server-controlled and should not be subject to repodata patching, without prescribing a specific schema keyword for it.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+
+[CEP 36](./cep-0036.md) defines the schema for entries in `packages` and `packages.conda` within `repodata.json`. This CEP adds one optional field to the package record metadata:
+
+- `upload_timestamp: int`. Optional. Unix time in milliseconds when the artifact was added to the channel index. It MUST be set by the channel server or indexing tool, not by the build tool. If absent, clients MAY fall back to `timestamp` for time-based operations.
+
+### Channel server requirements
+
+Channel servers and indexing tools (such as `conda-index`, `quetz`, or Anaconda's infrastructure) SHOULD set `upload_timestamp` on each package record when the artifact is first indexed into the channel. The value MUST reflect the actual time the artifact became available in the channel, not the build time or any value from the artifact itself.
+
+If a channel re-indexes existing artifacts without a prior `upload_timestamp`, it MAY use the file modification time or another server-side signal as a best-effort approximation. It MUST NOT copy the value from the artifact's `timestamp` field.
```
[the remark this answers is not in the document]
jezdez: I have revised the re-indexing paragraph to allow seeding `indexed_timestamp` from `timestamp` for artifacts that predate this CEP. The security rationale: cooldowns protect against the initial publication window, and most supply chain attacks have windows under one week (per Woodruff's analysis). That window has long closed for existing artifacts. For newly indexed artifacts going forward, the value must be set by the server at indexing time. This should also address @baszalmstra's concern about keeping `exclude-newer` working without interruption.

cep-XXXX.md
jezdez: After looking into this more, I agree. Renamed to `indexed_timestamp`. The field measures when the artifact appeared in the channel index, not when it was uploaded. In conda-forge's workflow, the upload (to `cf-staging`) and indexation (appearing in the main channel's `repodata.json`) are distinct events separated by validation steps. Daniel also naturally described this as "first indexed" from the `conda-index` perspective. The name fits well alongside the existing `timestamp` field in the package record schema.

cep-0047.md
```diff
@@ -0,0 +1,124 @@
...
+3. **Auditing and forensics**: Determining exactly when a given artifact appeared on a channel, independent of when it was built.
+
+## Specification
+
+### Package record metadata
+
+[CEP 36](./cep-0036.md) defines the schema for entries in `packages` and `packages.conda` within `repodata.json`. This CEP adds one optional field to the package record metadata:
+
+- `upload_timestamp: int`. Optional. Unix time in milliseconds when the artifact was added to the channel index. It MUST be set by the channel server or indexing tool, not by the build tool. If absent, clients MAY fall back to `timestamp` for time-based operations.
+
```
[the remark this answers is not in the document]
jezdez: @jaimergp I have added a recommendation that indexing tools SHOULD validate `timestamp <= indexed_timestamp` and reject packages with `timestamp` values in the future. Kept it as a SHOULD rather than MUST since this tightens rules on an existing field and probably deserves a gradual rollout.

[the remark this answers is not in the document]
jezdez: @dholth I have revised the approach based on your feedback. The field is now called `indexed_timestamp` since that is what `conda-index` actually knows. For existing artifacts, the CEP allows seeding from `timestamp` as a pragmatic migration path. For new artifacts, the value should be set to the current time when first indexed.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+[CEP 38](./cep-0038.md) specifies a `timestamp` field in `channeldata.json`, defined as "Upload date of the most recently published artifact". While this is server-controlled, it is per-package-name, not per-artifact. It only reflects the most recent upload and cannot be used to determine when a specific version or build was published.
+
+### Ecosystem precedent
+
+PyPI provides server-controlled `upload_time` and `upload_time_iso_8601` fields per release file in its JSON API. These are set by the PyPI server at upload time and cannot be modified by package authors. This is the foundation for reliable tools like:
+
+- `uv --exclude-newer` and `pip --exclude-newer` for reproducible resolution
+- Dependency cooldown features that delay installation of recently published packages as a supply chain security measure
+
+The conda ecosystem would benefit from an equivalent signal.
```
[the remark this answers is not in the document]
jezdez: Thanks for the pointer, I have added pixi's `exclude-newer` to the Ecosystem precedent section. This reinforces the need for a trustworthy server-controlled timestamp.

cep-XXXX.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Backwards compatibility
+
+Adding an optional field to existing package records in `packages` and `packages.conda` is a purely additive change. [CEP 36](./cep-0036.md) states that additional keys "SHOULD NOT be present and SHOULD be ignored", meaning older clients will silently ignore `upload_timestamp`. This does not require the repodata revision mechanism proposed in [conda/ceps#146](https://github.com/conda/ceps/pull/146), though it is consistent with that proposal's philosophy of backwards-compatible evolution.
+
+This follows the same pattern as the `url` field proposed in [conda/ceps#151](https://github.com/conda/ceps/pull/151), which also adds an optional per-record field to the existing schema.
+
+### Why milliseconds
+
+For consistency with the existing `timestamp` field (CEP 34) and `channeldata.json` timestamp (CEP 38), both of which use Unix time in milliseconds.
```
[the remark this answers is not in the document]
jezdez: Agreed, consistency is important here.

jezdez:
Thanks everyone for the thorough feedback. I have pushed changes that address all the review threads. Here is a summary:

**Renamed to `indexed_timestamp`**: The field now reflects what it actually measures: when the artifact appeared in the channel index, not when it was uploaded. In conda-forge's workflow, upload (to `cf-staging`) and indexation (appearing in `repodata.json`) are separate steps. Daniel's framing of "first indexed" from the `conda-index` perspective confirmed this was the right name. The CEP title is now "Index timestamp in package record metadata."

**Revised backfill strategy**: The previous "MUST NOT copy from `timestamp`" rule was too strict. The CEP now allows seeding `indexed_timestamp` from `timestamp` for pre-existing artifacts. The security rationale: cooldowns protect against the initial publication window, and that window has long closed for packages already on channels. New artifacts indexed after this CEP is approved must have the value set by the server.

**Removed the schema section**: Agreed with Jaime that schemas are downstream from CEPs. Replaced with a prose note that the `repodata-record-1` schema at schemas.conda.org should be updated accordingly. The non-patchability point is now covered in the rationale instead.

**Added timestamp validation**: Indexing tools SHOULD validate that `timestamp <= indexed_timestamp` and reject future `timestamp` values. Kept as SHOULD rather than MUST since this tightens rules on an existing field.

**Added pixi and channel model context**: pixi's `exclude-newer` is now mentioned in the Ecosystem precedent section, and there is a new section on how different channel models (curated vs community forge) have different risk profiles.

**Added references**: Andrew Nesbitt's posts on dependency cooldowns and package security for AI agents.

I have replied to all inline threads and resolved the ones that are fully addressed. Left the backfill, timestamp validation, and Requires threads open for confirmation.

cep-XXXX.md
jaimergp: Well, in a broad sense, all CEPs depend on the published ones unless they explicitly supersede them. I modelled this field after the PEP `requires` entry. In [PEP 12](https://peps.python.org/pep-0012/#how-to-use-this-template), they mention (emphasis mine)

> For Standards Track PEPs, if your feature depends on the acceptance of some other currently **in-development PEP**, add a Requires header right after the Type header. The value should be the PEP number of the PEP yours depends on. Don’t add this header if your dependent feature is described in a Final PEP.

jezdez: Good point, removed it. CEP 36 is already accepted so `Requires` doesn't apply here.

wolfv:
I have thought a bit about this, and sigstore uses something interesting called a "timestamp authority", which we could theoretically also investigate. If we have proper sigstore support, we'd get the timestamp for free from the attestation btw. 

The following was written by Claude because it can explain it better than me anyways :)

`indexed_timestamp` as proposed is a plain integer set by the channel server. This is better than `timestamp` (builder-controlled), but it's still a single party's unverifiable claim. If the server is compromised, or data is restored from backup, or a channel re-indexes, the value can silently change. There's no way for a client to independently verify that the timestamp is authentic.

## RFC 3161 Trusted Timestamps

RFC 3161 defines a protocol where a Trusted Third Party (a Timestamp Authority / TSA) issues signed timestamps:

1. You compute a hash of the data you want to timestamp
2. You send that hash to a TSA (e.g., Sigstore's `timestamp.sigstore.dev`, FreeTSA, DigiCert)
3. The TSA returns a signed Timestamp Response (TSR) containing the hash, the current time, and the TSA's signature
4. Anyone can later verify the timestamp by checking the TSA's signature against its certificate chain

The TSA never sees your data (only the hash), and the signed timestamp is cryptographically bound to that specific hash at that specific time. You can't backdate it or forge it without the TSA's private key.

This is already used in the sigstore ecosystem — the sigstore bundle format has a `TimestampVerificationData` field specifically for RFC 3161 timestamps, and tools like cosign support `--timestamp-server-url`.

## How this could work for conda packages

There are two complementary places where a TSA timestamp adds value:

### Build-time TSA timestamp (embedded in the package)

At build time (e.g., in `rattler-build` or `conda-build`), after assembling the package contents but before creating the final archive:

1. Compute a deterministic content hash over the package contents — e.g. a hash of a canonicalized manifest of `(path, sha256, size)` tuples, excluding the TSR field itself
2. Send that hash to a TSA
3. Embed the returned TSR in `info/index.json` (e.g. as `"build_tsr": "<base64>"`) or as a sidecar file `info/build.tsr`
4. Pack the `.conda` archive

Note: we can't hash the final archive itself because embedding the TSR changes the archive hash (chicken-and-egg). But a content hash works well — at verification time, a client unpacks the package, recomputes the content hash using the same deterministic scheme (skipping the TSR field), and verifies the TSA's signature. This is analogous to how git signed commits work.

This gives a cryptographically verifiable lower bound: "this package content existed no later than time T."

### Index-time TSA timestamp (alongside repodata)

When a channel server indexes a package, it could:

1. Hash the artifact (the `.conda` file hash is known at this point)
2. Request a signed timestamp from a TSA
3. Store the TSR alongside `indexed_timestamp` in repodata, or as a sidecar file

This makes `indexed_timestamp` independently verifiable rather than being a trust-me claim from the server.

## Relationship to this CEP

I'm not suggesting this CEP needs to specify all of this — `indexed_timestamp` as a simple queryable integer in repodata is useful on its own for `exclude-newer` and dependency cooldowns. But it might be worth:

1. Mentioning RFC 3161 TSA as a future direction for making `indexed_timestamp` cryptographically verifiable
2. Considering whether the CEP should reserve a field (e.g. `indexed_tsr`) for an optional base64-encoded TSR that proves the `indexed_timestamp` value
3. Noting that build tools could independently embed TSA timestamps in packages to provide a verifiable build time, complementing the server-set index time

For context, the sigstore integration in `rattler-build` already touches this infrastructure — the sigstore bundle format has native support for RFC 3161 timestamps via `TimestampVerificationData`. Extending this to also provide a verifiable build timestamp in `index.json` would be a natural next step.

Happy to discuss further or help spec this out if there's interest.

jezdez:
Thanks @wolfv, this is a great point. You're right that `indexed_timestamp` as proposed is a single-party claim and therefore not independently verifiable. RFC 3161 TSA is the standard tool to close that gap, and Sigstore's `timestamp.sigstore.dev` was added precisely because relying on Rekor's internal clock had the same trust problem.

I added a "Future work" section to the CEP that documents two directions:

- RFC 3161 TSA at index time, including the existing Sigstore bundle carrier (`TimestampVerificationData`) as a natural integration point for an attestation-aware indexer per [CEP 27](https://conda.org/learn/ceps/cep-0027).
- Build-time TSA timestamps as separate work that would more naturally amend [CEP 34](https://conda.org/learn/ceps/cep-0034) than this CEP.

I deliberately did not reserve a specific field name (e.g. `indexed_tsr`). There's no precedent for reserving fields in our CEPs, [CEP 27](https://conda.org/learn/ceps/cep-0027) follows the same "future work without reservation" pattern, and [CEP 36](https://conda.org/learn/ceps/cep-0036)'s additive convention means new fields can be added later without needing reservation. That also leaves room for a future CEP to pick the right design (sidecar file, sigstore bundle reference, in-toto attestation tied to indexing, ...) rather than baking in an RFC 3161 TSR encoding now.

For this CEP I'd like to keep the scope on the simple integer field to unblock `--exclude-newer` parity with uv, pip, pixi, and rattler, and dependency cooldowns. Happy to collaborate on a follow-up CEP for the verifiable-timestamp work whenever there's interest. Your sigstore expertise would be invaluable there.

jezdez:
Dear @conda/steering-council, this CEP is officially entering its Request For Comments (RFC) period. This will last two weeks, until May 12th EOD AoE.

jezdez:
I've updated the conda and conda plugin PRs and made another pass over the CEP. I think it's ready for a vote.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Motivation
+
+The conda ecosystem currently lacks a reliable, per-artifact indicator of when a package was made available on a channel. This gap has practical consequences for two emerging use cases: supply chain security and environment reproducibility.
+
+### The `timestamp` field is builder-controlled
+
+[CEP 34](./cep-0034.md) defines `timestamp` in `info/index.json` as "Starting time of the package build", expressed as Unix time in milliseconds. This value propagates into `repodata.json` as specified by [CEP 36](./cep-0036.md). Because it is set by the build tool (`conda-build`, `rattler-build`), a package author or malicious actor can set it to any value. The `common-1` schema at schemas.conda.org also marks this field as `patchable: true`.
```
jaimergp: > The `common-1` schema at schemas.conda.org also marks this field as `patchable: true`.

This doesn't really change much. All repodata record fields are patchable by today's standards. You can leave it as a side not but I wouldn't want that to become a precedent for something we hold true as an ecosystem.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### `channeldata.json` is too coarse
+
+[CEP 38](./cep-0038.md) specifies a `timestamp` field in `channeldata.json`, defined as "Upload date of the most recently published artifact". While this is server-controlled, it is per-package-name, not per-artifact. It only reflects the most recent upload and cannot be used to determine when a specific version or build was published.
```
jaimergp: > While this is server-controlled

Is it? I think that's an oversight in CEP 38. `conda-index` simply t[akes the largest record `timestamp` for the name](https://github.com/conda/conda-index/blob/90ea92e548316a8200dcb4d8c887fb3e13cf6d5d/conda_index/index/__init__.py#L1244-L1249), but it's not guaranteed that it's the upload time.

We should patch CEP 38, not sure if we can / should do it here. WDYT?

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+- `uv --exclude-newer` and `pip --exclude-newer` for reproducible resolution
+- Dependency cooldown features that delay installation of recently published packages as a supply chain security measure
+
+Within the conda ecosystem, pixi already provides an [`exclude-newer`](https://pixi.prefix.dev/latest/reference/pixi_manifest/#exclude-newer-optional) option, but it currently relies on the builder-controlled `timestamp` field. The conda ecosystem would benefit from a server-controlled equivalent that cannot be set by the package author.
+
+### Channel models have different risk profiles
+
+The value of a server-controlled index timestamp varies by channel model. Curated channels like Anaconda's defaults already have human review gates before packages reach users, similar to the Debian model. Community forges like conda-forge release at the speed of volunteer CI, where packages can go from merged PR to indexed artifact in minutes. Cooldown support is most critical for these faster-moving channels.
```
```suggestion
The value of a server-controlled index timestamp varies by channel model. Curated channels like Anaconda's `defaults` already have human review gates before packages reach users, similar to the Debian model. Community forges like conda-forge release at the speed of volunteer CI, where packages can go from merged PR to indexed artifact in minutes. Cooldown support is most critical for these faster-moving channels.
```

Or call it `main`.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Package record metadata
+
+[CEP 36](./cep-0036.md) defines the schema for entries in `packages` and `packages.conda` within `repodata.json`. This CEP adds one optional field to the package record metadata:
+
+- `indexed_timestamp: int`. Optional. Unix time in milliseconds when the artifact was added to the channel index. It MUST be set by the channel server or indexing tool, not by the build tool. If absent, clients MAY fall back to `timestamp` for time-based operations.
+
+### Channel server requirements
+
+Channel servers and indexing tools (such as `conda-index`, `quetz`, or Anaconda's infrastructure) SHOULD set `indexed_timestamp` on each package record when the artifact is first indexed into the channel. The value MUST reflect the actual time the artifact became available in the channel, not the build time or any value from the artifact itself. The `repodata-record-1` schema at schemas.conda.org SHOULD be updated accordingly.
```
jaimergp: > The `repodata-record-1` schema at schemas.conda.org SHOULD be updated accordingly.

So far, CEPs have been blind to `schemas.conda.org`, and I consider them "downstream" from the CEP process. As in, this is unnecessary to add in the specification. Similar to saying "documentation mentioning this field should be updated accordingly". Feel free to remove or move to an appendix of recommendations / action items.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Client behavior
+
+Conda clients that implement time-based filtering (e.g., dependency cooldowns, exclude-newer) SHOULD prefer `indexed_timestamp` when present. If `indexed_timestamp` is absent, clients MAY fall back to `timestamp` with appropriate documentation that the value is builder-controlled. Clients MAY support global, channel-scoped, package-scoped, or other policy scopes, but each policy should apply the same per-record effective timestamp.
+
+## Rationale
+
+### Why per-record, not per-package
+
+`channeldata.json` (CEP 38) already tracks an upload date, but at the package-name level. Different versions and builds of the same package are published at different times, so a per-record field is necessary for meaningful time-based filtering.
```
jaimergp: Same here, CEP 38 is sadly wrong here. Need to update.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Why `indexed_timestamp`, not `upload_timestamp`
+
+The field measures when the artifact appeared in the channel index, not when it was uploaded. In many conda workflows, these are distinct events. For example, conda-forge uploads built artifacts to a staging channel (`cf-staging`), validates them, and only then copies them to the production `conda-forge` channel where they are indexed into `repodata.json`. The `conda-index` maintainer also describes this as tracking when a package was "first indexed" rather than when it was uploaded.
+
+The name `indexed_timestamp` accurately reflects this, and fits naturally alongside the existing `timestamp` field in the package record schema.
+
+### Not subject to repodata patching
+
+Because `indexed_timestamp` is a server-controlled fact about when an artifact entered the index, it should not be modified by repodata patching. Repodata patches are intended for correcting package-side metadata (such as dependency constraints), not for altering server-side provenance.
```
jaimergp: May want to use capital letters here:

```suggestion
Because `indexed_timestamp` is a server-controlled fact about when an artifact entered the index, it SHOULD NOT be modified by repodata patching. Repodata patches are intended for correcting package-side metadata (such as dependency constraints), not for altering server-side provenance.
```

And I'd argue it should be a MUST NOT, but we don't have a CEP for this, yet. Maybe for the next batch of standards? Sooon :D

cep-0047.md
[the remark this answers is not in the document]
jezdez: Kept this as `SHOULD` in the validation section rather than `MUST`: `indexed_timestamp` itself is server-set as `MUST`, but rejecting existing artifacts based on the older builder-controlled `timestamp` field is a compatibility tightening. Current text says indexers SHOULD validate no future `timestamp` and `timestamp <= indexed_timestamp`, and reject or flag violations.

cep-XXXX.md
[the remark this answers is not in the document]
jezdez: Yes. The migration text now allows pre-CEP artifacts to seed `indexed_timestamp` from `timestamp` or another server-side signal. For artifacts first indexed after approval, `indexed_timestamp` MUST be set at indexing time. That preserves existing exclude-newer behavior without letting new packages choose their own index time.

cep-XXXX.md
jezdez: Agreed. Removed the `schemas.conda.org` / `patchable` sentence in 4edc038. The argument now rests on CEP 34/36: `timestamp` is builder-provided and therefore unsuitable as a server-controlled publication signal.

cep-XXXX.md
jezdez: Agreed. Reworded in 4edc038: the CEP now says CEP 38 defines the field that way, but current indexers do not reliably provide that semantic; for example, `conda-index` derives it from package record timestamps. I think CEP 38 should be patched separately rather than trying to fix it here.

cep-XXXX.md
jezdez: Adopted in substance in 4edc038, including backticks around `defaults`. I kept the conda-forge wording as `conda-forge` here; the later rationale already distinguishes staging from the production `conda-forge` channel.

cep-XXXX.md
jezdez: Agreed. This stays out of the specification. In 4edc038 I also removed the other `schemas.conda.org` mention, so schema updates remain downstream of the CEP rather than normative text here.

cep-XXXX.md
jezdez: Agreed. Same resolution as above: the CEP no longer calls the `channeldata.json` timestamp server-controlled. It now notes that current indexers derive it from package record timestamps. Updating CEP 38 should be a separate cleanup.

cep-XXXX.md
jezdez: Applied `SHOULD NOT` in 4edc038. I agree `MUST NOT` is the direction, but repodata patching semantics are bigger than this one field and probably deserve their own standards pass.

cep-XXXX.md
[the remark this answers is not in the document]
jaimergp: Maybe redundant, but the value of this field must not change regardless the source, right? Otherwise we get clunky solves if extreme `exclude-newer` configs are used. Thinking of filesystems changing `mtime` of some files by accident, and that value used for `indexed_timestamp`, for example. Shall we record that decision?

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### The `timestamp` field is builder-controlled
+
+[CEP 34](./cep-0034.md) defines `timestamp` in `info/index.json` as "Starting time of the package build", expressed as Unix time in milliseconds. This value propagates into `repodata.json` as specified by [CEP 36](./cep-0036.md). Because it is set by the build tool (`conda-build`, `rattler-build`), a package author or malicious actor can set it to any value.
+
+### `channeldata.json` is too coarse
+
+[CEP 38](./cep-0038.md) specifies a `timestamp` field in `channeldata.json`, defined as "Upload date of the most recently published artifact". Current indexer implementations do not reliably provide that exact semantic; for example, `conda-index` derives the value from package record timestamps. Even if corrected, `channeldata.json` is per-package-name, not per-artifact. It only reflects the most recent publication event and cannot be used to determine when a specific version or build was published.
```
jaimergp: Let's update this definition assuming https://github.com/conda/ceps/pull/172 will pass.

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Backwards compatibility
+
+Adding an optional field to existing package records in `packages` and `packages.conda` is a purely additive change. [CEP 36](./cep-0036.md) states that additional keys "SHOULD NOT be present and SHOULD be ignored", meaning older clients will silently ignore `indexed_timestamp`. This does not require the repodata revision mechanism proposed in [conda/ceps#146](https://github.com/conda/ceps/pull/146), though it is consistent with that proposal's philosophy of backwards-compatible evolution.
+
+This follows the same pattern as the `url` field proposed in [conda/ceps#151](https://github.com/conda/ceps/pull/151), which also adds an optional per-record field to the existing schema.
+
+### Why milliseconds
+
+For consistency with the existing `timestamp` field (CEP 34) and `channeldata.json` timestamp (CEP 38), both of which use Unix time in milliseconds.
```
jaimergp: Note that CEP 38's timestamp is in seconds.

See https://github.com/conda/ceps/pull/172

cep-XXXX.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Client behavior
+
+Conda clients that implement time-based filtering (e.g., dependency cooldowns, exclude-newer) SHOULD prefer `indexed_timestamp` when present. If `indexed_timestamp` is absent, clients MAY fall back to `timestamp` with appropriate documentation that the value is builder-controlled. Clients MAY support global, channel-scoped, package-scoped, or other policy scopes, but each policy should apply the same per-record effective timestamp.
+
+## Rationale
+
+### Why per-record, not per-package
+
+`channeldata.json` (CEP 38) already tracks an upload date, but at the package-name level. Different versions and builds of the same package are published at different times, so a per-record field is necessary for meaningful time-based filtering.
```
jaimergp: Same here (again :D), let's assume https://github.com/conda/ceps/pull/172 will pass.

cep-XXXX.md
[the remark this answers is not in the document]
jezdez: I agree, this invariant is worth making explicit. Suggested wording:

```suggestion
For artifacts that predate this CEP and lack an `indexed_timestamp`, channel servers MAY seed the value from the artifact's `timestamp` field or from another server-side signal such as file modification time. This is acceptable because the security benefits of `indexed_timestamp` (dependency cooldowns, exclude-newer) protect against the initial publication window, which has long closed for existing artifacts.

Once `indexed_timestamp` has been assigned for a package record, channel servers and indexing tools MUST preserve that value across subsequent indexing runs. They MUST NOT recompute it from file modification time, the artifact's `timestamp`, or any other fallback signal while an existing `indexed_timestamp` is present. Channel operators MAY correct demonstrably erroneous values as an intentional metadata correction.
```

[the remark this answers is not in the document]
jaimergp: Perfect!

cep-XXXX.md
jezdez: Suggested update based on conda/ceps#172:

```suggestion
[CEP 38](./cep-0038.md) specifies a `timestamp` field in `channeldata.json`, defined as the most recent `timestamp` value of all records for a package name, expressed as POSIX time in seconds. This is per-package-name, not per-artifact, and it is derived from builder-controlled package record timestamps rather than a server-controlled publication time. It cannot be used to determine when a specific version or build was published.
```

cep-XXXX.md
jezdez: Suggested update based on conda/ceps#172:

```suggestion
`channeldata.json` (CEP 38) already includes a timestamp, but at the package-name level and derived from package record timestamps. Different versions and builds of the same package are published at different times, so a per-record field is necessary for meaningful time-based filtering.
```

cep-XXXX.md
jezdez: Suggested update based on conda/ceps#172:

```suggestion
For consistency with the existing package record `timestamp` field defined by CEP 34, which uses Unix time in milliseconds. The `channeldata.json` timestamp defined by CEP 38 uses seconds, but `indexed_timestamp` belongs to package records in `repodata.json` and should follow the package record convention.
```

jaimergp:
Dear @conda/steering-council, 

The vote for this CEP has started. It will be open for two weeks, until **June 24th, 2026, 23:59 Anywhere on Earth**. This time period has been chosen to make it eligible for time-out rules. As an Enhancement Proposal vote, it requires 60% affirmative votes to pass. 

- Uwe Korn (@xhochy) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- Cheng H. Lee (@chenghlee) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Filipe Fernandes (@ocefpaf) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Michael Sarahan (@msarahan) 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- John Kirkham (@jakirkham) 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- Jannis Leidel (@jezdez) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Wolf Vollprecht (@wolfv) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Bas Zalmstra (@baszalmstra) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Hind Montassif (@Hind-M) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Pavel Zwerschke (@pavelzw) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain

jaimergp:
@mariusvniekerk, @chenghlee, @msarahan, @jakirkham: final reminder to vote, thanks!

jaimergp:

* SC size: 13
* Votes: 10 (76.9% of SC)
  * Yes: 10 (100% of votes)
  * No: 0
  * Abstain: 0
  
@conda/steering-council, this will be minted as CEP 47.

jaimergp (approved):

jaimergp:
Congrats, @jezdez!

PR #159, opened 2026-04-07
CEP XXXX: Defining Package-URLs (PURLs) for conda packages

chenghlee:
## Checklist for submitter

- [x] I am submitting a new CEP: [Package-URLs (PURLs) for conda packages](https://github.com/chenghlee/ceps/blob/conda-purl-definition/cep-XXXX.md).

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -5,8 +5,8 @@
 <tr><td> Status </td><td> Draft </td></tr>
 <tr><td> Author(s) </td><td> Cheng H. Lee &lt;clee@anaconda.com&gt; </td></tr>
 <tr><td> Created </td><td> 2026-04-02 </td></tr>
-<tr><td> Updated </td><td> 2026-04-02 </td></tr>
-<tr><td> Discussion </td><td> ... </td></tr>
+<tr><td> Updated </td><td> 2026-04-08 </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159 </td></tr>
 <tr><td> Implementation </td><td> NA </td></tr>
 </table>
 
@@ -46,7 +46,7 @@ Edition (December 2025) of that standard.
 
 ### ECMA-427 PURL components
 
-ECMA-427 Section 5 ("Package-URL specification") states:
+ECMA-427 Clause 5 ("Package-URL specification") states:
 
 > A PURL is a URL composed of seven components:  
 >   scheme:type/namespace/name@version?qualifiers#subpath
@@ -55,16 +55,16 @@ For conda PURLs, these seven components are defined as follows:
 
 #### `scheme` component
 
-In accordance with ECMA-427 Section 5.6.1 ("Scheme"), the `scheme` component for a conda PURL is
+In accordance with ECMA-427 Clause 5.6.1 ("Scheme"), the `scheme` component for a conda PURL is
 the ASCII string literal "pkg".  Systems handling conda PURLs SHOULD also follow the other rules
-for the `scheme` component specified in ECMA-427 Section 5.6.1.
+for the `scheme` component specified in ECMA-427 Clause 5.6.1.
 
 #### `type` component
 
-In accordance with ECMA-427 Section 5.6.2 ("Type"), the canonical form for a conda PURL's `type`
+In accordance with ECMA-427 Clause 5.6.2 ("Type"), the canonical form for a conda PURL's `type`
 component is the ASCII string literal "conda".  The `type` component MUST be treated as
 case-insensitive, and system handling conda PURLs are encouraged to normalize the `type` component
-to its canonical lowercase form, following the behaviors specified in ECMA-427 Section 5.5.
+to its canonical lowercase form, following the behaviors specified in ECMA-427 Clause 5.5.
 
 #### `namespace` component
 
@@ -73,16 +73,16 @@ components of the [channel base URL defined in CEP 26](./cep-0026.md#channel-bas
 concept most users have when using the `channel` configuration option supported by [most]
 conda-compatible tools.
 
-For conda PURLs, the ECMA-427 Section 6.5.1 requirement property for the `namespace` component is
+For conda PURLs, the ECMA-427 Clause 6.5.1 requirement property for the `namespace` component is
 optional.  However, the use or non-use of the `namespace` component within a conda PURL MUST
 conform to the following rules:
 
 - If the `namespace` component is omitted from a conda PURL, then the `repository_url` qualifier
   MUST be explicitly provided in the PURL, and its value MUST be a valid channel base URL, as
   defined in CEP 26.
-- If the `namespace` component is provided, each ECMA-427 Section 5.6.3 `namespace` segment MUST be
+- If the `namespace` component is provided, each ECMA-427 Clause 5.6.3 `namespace` segment MUST be
   a valid channel base URL path component, as defined in CEP 26.  All other rules for the
-  `namespace` component specified in ECMA-417 Section 5.6.3 apply to conda PURLs, including the
+  `namespace` component specified in ECMA-417 Clause 5.6.3 apply to conda PURLs, including the
   requirement that segments within a PURL `namespace` MUST be separated by a single unencoded
   ASCII "/" (slash) character.
 - If the `namespace` component is provided, then the concatenation of the `repository_url`
@@ -175,7 +175,7 @@ The following `qualifiers` are defined for conda PURLs:
 
   The lowercase hex-encoded checksum value MUST only contain the ASCII lowercase characters `a-f`
   and digits `0-9`.  The checksum value is explicitly understood to be a hexadecimal value, so
-  commonly-used prefixes and postfixes (e.g., "0x" and "_16", respectively) MUST NOT be included as
+  commonly-used prefixes and postfixes (e.g., "0x" and "\_16", respectively) MUST NOT be included as
   part of the checksum value.
 
   Historically, the conda ecosystem has used the MD5 and SHA-256 algorithms for checksumming
@@ -183,11 +183,11 @@ The following `qualifiers` are defined for conda PURLs:
   sub-optimal security properties of MD5, the RECOMMENDED algorithm for the `checksum` qualifier is
   SHA-256.
 
-Qualifiers in conda PURLs SHALL conform with all the rules specified ECMA-427 Section 5.6.6
-("Qualifiers"), with the key words "shall", "shall not", and "may" in ECMA-427 Section 5.6.6 being
+Qualifiers in conda PURLs SHALL conform with all the rules specified ECMA-427 Clause 5.6.6
+("Qualifiers"), with the key words "shall", "shall not", and "may" in ECMA-427 Clause 5.6.6 being
 interpreted as their BCP 14 equivalents "SHALL", "SHALL NOT", and "MAY", respectively.
 Importantly, this means that the entirety of any `repository_url` value, including colons (`:`) and
-slashes (`/`), must be percent-encoded in accordance with ECMA-427 Section 5.4.
+slashes (`/`), must be percent-encoded in accordance with ECMA-427 Clause 5.4.
 
 Since existing CEPs do not define a strict relationship between a package artifact's build string
 and build number, this CEP does not require that the `build` and `build_number` qualifiers in a
@@ -216,19 +216,19 @@ rules defined in CEP 26.
 The optional `subpath` component of a conda PURL is used to specify a path within the identified
 package(s), subject to the following:
 
-- The `subpath` component SHALL confirm to all rules specified in ECMA-427 Section 5.6.7
-  ("Subpath"), with the key words "shall", "shall not", and "may" in ECMA-427 Section 5.6.7 being
+- The `subpath` component SHALL confirm to all rules specified in ECMA-427 Clause 5.6.7
+  ("Subpath"), with the key words "shall", "shall not", and "may" in ECMA-427 Clause 5.6.7 being
   interpreted as their BCP 14 equivalents "SHALL", "SHALL NOT", and "MAY", respectively.
 - The `subpath` component MUST be treated as case-sensitive.
 - If the `subpath` component identifies a symbolic link with the package(s), the `subpath`
   component MUST be interpreted as a reference to the symbolic link file itself and _not_ as a
   reference to the target path after the symbolic link has been partially or fully dereferenced.
 - When applied to [`tar.bz2`-/`v1`-format packages][CEP35.v1], "relative to the root of the
-  package" in ECMA-427 Section 5.6.7 MUST be interpreted as the path to an archive member within
+  package" in ECMA-427 Clause 5.6.7 MUST be interpreted as the path to an archive member within
   the package tarball, or equivalently, as a relative path from a root directory into which the
   package tarball was extracted.
 - When applied to [`conda`-/`v2`-format packages][CEP35.v2], "relative to the root of the package"
-  in ECMA-427 Section 5.6.7 MUST be interpreted as a relative path from a common root directory
+  in ECMA-427 Clause 5.6.7 MUST be interpreted as a relative path from a common root directory
   into which both inner archives (i.e., `info-{name}-{version}-{build}.tar.zst` and
   `pkg-{name}-{version}-{build}.tar.zst`) have been extracted.
 - The `subpath` component MUST be interpreted as a reference to a path as it exist in the package
@@ -243,21 +243,21 @@ archives within `conda`-/`v2`-format packages.
 
 ### Other ECMA-427 root object properties
 
-ECMA-427 Section 6 ("Package-URL Type Definition Schema") specifies other required and optional
+ECMA-427 Clause 6 ("Package-URL Type Definition Schema") specifies other required and optional
 properties for a PURL type definition.  For conda PURLs, these properties are defined as follows:
 
 #### `type name` property
 
-The ECMA-427 Section 6.2 `type name` property of the conda PURL type definition SHALL be "Conda".
+The ECMA-427 Clause 6.2 `type name` property of the conda PURL type definition SHALL be "Conda".
 
 #### `description` property
 
-The ECMA-427 Section 6.3 `description` property of the conda PURL type definition SHALL be
+The ECMA-427 Clause 6.3 `description` property of the conda PURL type definition SHALL be
 "Conda packages".
 
 #### `repository` property
 
-The ECMA-427 Section 6.4 `repository` property is REQUIRED for conda PURLs.  Its use and
+The ECMA-427 Clause 6.4 `repository` property is REQUIRED for conda PURLs.  Its use and
 interpretation is defined in the "`namespace` component" and "`qualifier` component" sections.
 
 To maintain compatibility with the default behavior of existing conda-compatible clients, the
@@ -266,17 +266,17 @@ default public repository (i.e., `default_repository_url`) for conda packages is
 
 #### `examples` property
 
-The ECMA-427 Section 6.10 `PURL examples` property of the conda PURL type definition SHALL consist
+The ECMA-427 Clause 6.10 `PURL examples` property of the conda PURL type definition SHALL consist
 of the PURLs listed in the "Examples" section of this CEP.
 
 #### `reference_urls` property
 
-The ECMA-427 Section 6.12 `reference_urls` property for the conda PURL type definition SHALL be a
+The ECMA-427 Clause 6.12 `reference_urls` property for the conda PURL type definition SHALL be a
 list consisting of:
 
 - The canonical URL of this CEP
 
-Note that ECMA-427 Section 6.12 specifies that the URLs listed in the `reference_urls` PURL type
+Note that ECMA-427 Clause 6.12 specifies that the URLs listed in the `reference_urls` PURL type
 definition are informative.  However, to ensure standardized implementations, this CEP SHOULD be
 treated as normative and not just informative.
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,374 @@
+# CEP XXXX - PURLs for conda packages
+
+<table>
+<tr><td> Title </td><td> Package-URLs (PURLs) for conda packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Cheng H. Lee &lt;clee@anaconda.com&gt; </td></tr>
+<tr><td> Created </td><td> 2026-04-02 </td></tr>
+<tr><td> Updated </td><td> 2026-04-02 </td></tr>
+<tr><td> Discussion </td><td> ... </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+
+## Abstract
+
+This CEP describes the Package-URL type definition for conda packages in a way that conforms to
+other relevant standards, including other CEPs and ECMA-427.
+
+
+## Motivation
+
+Package URLs (PURLs) provide "a standardized, URL-based syntax that uniquely identifies software
+packages, independent of their ecosystem or distribution channel".  Users may want or need to use
+PURLs to identify conda packages in various contexts, including but not limited to, software bills
+of materials (SBOMs), vulnerability reporting, and cross-ecosystem compatibility.
+
+However, the [existing conda PURL definition][purl-conda-def] fails to properly capture the
+multi-stakeholder nature of the conda ecosystem, various commonly-understood concepts within the
+ecosystem, and critically, the standards specified in other accepted CEPs.  This CEP seeks to
+address such shortcomings with an ecosystem-approved specification for the conda PURL definition.
+
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in BCP 14 ([RFC2119][RFC2119], [RFC8174][RFC8174]) when, and only when, they appear in
+all capitals, as shown here.
+
+For the purposes of this CEP, the phrase "existing conda PURL definition" refers to the [commit
+a6c97bc version of `types/conda-definition.json`][purl-conda-def], which is canonically available
+in the https://github.com/package-url/purl-spec repository.
+
+For the purposes of this CEP, references to sections of [ECMA-427][ECMA427] are relative to the 1st
+Edition (December 2025) of that standard.
+
+### ECMA-427 PURL components
+
+ECMA-427 Section 5 ("Package-URL specification") states:
+
+> A PURL is a URL composed of seven components:  
+>   scheme:type/namespace/name@version?qualifiers#subpath
+
+For conda PURLs, these seven components are defined as follows:
+
+#### `scheme` component
+
+In accordance with ECMA-427 Section 5.6.1 ("Scheme"), the `scheme` component for a conda PURL is
+the ASCII string literal "pkg".  Systems handling conda PURLs SHOULD also follow the other rules
+for the `scheme` component specified in ECMA-427 Section 5.6.1.
+
+#### `type` component
+
+In accordance with ECMA-427 Section 5.6.2 ("Type"), the canonical form for a conda PURL's `type`
+component is the ASCII string literal "conda".  The `type` component MUST be treated as
+case-insensitive, and system handling conda PURLs are encouraged to normalize the `type` component
+to its canonical lowercase form, following the behaviors specified in ECMA-427 Section 5.5.
+
+#### `namespace` component
+
+Roughly speaking, the `namespace` component of a conda PURL corresponds to the path and label name
+components of the [channel base URL defined in CEP 26](./cep-0026.md#channel-base-urls) and/or the
+concept most users have when using the `channel` configuration option supported by [most]
+conda-compatible tools.
+
+For conda PURLs, the ECMA-427 Section 6.5.1 requirement property for the `namespace` component is
+optional.  However, the use or non-use of the `namespace` component within a conda PURL MUST
+conform to the following rules:
+
+- If the `namespace` component is omitted from a conda PURL, then the `repository_url` qualifier
+  MUST be explicitly provided in the PURL, and its value MUST be a valid channel base URL, as
+  defined in CEP 26.
+- If the `namespace` component is provided, each ECMA-427 Section 5.6.3 `namespace` segment MUST be
+  a valid channel base URL path component, as defined in CEP 26.  All other rules for the
+  `namespace` component specified in ECMA-417 Section 5.6.3 apply to conda PURLs, including the
+  requirement that segments within a PURL `namespace` MUST be separated by a single unencoded
+  ASCII "/" (slash) character.
+- If the `namespace` component is provided, then the concatenation of the `repository_url`
+  qualifier, a single unencoded ASCII "/" (slash) character, and the `namespace` component MUST be
+  a valid channel base URL, as defined in CEP 26.
+- The string literals "defaults" and "local" MUST NOT be used as the entire value of the
+  `namespace` component.  This prohibition exists to minimize potential confusion with the
+  reserved, user-configurable [multi-]channel names supported by many conda-compatible clients.
+- The final path component of `namespace` SHOULD NOT match any channel subdir name, as this can
+  lead to ambiguous PURLs and channel base URLs, as noted in CEP 26.  If platform information is
+  needed to properly identify the package(s), then the `subdir` qualifier (defined below) should be
+  used instead.
+
+An important implication of these rules is that a valid conda PURL must provide at least one of the
+`namespace` component or the `repository_url` qualifier, as a conda PURL that both omits the
+`namespace` component _and_ assumes the default value for the `repository_url` qualifier (defined
+below) specifies a channel base URL that is insufficient to properly locate the package.
+
+Further, the prohibition on using "defaults" as the `namespace` component value means that PURLs
+for Anaconda-built packages should use "main", "r", or "msys2" for their `namespace` component.
+
+#### `name` component
+
+The `name` component MUST be the [distributable package name](./cep-0026.md#package-names) of the
+identified artifact(s), as defined in CEP 26.
+
+Conda PURLs SHOULD NOT be used to identify conda virtual packages since they do not correspond to
+any concrete [package] artifact and only exist on the client side.
+
+#### `version` component
+
+The `version` component MUST be the [package version string](./cep-0026.md#version-strings), as
+defined in CEP 26.
+
+#### `qualifiers` component
+
+The following `qualifiers` are defined for conda PURLs:
+
+- `repository_url`: Optional qualifier, with a default value defined in the "`repository` property"
+  section of this CEP.  Its value, when combined with any `namespace` component as described in the
+  "`namespace` component" section, MUST specify the channel base URL of the identified package(s),
+  as defined in CEP 26.
+- `build`: Optional qualifier, with no default value.  If provided, its value MUST be the
+  [build string](./cep-0026.md#build-strings), as defined in CEP 26, of the identified package(s).
+- `build_number`: Optional qualifier, with no default value.  If provided, its value MUST the
+  [build number](./cep-0034.md#infoindexjson), as defined in CEP 34, of the identified package(s).
+- `subdir`: Optional qualifier, with no default value.  If provided, its value MUST be the
+  [channel subdir name](./cep-0026.md#subdir-names), as defined in CEP 26, of the identified
+  package(s).
+- `type`: Optional qualifier, with no default value.  If provided, its value MUST be exactly one of
+  [currently-recognized artifact extensions](./cep-0026.md#artifact-extensions) or its versioned
+  synonym, as defined in CEP 26.  At the time of this CEP's adoption, the accepted values are
+  `tar.bz2` (versioned synonym: `v1`), or `conda` (versioned synonym: `v2`).
+- `checksum`:  Optional qualifier, with no default value.  If provided, its value MUST consist of
+  one or more checksum specifications, with consecutive checksum specifications separated by a
+  single, unencoded ASCII "," (comma) character.  Each checksum specification is defined as the
+  concatenating of the lowercase checksum algorithm name (e.g., "md5", "sha256"); a single,
+  unencoded ":" (colon) character; and the lowercase, hex-encoded checksum value.
+
+  The lowercase checksum algorithm name MUST only contain the ASCII lowercase characters `a-z` and
+  digits `0-9`.  The recommended way to specify checksum algorithms whose names are commonly
+  written with upper case and/or non-alphanumeric characters such as dashes (e.g., "SHA-1",
+  "SHA-256") is to simply remove any non-alphanumeric characters in the commonly-written name and
+  lowercase the remaining string.
+
+  The lowercase hex-encoded checksum value MUST only contain the ASCII lowercase characters `a-f`
+  and digits `0-9`.  The checksum value is explicitly understood to be a hexadecimal value, so
+  commonly-used prefixes and postfixes (e.g., "0x" and "_16", respectively) MUST NOT be included as
+  part of the checksum value.
+
+  Historically, the conda ecosystem has used the MD5 and SHA-256 algorithms for checksumming
+  packages; see, e.g., [CEP 16][CEP16], [CEP 32][CEP32], and [CEP 36][CEP36].  Given this and the
+  sub-optimal security properties of MD5, the RECOMMENDED algorithm for the `checksum` qualifier is
+  SHA-256.
+
+Qualifiers in conda PURLs SHALL conform with all the rules specified ECMA-427 Section 5.6.6
+("Qualifiers"), with the key words "shall", "shall not", and "may" in ECMA-427 Section 5.6.6 being
+interpreted as their BCP 14 equivalents "SHALL", "SHALL NOT", and "MAY", respectively.
+Importantly, this means that the entirety of any `repository_url` value, including colons (`:`) and
+slashes (`/`), must be percent-encoded in accordance with ECMA-427 Section 5.4.
+
+Since existing CEPs do not define a strict relationship between a package artifact's build string
+and build number, this CEP does not require that the `build` and `build_number` qualifiers in a
+conda PURL be mutually exclusive.  Further, it does specify which of these two qualifiers takes
+precedence in cases where their simultaneous use in a conda PURL creates a conflict.  Producers of
+conda PURLs should take care that their use of these qualifiers does not create a conflict, and
+consumers of conda PURLs are RECOMMENDED to treat conflicting use of these qualifiers as an invalid
+or erroneous PURL.
+
+This CEP explicitly _undefines_ the `channel` qualifier found in the [existing conda PURL
+definition][purl-conda-def].  The `channel` qualifier corresponds to a weakly-defined concept not
+properly captured in any existing CEP, and its presumed purpose in the existing definition is
+better handled by the combination of the `namspace` component and `repository_url` qualifier, as
+described in this CEP.  Upon acceptance of this CEP, systems handling conda PURLs SHOULD treat
+PURLs containing a `channel` qualifier as invalid.
+
+To avoid the need for potentially complex precedence rules in this CEP, the [commonly-used
+qualifiers][purl-quals-guide] `filename` and `download_url` used by other PURL types are explicitly
+_not_ defined for conda PURLs.  In other words, the `filename` and `download_url` qualifiers MUST
+NOT be accepted when processing conda PURLs.  Should such values be needed to handle a conda PURL,
+they should be constructed from the other components and qualifiers defined in this CEP and the
+rules defined in CEP 26.
+
+#### `subpath` component
+
+The optional `subpath` component of a conda PURL is used to specify a path within the identified
+package(s), subject to the following:
+
+- The `subpath` component SHALL confirm to all rules specified in ECMA-427 Section 5.6.7
+  ("Subpath"), with the key words "shall", "shall not", and "may" in ECMA-427 Section 5.6.7 being
+  interpreted as their BCP 14 equivalents "SHALL", "SHALL NOT", and "MAY", respectively.
+- The `subpath` component MUST be treated as case-sensitive.
+- If the `subpath` component identifies a symbolic link with the package(s), the `subpath`
+  component MUST be interpreted as a reference to the symbolic link file itself and _not_ as a
+  reference to the target path after the symbolic link has been partially or fully dereferenced.
+- When applied to [`tar.bz2`-/`v1`-format packages][CEP35.v1], "relative to the root of the
+  package" in ECMA-427 Section 5.6.7 MUST be interpreted as the path to an archive member within
+  the package tarball, or equivalently, as a relative path from a root directory into which the
+  package tarball was extracted.
+- When applied to [`conda`-/`v2`-format packages][CEP35.v2], "relative to the root of the package"
+  in ECMA-427 Section 5.6.7 MUST be interpreted as a relative path from a common root directory
+  into which both inner archives (i.e., `info-{name}-{version}-{build}.tar.zst` and
+  `pkg-{name}-{version}-{build}.tar.zst`) have been extracted.
+- The `subpath` component MUST be interpreted as a reference to a path as it exist in the package
+  artifact(s), and _not_ as a reference to a path after any modifications that may have occurred
+  when the package artifact(s) are processed by conda-compatible clients (e.g., prefix replacement,
+  pre- or post-link scripts, etc.).
+
+The last three rules are intended to ensure consistent interpretation of the `subpath` component,
+regardless of the actual package format and processing system.  Note that these rules imply that
+conda PURLs cannot be used to directly identify the `metadata.json` nor the inner `.tar.zst`
+archives within `conda`-/`v2`-format packages.
+
+### Other ECMA-427 root object properties
+
+ECMA-427 Section 6 ("Package-URL Type Definition Schema") specifies other required and optional
+properties for a PURL type definition.  For conda PURLs, these properties are defined as follows:
+
+#### `type name` property
+
+The ECMA-427 Section 6.2 `type name` property of the conda PURL type definition SHALL be "Conda".
+
+#### `description` property
+
+The ECMA-427 Section 6.3 `description` property of the conda PURL type definition SHALL be
+"Conda packages".
+
+#### `repository` property
+
+The ECMA-427 Section 6.4 `repository` property is REQUIRED for conda PURLs.  Its use and
+interpretation is defined in the "`namespace` component" and "`qualifier` component" sections.
+
+To maintain compatibility with the default behavior of existing conda-compatible clients, the
+default public repository (i.e., `default_repository_url`) for conda packages is SHALL be
+"https://conda.anaconda.org".
+
+#### `examples` property
+
+The ECMA-427 Section 6.10 `PURL examples` property of the conda PURL type definition SHALL consist
+of the PURLs listed in the "Examples" section of this CEP.
+
+#### `reference_urls` property
+
+The ECMA-427 Section 6.12 `reference_urls` property for the conda PURL type definition SHALL be a
+list consisting of:
+
+- The canonical URL of this CEP
+
+Note that ECMA-427 Section 6.12 specifies that the URLs listed in the `reference_urls` PURL type
+definition are informative.  However, to ensure standardized implementations, this CEP SHOULD be
+treated as normative and not just informative.
+
+
+## Examples
+
+- `pkg:conda/main/python`: basic usage, identifying [all] `python` packages from Anaconda
+- `pkg:conda/conda-forge/python@3.13.12?build=hc97d973_100_cp313&subdir=linux-64`: basic usage,
+  identifying a specific `python` build from conda-forge.
+- `pkg:conda/python?repository_url=https%3A%2F%2Frepo.anaconda.com%2Fpkgs%2Fmain`: using the
+  `repository_url` qualifier to fully identify the channel base URL
+- `pkg:conda/conda-forge/python?repository_url=https%3A%2F%2Fprefix.dev`: using the
+  `repository_url` qualifier to identify packages hosted by a conda-forge mirror.
+
+
+## Backwards Compatibility
+
+This CEP intentionally breaks backwards-compatibility for PURLs based on the [existing conda PURL
+definition][purl-conda-def] in the following ways:
+
+- The `default_repository_url` has been changed from `https://repo.anaconda.com` to
+  `https://conda.anaconda.org`.
+- The requirement property of the `namespace` component has changed from "prohibited" to
+  "optional", with additional rules as described above.
+- `channel` is no longer an accepted qualifier for conda PURLs.
+
+These breaking changes are considered acceptable by the author(s) of this CEP, as the existing
+conda PURL definition fails to properly capture existing standards (i.e., CEPs) and commonly-used
+patterns in the conda ecosystem.  Among the motivations for these breaking changes:
+
+- Simple PURLs like `pkg:conda/python` and `pkg:conda/python?channel=conda-forge` are accepted
+  under the existing conda PURL definition but cannot be used to identify actual package artifacts,
+  as that definition's default values produces non-existent URLs.  The existing conda PURL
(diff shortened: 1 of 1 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,374 @@
...
+
+  The lowercase hex-encoded checksum value MUST only contain the ASCII lowercase characters `a-f`
+  and digits `0-9`.  The checksum value is explicitly understood to be a hexadecimal value, so
+  commonly-used prefixes and postfixes (e.g., "0x" and "_16", respectively) MUST NOT be included as
+  part of the checksum value.
+
+  Historically, the conda ecosystem has used the MD5 and SHA-256 algorithms for checksumming
+  packages; see, e.g., [CEP 16][CEP16], [CEP 32][CEP32], and [CEP 36][CEP36].  Given this and the
+  sub-optimal security properties of MD5, the RECOMMENDED algorithm for the `checksum` qualifier is
+  SHA-256.
```
chenghlee, line 196: Having thought about it for a bit, I'm open to rewriting this `checksum` qualifier specification to only support `MD5`, `SHA256`, and/or [some other finite set of checksum algorithms], mostly so this and other CEPs don't have to think about support PURLs using, e.g., [crc8](https://en.wikipedia.org/wiki/Cyclic_redundancy_check) or [xkcd221](https://xkcd.com/221/), as one of their named `checksum` algorithms.

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
+# CEP XXXX - PURLs for conda packages
+
+<table>
+<tr><td> Title </td><td> Package-URLs (PURLs) for conda packages </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Cheng H. Lee &lt;clee@anaconda.com&gt; </td></tr>
+<tr><td> Created </td><td> 2026-04-02 </td></tr>
+<tr><td> Updated </td><td> 2026-04-02 </td></tr>
+<tr><td> Discussion </td><td> ... </td></tr>
```
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159 </td></tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+
+Package URLs (PURLs) provide "a standardized, URL-based syntax that uniquely identifies software
+packages, independent of their ecosystem or distribution channel".  Users may want or need to use
+PURLs to identify conda packages in various contexts, including but not limited to, software bills
+of materials (SBOMs), vulnerability reporting, and cross-ecosystem compatibility.
+
+However, the [existing conda PURL definition][purl-conda-def] fails to properly capture the
+multi-stakeholder nature of the conda ecosystem, various commonly-understood concepts within the
+ecosystem, and critically, the standards specified in other accepted CEPs.  This CEP seeks to
+address such shortcomings with an ecosystem-approved specification for the conda PURL definition.
```
```suggestion
address such shortcomings with an ecosystem-approved specification for the conda PURL definition,
that will be submitted for review to the PURL standard.
```

(or similar wording)

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+</table>
+
+
+## Abstract
+
+This CEP describes the Package-URL type definition for conda packages in a way that conforms to
+other relevant standards, including other CEPs and ECMA-427.
+
+
+## Motivation
```
jaimergp, line 20: Can we add a practical example here? Something like:

> In other words, this CEP proposes a change so that conda PURLs currently written like `old-example-here` become `new-example-here` instead.

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+> A PURL is a URL composed of seven components:  
+>   scheme:type/namespace/name@version?qualifiers#subpath
+
+For conda PURLs, these seven components are defined as follows:
+
+#### `scheme` component
+
+In accordance with ECMA-427 Section 5.6.1 ("Scheme"), the `scheme` component for a conda PURL is
+the ASCII string literal "pkg".  Systems handling conda PURLs SHOULD also follow the other rules
+for the `scheme` component specified in ECMA-427 Section 5.6.1.
```
jaimergp: Isn't this essentially saying like "No changes in this section. Follow ECMA-427 as is"?

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+
+#### `type` component
+
+In accordance with ECMA-427 Section 5.6.2 ("Type"), the canonical form for a conda PURL's `type`
+component is the ASCII string literal "conda".  The `type` component MUST be treated as
+case-insensitive, and system handling conda PURLs are encouraged to normalize the `type` component
+to its canonical lowercase form, following the behaviors specified in ECMA-427 Section 5.5.
```
jaimergp: Same here. Isn't this essentially saying like "No changes in this section. Follow ECMA-427 as is".

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+  needed to properly identify the package(s), then the `subdir` qualifier (defined below) should be
+  used instead.
+
+An important implication of these rules is that a valid conda PURL must provide at least one of the
+`namespace` component or the `repository_url` qualifier, as a conda PURL that both omits the
+`namespace` component _and_ assumes the default value for the `repository_url` qualifier (defined
+below) specifies a channel base URL that is insufficient to properly locate the package.
+
+Further, the prohibition on using "defaults" as the `namespace` component value means that PURLs
+for Anaconda-built packages should use "main", "r", or "msys2" for their `namespace` component.
```
jaimergp: What about https://anaconda.org/main? That was always ambiguous to me.

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+comparisons.  In other words, the `version` component should be be treated as equivalent to a
+`==version` MatchSpec expression, as described in [CEP 29][matchspec-ver-match].  As an example, a
+`version` component with the value `1.2` should match versions `1.2` and `1.2.0` of the named
+package but not versions `1.2.0.1` and `1.2.3`.
+
+If [fuzzy equality][matchspec-ver-match] or other, more complex version matching is needed, the
+`version` component SHOULD be omitted and [the `vers` qualifier](#qualifiers-component) used
+instead.  The `version` component and `vers` qualifier are mutually exclusive and SHOULD NOT be
+used together in a conda PURL; in cases where both the `version` component and `vers` qualifier are
+provided, the `version` component SHALL take precedence.
```
jaimergp: IMO, the version component MUST be a version literal, end of the story. For anything else, `vers` needs to be used.

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+list consisting of:
+
+- The canonical URL of this CEP
+
+Note that ECMA-427 Section 6.12 specifies that the URLs listed in the `reference_urls` PURL type
+definition are informative.  However, to ensure standardized implementations, this CEP SHOULD be
+treated as normative and not just informative.
+
+
+## Examples
```
jaimergp, line 296: I think we should cover different use cases here:

- Fully identifying a single artifact in a given channel mirror (channel, subdir, name, version, build, repository url)
- Fully identifying a single artifact across several channel mirrors (channel, subdir, name, version, build)
- Identifying one or more artifacts for a given source code release (name, version, channel)
- Identifying one or more artifacts for a given project with a version range (name, vers, channel)
- Identifying all artifacts from a project (name, channel)
- Identifying all artifacts from a name across channels (name). Warn that this may be ambiguous for some names since not all projects are guaranteed to be published under the same name in different channels.

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+For conda PURLs, the ECMA-427 Section 6.5.1 requirement property for the `namespace` component is
+optional.  However, the use or non-use of the `namespace` component within a conda PURL MUST
+conform to the following rules:
+
+- If the `namespace` component is omitted from a conda PURL, then the `repository_url` qualifier
+  MUST be explicitly provided in the PURL, and its value MUST be a valid channel base URL, as
+  defined in CEP 26.
+- If the `namespace` component is provided, each ECMA-427 Section 5.6.3 `namespace` segment MUST be
+  a valid channel base URL path component, as defined in CEP 26.  All other rules for the
+  `namespace` component specified in ECMA-417 Section 5.6.3 apply to conda PURLs, including the
```
```suggestion
  `namespace` component specified in ECMA-427 Section 5.6.3 apply to conda PURLs, including the
```

Right?

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+[CEP 26](./cep-0026.md#version-strings) and [CEP 33](./cep-0033.md).
+
+The `version` component MUST comply with all character encoding rules specified in ECMA-427 Section
+5.4 ("Character Encoding").  In particular, if they are used, the epoch segment delimiter `!`
+(exclamation point) and the local version segment delimiter `+` (plus) must be percent encoded as
+`%21` and `%2B`, respectively.
+
+For conda PURLs to support the desired level of specificity, the `version` component SHOULD be
+treated as an exact equality, with any necessary `.0` padding, when used for package version
+comparisons.  In other words, the `version` component should be be treated as equivalent to a
```
```suggestion
comparisons.  In other words, the `version` component SHOULD be treated as equivalent to a
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+
+The following `qualifiers` are defined for conda PURLs:
+
+- `repository_url`: Optional qualifier, with a default value defined in the "`repository` property"
+  section of this CEP.  Its value, when combined with any `namespace` component as described in the
+  "`namespace` component" section, MUST specify the channel base URL of the identified package(s),
+  as defined in CEP 26.
+- `build`: Optional qualifier, with no default value.  If provided, its value MUST be the
+  [build string](./cep-0026.md#build-strings), as defined in CEP 26, of the identified package(s).
+- `build_number`: Optional qualifier, with no default value.  If provided, its value MUST the
```
```suggestion
- `build_number`: Optional qualifier, with no default value.  If provided, its value MUST be the
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+- `vers`: Optional qualifier, with no default value.  If provided, its value MUST be a valid
+  [Version Range Specifier (VERS)][vers-spec], whose `version-scheme` is `conda`; the specification
+  of the `conda` VERS and its relationship to concepts like the [MatchSpec query language][CEP29]
+  are considered out of scope for this CEP and will be discussed elsewhere.  This qualifier and the
+  `version` component are mutually exclusive and SHOULD NOT be used together; additional details
+  about this constraint are provided in the "`version` component" section of this CEP.
+- `checksum`:  Optional qualifier, with no default value.  If provided, its value MUST consist of
+  one or more checksum specifications, with consecutive checksum specifications separated by a
+  single, unencoded ASCII "," (comma) character.  Each checksum specification is defined as the
+  concatenating of the lowercase checksum algorithm name (e.g., "md5", "sha256"); a single,
```
```suggestion
  concatenation of the lowercase checksum algorithm name (e.g., "md5", "sha256"); a single,
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+conda PURL be mutually exclusive.  Further, it does specify which of these two qualifiers takes
+precedence in cases where their simultaneous use in a conda PURL creates a conflict.  Producers of
+conda PURLs should take care that their use of these qualifiers does not create a conflict, and
+consumers of conda PURLs are RECOMMENDED to treat conflicting use of these qualifiers as an invalid
+or erroneous PURL.
+
+This CEP explicitly _undefines_ the `channel` qualifier found in the [existing conda PURL
+definition][purl-conda-def].  The `channel` qualifier corresponds to a weakly-defined concept not
+properly captured in any existing CEP, and its presumed purpose in the existing definition is
+better handled by the combination of the `namspace` component and `repository_url` qualifier, as
```
```suggestion
better handled by the combination of the `namespace` component and `repository_url` qualifier, as
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+NOT be accepted when processing conda PURLs.  Should such values be needed to handle a conda PURL,
+they should be constructed from the other components and qualifiers defined in this CEP and the
+rules defined in CEP 26.
+
+#### `subpath` component
+
+The optional `subpath` component of a conda PURL is used to specify a path within the identified
+package(s), subject to the following:
+
+- The `subpath` component SHALL confirm to all rules specified in ECMA-427 Section 5.6.7
```
```suggestion
- The `subpath` component SHALL conform to all rules specified in ECMA-427 Section 5.6.7
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+  reference to the target path after the symbolic link has been partially or fully dereferenced.
+- When applied to [`tar.bz2`-/`v1`-format packages][CEP35.v1], "relative to the root of the
+  package" in ECMA-427 Section 5.6.7 MUST be interpreted as the path to an archive member within
+  the package tarball, or equivalently, as a relative path from a root directory into which the
+  package tarball was extracted.
+- When applied to [`conda`-/`v2`-format packages][CEP35.v2], "relative to the root of the package"
+  in ECMA-427 Section 5.6.7 MUST be interpreted as a relative path from a common root directory
+  into which both inner archives (i.e., `info-{name}-{version}-{build}.tar.zst` and
+  `pkg-{name}-{version}-{build}.tar.zst`) have been extracted.
+- The `subpath` component MUST be interpreted as a reference to a path as it exist in the package
```
```suggestion
- The `subpath` component MUST be interpreted as a reference to a path as it exists in the package
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
+# CEP XXXX - PURLs for conda packages
+
+<table>
+<tr><td> Title </td><td> Package-URLs (PURLs) for conda packages </td>
```
```suggestion
<tr><td> Title </td><td> Package-URLs (PURLs) for conda packages </td> </tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+
+The `version` component MUST comply with all character encoding rules specified in ECMA-427 Section
+5.4 ("Character Encoding").  In particular, if they are used, the epoch segment delimiter `!`
+(exclamation point) and the local version segment delimiter `+` (plus) must be percent encoded as
+`%21` and `%2B`, respectively.
+
+For conda PURLs to support the desired level of specificity, the `version` component SHOULD be
+treated as an exact equality, with any necessary `.0` padding, when used for package version
+comparisons.  In other words, the `version` component should be be treated as equivalent to a
+`==version` MatchSpec expression, as described in [CEP 29][matchspec-ver-match].  As an example, a
```
jaimergp, line 140:
```suggestion
`==version` MatchSpec expression, as described in [CEP 29][CEP29].  As an example, a
```

cep-XXXX.md
```diff
@@ -0,0 +1,407 @@
...
+  Updates to this CEP or additional CEPs may be needed if any gaps are identified in the `subpath`
+  component specifications.
+- This CEP provides the `repository_url` qualifier that can be used identify a package on a
+  _specific_ mirror of a channel.  However, this CEP does _not_ define a way to identify a specific
+  package artifact that appears across multiple mirrors; e.g., a single PURL that would
+  [simultaneously] identify a conda-forge package hosted on both `conda.anaconda.org/conda-forge`
+  and `prefix.dev/conda-forge`.  Supporting such PURLs would require additional work, including
+  formalizing the concept of "mirrors" for conda channels and specifying which PURL components or
+  qualifiers should be used to identify the set of mirrors for a given packages.  (Due to the
+  fairly portable nature of conda packages, the combination is [shortened] channel name, package
```
```suggestion
  fairly portable nature of conda packages, the combination of [shortened] channel name, package
```

jaimergp (commented):
Thanks, Cheng! I love to see such a thorough review of the PURL standards so we can submit edits to their team.

I added a few comments about typos and suggestions for the flow, but there's one point I want to emphasize: the `version` component must only be a literal. Anything else that may potentially identify more than one version needs to be in `vers`.

mjherzog:
@jaimergp @chenghlee The issue template for requesting changes to a registered PURL **type** is: 
https://github.com/package-url/purl-spec/blob/main/.github/ISSUE_TEMPLATE/Change%20registered%20PURL%20type.md
It will appear as a template option when you create a New Issue.

cep-XXXX.md
chenghlee: Currently, `anaconda.org/main` is a mirror of `repo.anaconda.com/pkgs/main`.  There's definitely work we at Anaconda need to do to resolve that ambiguity, but I think for the purposes of this CEP, it's okay to allow use of `main` as the conda PURL namespace component value.

cep-XXXX.md
```diff
@@ -0,0 +1,419 @@
...
+[CEP 26](./cep-0026.md#version-strings) and [CEP 33](./cep-0033.md).
+
+The `version` component MUST comply with all character encoding rules specified in ECMA-427 Section
+5.4 ("Character Encoding").  In particular, if they are used, the epoch segment delimiter `!`
+(exclamation point) and the local version segment delimiter `+` (plus) must be percent encoded as
+`%21` and `%2B`, respectively.
+
+For conda PURLs to support the desired level of specificity, the `version` component SHOULD be
+treated as an exact equality, with any necessary `.0` padding, when used for package version
+comparisons.  In other words, the `version` component SHOULD be treated as equivalent to a
```
beckermr: Why are we using SHOULD here? Given the next paragraph, I would make this a MUST since if we don't want exact equality, we have to use the `vers` qualifier and leave off the version.

```suggestion
For conda PURLs to support the desired level of specificity, the `version` component MUST be
treated as an exact equality, with any necessary `.0` padding, when used for package version
comparisons.  In other words, the `version` component MUST be treated as equivalent to a
```

beckermr (dismissed):
There are many instances of `must` in the text that I think should be `MUST`. We should clean that up for clarity.

baszalmstra:
@chenghlee Are you still working on this? There are a number of review comments that seem unaddressed for a while. No rush, just checking the status. :)

cep-XXXX.md
chenghlee: Changed to `MUST` here and in other places.

jaimergp:
@conda/steering-council, the RFC period has officially opened for this CEP. It will last two weeks, so please place your comments **before 2026-08-20 EOD AoE**. In there are no blocking concerns, a vote will start immediately after the RFC period.

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+<tr><td> Updated </td><td> 2026-04-08 </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+
+## Abstract
+
+This CEP describes the Package-URL type definition for conda packages in a way that conforms to
+other relevant standards, including other CEPs and ECMA-427.
```
jaimergp, line 17:
```suggestion
other relevant standards, including other CEPs and [ECMA-427][ECMA427].
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+<tr><td> Created </td><td> 2026-04-02 </td></tr>
+<tr><td> Updated </td><td> 2026-04-08 </td></tr>
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159 </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+</table>
+
+
+## Abstract
+
+This CEP describes the Package-URL type definition for conda packages in a way that conforms to
```
jaimergp, line 16:
```suggestion
This CEP describes the Package URL type definition for conda packages in a way that conforms to
```

Other instances in the text don't use a dash 🤔

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
+# CEP XXXX - PURLs for conda packages
+
+<table>
+<tr><td> Title </td><td> Package-URLs (PURLs) for conda packages </td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Cheng H. Lee &lt;clee@anaconda.com&gt; </td></tr>
+<tr><td> Created </td><td> 2026-04-02 </td></tr>
+<tr><td> Updated </td><td> 2026-04-08 </td></tr>
```
jaimergp, line 8:
```suggestion
<tr><td> Updated </td><td> 2026-08-10 </td></tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159 </td></tr>
```
jaimergp, line 9:
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/159, https://github.com/package-url/purl-spec/issues/868 </td></tr>
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+
+
+## Motivation
+
+Package URLs (PURLs) provide "a standardized, URL-based syntax that uniquely identifies software
+packages, independent of their ecosystem or distribution channel".  Users may want or need to use
+PURLs to identify conda packages in various contexts, including but not limited to, software bills
+of materials (SBOMs), vulnerability reporting, and cross-ecosystem compatibility.
+
+However, the [existing conda PURL definition][purl-conda-def] fails to properly capture the
```
jaimergp, line 27:
```suggestion
However, the [existing conda PURL definition][purl-conda-def] (as of v1.0.1) fails to properly capture the
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+[CEP33]: ./cep-0033.md
+[CEP34]: ./cep-0034.md
+[CEP35]: ./cep-0035.md
+[CEP35.v1]: ./cep-0035.md#tarbz2
+[CEP35.v2]: ./cep-0035.md#conda
+[CEP36]: ./cep-0036.md
+[gh-purl-spec]: https://github.com/package-url/purl-spec
+[matchspec-ver-match]: ./cep-0029.md#version-matching
+[purl-conda-def]: https://github.com/package-url/purl-spec/blob/a6c97bcfe5c83985a1da348f73a65c9842c7e354/types/conda-definition.json
+[purl-quals-guide]: https://github.com/package-url/purl-spec/blob/a6c97bcfe5c83985a1da348f73a65c9842c7e354/docs/common-qualifiers.md
```
jaimergp, line 414: Shall we use the v1.0.1 tag, cut last week?

```suggestion
[purl-conda-def]: https://github.com/package-url/purl-spec/blob/v1.0.1/types/conda-definition.json
[purl-quals-guide]: https://github.com/package-url/purl-spec/blob/v1.0.1/docs/specification/common-qualifiers.md
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+
+## Specification
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+"RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+described in BCP 14 ([RFC2119][RFC2119], [RFC8174][RFC8174]) when, and only when, they appear in
+all capitals, as shown here.
+
+For the purposes of this CEP, the phrase "existing conda PURL definition" refers to the [commit
+a6c97bc version of `types/conda-definition.json`][purl-conda-def], which is canonically available
```
jaimergp, line 50:
```suggestion
For the purposes of this CEP, the phrase "existing conda PURL definition" refers to the [v1.0.1 revision of `types/conda-definition.json`][purl-conda-def], which is canonically available
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+key word "shall" in ECMA-427 Clause 5.6.1 being interpreted as its BCP 14 equivalent "SHALL".
+
+#### `type` component
+
+The canonical form for conda PURL's `type` component is the ASCII string literal "conda".  Systems
+processing conda PURLs MUST follow all of the rules specified in ECMA-427 Clause 5.6.2 ("Type"),
+with the key word "shall" in ECMA-427 Clause 5.6.2 being interpreted as its BCP 14 equivalent
+"SHALL".  Systems processing conda PURLs are RECOMMENDED to normalize the `type` component to its
+canonical lowercase form, following the behaviors specified in ECMA-427 Clause 5.5 ("Case
+folding").
```
jaimergp, line 78: Do we really need to "repeat" the ECMA standard here?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+
+#### `namespace` component
+
+Roughly speaking, the `namespace` component of a conda PURL corresponds to the path and label name
+components of the [channel base URL defined in CEP 26](./cep-0026.md#channel-base-urls) and/or the
+concept most users have when using the `channel` configuration option supported by [most]
```
jaimergp, line 84: What is `[most]` linking to?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+Further, the prohibition on using "defaults" as the `namespace` component value means that PURLs
+for Anaconda-built packages must use "main", "r", or "msys2" for their `namespace` component.
+
+#### `name` component
+
+The `name` component MUST be the [distributable package name](./cep-0026.md#package-names) of the
+identified artifact(s), as defined in CEP 26.
+
+Conda PURLs MUST NOT be used to identify conda virtual packages since they do not correspond to
+any concrete [package] artifact and only exist on the client side.
```
jaimergp, line 125:
```suggestion
The `name` component MUST be the [package name](./cep-0026.md#package-names) of the
identified artifact(s), as defined in CEP 26.
```

PURLs do not necessarily need to encode concrete packages. I wouldn't exclude virtual packages from the spec, especially given the pending work at https://github.com/package-url/purl-spec/issues/222.

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+5.4 ("Character Encoding").  In particular, if they are used, the epoch segment delimiter `!`
+(exclamation point) and the local version segment delimiter `+` (plus) must be percent encoded as
+`%21` and `%2B`, respectively.
+
+For conda PURLs to support the desired level of specificity, the `version` component MUST be
+treated as an exact equality, with any necessary `.0` padding, when used for package version
+comparisons.  In other words, the `version` component MUST be treated as equivalent to a
+`==version` MatchSpec expression, as described in [CEP 29][matchspec-ver-match].  As an example, a
+`version` component with the value `1.2` matches versions `1.2` and `1.2.0` of the named
+package but not versions `1.2.0.1` and `1.2.3`.
```
jaimergp, line 142: Wouldn't it be simpler to say that it must be the version literal (i.e. the value in `index.json`? The version component is not meant to have operators of any kind, so only exact equality applies. For any other version ranges `vers` needs to be used.

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+
+If [fuzzy equality][matchspec-ver-match] or other, more complex version matching is needed, the
+`version` component MUST be omitted and [the `vers` qualifier](#qualifiers-component) used
+instead.  The `version` component and `vers` qualifier are mutually exclusive and MUST NOT be
```
jaimergp, line 146:
```suggestion
XXX. The `version` component and `vers` qualifier are mutually exclusive and MUST NOT be
```

Adjust as needed given the comment above.

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+  [build string](./cep-0026.md#build-strings), as defined in CEP 26, of the identified package(s).
+- `build_number`: Optional qualifier, with no default value.  If provided, its value MUST be the
+  [build number](./cep-0034.md#infoindexjson), as defined in CEP 34, of the identified package(s).
+- `subdir`: Optional qualifier, with no default value.  If provided, its value MUST be the
+  [channel subdir name](./cep-0026.md#subdir-names), as defined in CEP 26, of the identified
+  package(s).
+- `type`: Optional qualifier, with no default value.  If provided, its value MUST be exactly one of
+  [currently-recognized artifact extensions](./cep-0026.md#artifact-extensions) or its versioned
+  synonym, as defined in CEP 26.  At the time of this CEP's adoption, the accepted values are
+  `tar.bz2` (versioned synonym: `v1`), or `conda` (versioned synonym: `v2`).
```
jaimergp, line 168: No leading period, correct?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+- `vers`: Optional qualifier, with no default value.  If provided, its value MUST be a valid
+  [Version Range Specifier (VERS)][vers-spec], whose `version-scheme` is `conda`; the specification
+  of the `conda` VERS and its relationship to concepts like the [MatchSpec query language][CEP29]
+  are considered out of scope for this CEP and will be discussed elsewhere.  This qualifier and the
+  `version` component are mutually exclusive and MUST NOT be used together in a conda PURL; in
+  cases where both the `version` component and `vers` qualifier are provided, the corresponding
+  conda PURL MUST be treated as an invalid or erroneous URL.
```
jaimergp, line 175: This is repeated above, not sure if intentional.

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+- `checksum`:  Optional qualifier, with no default value.  If provided, its value MUST consist of
+  one or more checksum specifications, with consecutive checksum specifications separated by a
+  single, unencoded ASCII "," (comma) character.  Each checksum specification is defined as the
+  concatenation of the lowercase checksum algorithm name (e.g., "md5", "sha256"); a single,
```
jaimergp, line 179: Do we (or PURL authors) have a list of accepted algorithms?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+  unencoded ":" (colon) character; and the lowercase, hex-encoded checksum value.
```
jaimergp, line 180: How does the colon play with the PURL encoding rules? Don't we need to encode it?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+This CEP explicitly _undefines_ the `channel` qualifier found in the [existing conda PURL
+definition][purl-conda-def].  The `channel` qualifier corresponds to a weakly-defined concept not
+properly captured in any existing CEP, and its presumed purpose in the existing definition is
+better handled by the combination of the `namespace` component and `repository_url` qualifier, as
+described in this CEP.  Upon acceptance of this CEP, systems handling conda PURLs MUST treat
+PURLs containing a `channel` qualifier as invalid.
+
+To avoid the need for potentially complex precedence rules in this CEP, the [commonly-used
+qualifiers][purl-quals-guide] `filename` and `download_url` used by other PURL types are explicitly
+_not_ defined for conda PURLs.  In other words, the `filename` and `download_url` qualifiers MUST
```
jaimergp, line 221: Removing download_url is problematic if https://github.com/conda/ceps/pull/151 is ever approved. Shall we leave some space in the specification in case it makes it? And, regardless, we need to consider channels that define `base_url` -- what's their download_url?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+The ECMA-427 Clause 6.3 `description` property of the conda PURL type definition SHALL be
+"Conda packages".
+
+#### `repository` property
+
+The ECMA-427 Clause 6.4 `repository` property is REQUIRED for conda PURLs.  Its use and
+interpretation is defined in the "`namespace` component" and "`qualifier` component" sections.
+
+To maintain compatibility with the default behavior of existing conda-compatible clients, the
+default public repository (i.e., `default_repository_url`) for conda packages is SHALL be
```
jaimergp, line 276:
```suggestion
default public repository (i.e., `default_repository_url`) for conda packages SHALL be
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+- The canonical URL of this CEP
+
+Note that ECMA-427 Clause 6.12 specifies that the URLs listed in the `reference_urls` PURL type
+definition are informative.  However, to ensure standardized implementations, this CEP MUST be
+treated as normative and not just informative.
+
+
+## Examples
+
+- `pkg:conda/main/python`: basic usage, identifying [all] `python` packages from Anaconda
```
jaimergp, line 298:
```suggestion
- `pkg:conda/main/python`: basic usage, identifying _all_ `python` packages from Anaconda
```

Missing link or emphasis?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+- `pkg:conda/conda-forge/python@3.13.12?build=hc97d973_100_cp313&subdir=linux-64`: basic usage,
+  identifying a specific `python` build from conda-forge.
```
jaimergp, line 300:
```suggestion
  identifying exactly one specific `python` build from conda-forge.
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+## Backwards Compatibility
+
+This CEP intentionally breaks backwards-compatibility for PURLs based on the [existing conda PURL
+definition][purl-conda-def] in the following ways:
+
+- The `default_repository_url` has been changed from `https://repo.anaconda.com` to
+  `https://conda.anaconda.org`.
+- The requirement property of the `namespace` component has changed from "prohibited" to
+  "optional", with additional rules as described above.
+- `channel` is no longer an accepted qualifier for conda PURLs.
```
jaimergp, line 316: Shall we say "forbidden"? It's rejected now, isn't it?

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+
+These breaking changes are considered acceptable by the author(s) of this CEP, as the existing
+conda PURL definition fails to properly capture existing standards (i.e., CEPs) and commonly-used
+patterns in the conda ecosystem.  Among the motivations for these breaking changes:
+
+- Simple PURLs like `pkg:conda/python` and `pkg:conda/python?channel=conda-forge` are accepted
+  under the existing conda PURL definition but cannot be used to identify actual package artifacts,
+  as that definition's default values produces non-existent URLs.  The existing conda PURL
+  definition effectively requires that _every_ conda PURL explicitly provide a `repository_url`
```
jaimergp, line 325:
```suggestion
  definition effectively requires that _every_ conda PURL explicitly provides a `repository_url`
```

cep-XXXX.md
```diff
@@ -0,0 +1,420 @@
...
+  qualifier, possibly combined with a `channel` qualifier in an unspecified way, to produce a URL
+  that would correspond to an actual package artifact.
+- The prohibition of a `namespace` component in the existing conda PURL definition means
+  commonly-used patterns in the conda ecosystem (e.g., `channel::package=version`) cannot be
+  translated to "intuitive"-/similar-looking PURLs (e.g., `pkg:conda/channel/package@version`).
+  Coupled with the previously-noted choice of [existing] default values, this creates an unwanted
```
jaimergp, line 331:
```suggestion
  Coupled with the previously-noted choice of existing default values, this creates an unwanted
```

jaimergp (changes requested):
Excellent CEP, @chenghlee. I did a more thorough pass and came up with some suggestions. My two main concerns are:

- PURLs can be used to identify package requirements (e.g. as we are trying to do in PEP 725 / 804), so I'm not sure there's much value in forbidding virtual packages here.
- The version field should be a literal string, to be processed with exact equality only. All other use cases (trailing zeros, globs, ranges, etc) belong in `vers`.

I'm also curious: what happens in the PURL spec body refuses to accept some of these decisions? Do we withdraw the CEP and try again? Update it? What's the intended lifecycle once approved? Do we defer to future revisions of the PURL spec if there are conflict with this (historic) document?

On the editorial side, I've observed some instances of `[emphasis]`, double space after period, and double newline before headings. Those have not been common in previous CEPs, and I don't feel particularly strongly about it, but if you don't care either,  let's make it consistent: `_emphasis_`, single-space after period, single newline before headings.

cep-XXXX.md
chenghlee: Re: "forbidden", are you referring to `channel` qualifier? If so, the slightly clunky "no longer an accepted" language is there because ECMA-427 allows us to not define (or un-define) qualifier keys, but it does not allow us to explicitly prohibit (or forbid) a qualifier key.

cep-XXXX.md
chenghlee: It's a parenthetical that didn't need to be one. Fixed.

cep-XXXX.md
chenghlee: "repeat" as in, e.g., saying `scheme` must be the string literal `"pkg"`? Strictly speaking no, but I included these sections to make this CEP (relatively) self-contained; as in, someone could read this CEP, look at a `repodata.json` (or `index.json`) entry, and stand a reasonable chance of constructing a valid conda PURL without having to read through ECMA-427 or the PURL specification on GitHub.

I'm less concerned not repeating the other parts of the referenced ECMA-427 clauses since I think it's unlikely that anyone would run afoul of them unless they're purposely setting out to abuse the specification.

chenghlee: In any case, I removed the "RECOMMENDED" sentence because I agree that it does not really add to the goals of this CEP.

chenghlee:
> I'm also curious: what happens in the PURL spec body refuses to accept some of these decisions? Do we withdraw the CEP and try again? Update it? What's the intended lifecycle once approved? Do we defer to future revisions of the PURL spec if there are conflict with this (historic) document?

Honestly, I don't know. I guess the conda steering council could decide what should happen, depending on the extent of rejections and/or the specific changes made to PURL spec in the future. I suppose "minor" edits or clarifications could be made via PRs to this CEP, and anything "substantive" would require one or more new CEPs that clarify or supersede this one.

> On the editorial side, I've observed some instances of `[emphasis]`, double space after period, and double newline before headings. Those have not been common in previous CEPs, and I don't feel particularly strongly about it, but if you don't care either, let's make it consistent: _emphasis_, single-space after period, single newline before headings.

Double spaces and newlines just made it easier for me to read in my editor; I'm not particularly opinionated about the style, so I've modified the text to follow the styles used in other CEPs. (I was using `[...]` for parentheticals, not emphasis, which I guess in standard written English is done via `(...)`, and in any case, I've removed all but one of the parentheticals.)

cep-XXXX.md
chenghlee: Resolved by doing the opposite. 😄 

ECMA-427 uses "Package-URL" everywhere, so I made the one with the space use the dash instead.

cep-XXXX.md
chenghlee: Good point. Would you be okay with saying `download_url` is allowed, and that if provided:

* It MUST be the complete URL to the package, including the filename; and
* It MUST take precedence over all other ways of constructing a download URL for the package using other parts of the PURL?

I think I'm mostly okay with that, even if that last rule allows a PURL like `pkg:conda/conda-forge/python-3.15.0-habcdefg.conda?download_url=ftp%3A%2F%2Fexample.com%2Fmy-channel%2Fperl-5.1.0-h1234567.tar.bz2`, which I'm not sure what you're is supposed to do with. (Other than apply your own judgement as a processor and reject that PURL as "_technically_ valid under the CEP but completely nonsensical and unusable".)

cep-XXXX.md
mjherzog: We do not currently have a list of "registered" (our terminology for accepted/approved) checksum algorithms. We are working atm on adding more details about qualifiers keys for ECMA-427 (PURL) 2nd edition Clause 5.6.6 - final draft due to Ecma end of August.

cep-XXXX.md
mjherzog: In the **qualifiers** component the colon is not encoded.

chenghlee: ECMA-427 Clause 5.4 ("Character encoding") says we should not encode the `:`:

> The following characters shall not be percent-encoded:
> * the Alphanumeric Characters
> * the Punctuation Characters
> * the Separator Characters when being used as PURL separators
> * the colon ':', whether used as a Separator Character or otherwise
> * the percent sign '%' when used to represent a percent-encoded character

cep-XXXX.md
mjherzog: PURL does not have a concept of "invalid" *qualifier keys**. We are working atm on adding "registered" **keys** for a PURL type - see https://github.com/package-url/purl-spec/pull/945.
Also note that we have recently added a place to record important information for a PURL type that is not in the JSON definition for a PURL type - see also `docs/types/notes/purl-type-notes-template.md`. This should be a good place to summarize the major changes for the 'conda` PURL type.

cep-XXXX.md
chenghlee: As far as I know, the PURL standard does _not_ have a list of accepted algorithms.

In [a previous comment](https://github.com/conda/ceps/pull/159#discussion_r3048300335), I had considered simplifying this section by having the CEP explicitly list the accepted algorithms, but I'm increasingly of the opinion that we should leave the specification as is and _not_ limit the accepted algorithms. One main reason is that I've since seen documents tied to various existing or upcoming regulatory standards that roughly say, "MD5 for checksums is a hard no, SHA-256 is acceptable but not preferred, and we want you to use other algorithms from the SHA-2 or SHA-3 family." Given that, I'm now more in favor of leaving this ugly, overly complex specification, rather than provide an explicit list and have to scramble to update that list as various regulatory and/or standards bodies update their requirements/guidance.

cep-XXXX.md
chenghlee: I'll remove the _prohibition_ on using conda PURLs to reference virtual packages. Though after doing that, I'm inclined to state that specification of virtual packages using conda PURLs is currently undefined and deferred to a later CEP.

The problem I ran into with virtual packages was how to deal with the fact that they aren't attached to a specific channel, which is a bit hard to reconcile with what we aim to do with the `namespace` component and `repository_url` qualifier without increasing the complexity of the processing rules for them.

I suppose have this CEP say something like:

* The `namespace` component for virtual packages MUST be `*`, though that would have to be percent-encoded as `%2A` to comply with ECMA-427 Clause 5.2 and 5.4; and
* The `repository_url` qualifier MUST be ignored, or perhaps, the `repository_url` qualifier MUST be the percent-encoded form of the literal `*`. (I would prefer something like `repository_url=` [empty string], but ECMA-427 prohibits qualifier values from being empty or null.)

So the conda PURL for the `__unix` virtual package would be `pkg:conda/%2A/__unix` (with the `repository_url` being ignored) or `pkg:conda/%2A/__unix?repository_url=%2A`.

chenghlee: Now that I think about it, allowing `*` in the namespace component + rules on the corresponding `repository_url` qualifier could also be used as one approach for how PURLs can be used identify conda packages across channels and/or mirrors. Or at least, in the use cases where the goal is to say "all channels" or "all mirrors".

cep-XXXX.md
chenghlee: > ...only exact equality applies...

I guess this comes down to "whose definition of exact equality for versions do we want to use?"

I wrote this part of the specification from the standpoint of how we define "equality" in the context of conda MatchSpecs, where we really don't really have way of saying version "1.2" should match "1.2" but _not_ "1.2.0" nor "1.2.0.0.0". So I'm worried that if we were to prohibit use of ".0" padding when handling the PURL `version` component, we would have to introduce a lot of complexity in any tooling that converts PURLs to MatchSpecs. E.g., how would we convert `purl:conda/conda-forge/python@3.15` to a MatchSpec? Make it `conda-forge::python ==3.15,!=3.15.0`, which I'm not even sure would work? Define a third equality-like operator to MatchSpec (`===`)? Etc.

cep-XXXX.md
chenghlee: Correct, no leading periods according to CEP 26:

> The conda ecosystem currently recognizes two artifact extensions: `tar.bz2` and `conda`...

cep-XXXX.md
chenghlee: It's intentional, so reading this part of the CEP doesn't require you to go back and look at the other part to be sure of the behavior. (And vice versa.)

cep-XXXX.md
chenghlee: I think "define exact equality for versions" is less of an issue in other package ecosystems because they don't make a distinction between `=3.15`, `==3.15`, and (arguable) `~=3.15` in the way conda & friends do.

cep-XXXX.md
```diff
@@ -0,0 +1,409 @@
...
+[CEP34]: ./cep-0034.md
+[CEP35]: ./cep-0035.md
+[CEP35.v1]: ./cep-0035.md#tarbz2
+[CEP35.v2]: ./cep-0035.md#conda
+[CEP36]: ./cep-0036.md
+[gh-purl-spec]: https://github.com/package-url/purl-spec
+[matchspec-ver-match]: ./cep-0029.md#version-matching
+[purl-conda-def]: https://github.com/package-url/purl-spec/blob/a6c97bcfe5c83985a1da348f73a65c9842c7e354/types/conda-definition.json
+[purl-quals-guide]: https://github.com/package-url/purl-spec/blob/a6c97bcfe5c83985a1da348f73a65c9842c7e354/docs/common-qualifiers.md
+[vers-spec]: https://github.com/package-url/vers-spec/blob/main/docs/standard/specification.md
```
Hind-M:
```suggestion
[vers-spec]: https://github.com/package-url/vers-spec/blob/main/docs/specification/standard/specification.md
```

cep-XXXX.md
```diff
@@ -0,0 +1,409 @@
...
+- CEP 26 allows `file://` URLs to be channel base URLs, and the diversity of file systems and their
+  behaviors may lead to conflicts with the requirements in this CEP and/or ECMA-427. To avoid
+  making this CEP overly complicated, the author(s) have intentionally chosen to defer any work
+  needed to properly handle certain types of `file://`-based channels, especially since it is
+  unclear if PURLs are meaningfully useful for such channels.
+- The specifications for the `subpath` component may not be fully complete with respect to every
+  type of path (i.e., file or directory) that could possibly be included in conda packages.
+  Updates to this CEP or additional CEPs may be needed if any gaps are identified in the `subpath`
+  component specifications.
+- This CEP provides the `repository_url` qualifier that can be used identify a package on a
```
Hind-M:
```suggestion
- This CEP provides the `repository_url` qualifier that can be used to identify a package on a
```

cep-XXXX.md
```diff
@@ -0,0 +1,409 @@
...
+  _specific_ mirror of a channel. However, this CEP does _not_ define a way to identify a specific
+  package artifact that appears across multiple mirrors; e.g., a single PURL that would
+  simultaneously identify a conda-forge package hosted on both `conda.anaconda.org/conda-forge`
+  and `prefix.dev/conda-forge`. Supporting such PURLs would require additional work, including
+  formalizing the concept of "mirrors" for conda channels and specifying which PURL components or
+  qualifiers should be used to identify the set of mirrors for a given packages. (Due to the
```
Hind-M:
```suggestion
  qualifiers should be used to identify the set of mirrors for a given package. (Due to the
```

cep-XXXX.md
```diff
@@ -0,0 +1,409 @@
...
+
+## Examples
+
+- `pkg:conda/main/python`: basic usage, identifying _all_ `python` packages from Anaconda
+- `pkg:conda/conda-forge/python@3.13.12?build=hc97d973_100_cp313&subdir=linux-64`: basic usage,
+  identifying exactly one specific `python` build from conda-forge.
+- `pkg:conda/python?repository_url=https%3A%2F%2Frepo.anaconda.com%2Fpkgs%2Fmain`: using the
+  `repository_url` qualifier to fully identify the channel base URL
+- `pkg:conda/conda-forge/python?repository_url=https%3A%2F%2Fprefix.dev`: using the
+  `repository_url` qualifier to identify packages hosted by a conda-forge mirror.
```
Hind-M: Maybe add invalid/conflicting PURLs examples?

- An invalid PURL containing both `version` component and `vers` qualifier
- A PURL illustrating a conflict between the `build` and `build number` qualifiers
- An invalid PURL using the undefined `channel` qualifier

jaimergp:
@conda/steering-council, gentle reminder <3

msarahan:
PURLs are great. This is a very good direction. I have been using PURLs for tasks such as detecting false positives from CVE scans. Most CVE records are not including PURL relationships, but I've been manually augmenting our database with them. I've been associating them primarily by source code location. For example, I have https://openhub.net/vulnerabilities/bdsa/BDSA-2026-16119 associated with the purl "pkg:github/python/cpython", and then on the other side, I have the conda package name "python" or "cpython" mapped to the same purl. A mismatch of these two sides helps me exclude false positives.

Conda packages are part of the story, but they're also a container for other things, and the identity of the stuff inside the container is often just as important as the container (and its implied build details) are.

Does that fit somewhere in this CEP, or is it a separate (perhaps not even a Conda-related) issue?

jaimergp:
> Does that fit somewhere in this CEP, or is it a separate (perhaps not even a Conda-related) issue?

Hey @msarahan, you'll be delighted to check https://github.com/conda/ceps/pull/63 and https://github.com/conda/ceps/pull/114 (these should be merged into one)! Also: https://github.com/conda-incubator/sig-purls/issues

cep-XXXX.md
jaimergp: But namespaces can be empty, right? And in that case they result in an underspecified PURL, but a valid "requirement" as in "any package named like this, no matter where it comes from"

jaimergp: If we allow that, we don't need `*` or the discussion of how that's allowed to mean "any channel" (but channel globstrings are disallowed).

cep-XXXX.md
jaimergp: I'd argue that a version field cannot be transformed into a matchspec; only a `vers` qualifier can. For PyPI, they don't really go into that detail https://packageurl.org/docs/purl-spec/types/definitions/pypi-definition#version-definition. But yea, it's not very practical 🤔 Fine with leaving the `.0` padding as a possible interpretation / normalization.

Fun fact: you can use as many equal signs as you want 😂 

image

cep-XXXX.md
jaimergp: That example PURL would need to be interpreted as "the `python-3.15.0-habcdefg.conda` record found in the `conda-forge` channel (subdir missing), with download URL available at given location; validity only checkable if the repodata is consulted for the `url` field"

Which makes me think that we can also add some language in the `url` field CEP to prevent this nonsense (e.g. package name must match or something like that)

cep-XXXX.md
chenghlee: > But namespaces can be empty, right?

As this CEP is currently written, yes, but only if the `repository_url` is supplied and is a valid channel base URL:

> - If the `namespace` component is omitted from a conda PURL, then the `repository_url` qualifier MUST be explicitly provided in the PURL, and its value MUST be a valid channel base URL, as defined in CEP 26.

I guess where I am getting stuck trying to figure out how to write this part of the specification comes from these seemingly conflicting requirements:

* We want "all" combinations of the `repository_url` qualifier and the `namespace` components to form a valid (CEP 26) channel base URL. I think this is especially important because unlike many other package ecosystems, the conda ecosystem doesn't really have a centralized artifact registry, so a means of identifying/locating the registry (i.e., channel) being used is critical.

* We want to `repository_url` qualifier to have a "sensible" default value so conda PURLs are less verbose and look like something the ecosystem is already familiar with (i.e., the `-c ...` argument and `channel::pkg` syntax)

* ECMA-427 explicitly prohibits empty/null namespace segments and empty/null qualifier values.

* We want conda PURLs to be able to support identifying packages that are not attached to a channel (e.g., virtual packages).

* We want conda PURLs to be able to support identifying packages without locating them to a specific channel.

If those ECMA-427 prohibitions did not exist, I would propose something like `pkg:conda//__unix` (empty first namespace segment) or `pkg:conda/__unix?repository_url=` (explicitly empty `repository_url` qualifier) to support the last two "any channel" cases, with a preference for the latter syntax.

I personally don't like the idea of allowing the `namespace` component to be empty _and_ the `repository_url` to be unspecified (and therefore implicitly adopting its default value) as a means to specify "any channel". That's (roughly) what the current PURL specification allows, and in my opinion, makes it really hard to distinguish between a mistakenly underspecified PURL and an intentional attempt to identify a conda package regardless of channel.

cep-XXXX.md
chenghlee: Added the `download_url` qualifier.

cep-XXXX.md
chenghlee: @jaimergp: I think this is the only remaining part of the CEP that has major sticking points, but I doubt we'll be able to resolve these issues quickly, especially since it may involve larger conversations with "outside" entities like the PURL spec body.

I do think the rest of this CEP is in good shape and is enough of specification to cover a broad range up to majority of use cases. In the interests of avoiding perfect being the enemy of good, what do you think about the idea of simply removing the prohibition on virtual packages, turning how to handle them and other cases of N≠1 channels into deferred work, and then calling a vote on the this CEP?

jaimergp: > [...] allowing the namespace component to be empty and the repository_url to be unspecified (and therefore implicitly adopting its default value) as a means to specify "any channel" [...] makes it really hard to distinguish between a mistakenly underspecified PURL and an intentional attempt to identify a conda package regardless of channel.

You have good points here; the implied default would be an issue. What if we only let the default for repository_url kick in when the namespace is non-empty? Otherwise it's null.

> I do think the rest of this CEP is in good shape and is enough of specification to cover a broad range up to majority of use cases. In the interests of avoiding perfect being the enemy of good, what do you think about the idea of simply removing the prohibition on virtual packages, turning how to handle them and other cases of N≠1 channels into deferred work, and then calling a vote on the this CEP?

I'm fine leaving this aspect in a "Further work" section or similar. This will also allow us to know the PURL body better and see what can fly for that trickier follow up work.

cep-XXXX.md
```diff
@@ -313,12 +334,13 @@ These breaking changes are considered acceptable by the author(s) of this CEP, a
...
-- Simple PURLs like `pkg:conda/python` and `pkg:conda/python?channel=conda-forge` are accepted
-  under the existing conda PURL definition but cannot be used to identify actual package artifacts,
-  as that definition's default values produces non-existent URLs. The existing conda PURL
-  definition effectively requires that _every_ conda PURL explicitly provides a `repository_url`
-  qualifier, possibly combined with a `channel` qualifier in an unspecified way, to produce a URL
-  that would correspond to an actual package artifact.
+- PURLs like `pkg:conda/python` and `pkg:conda/python?channel=conda-forge` are accepted under the
+  existing conda PURL definition but cannot be used to identify actual package artifacts, as that
+  definition's default values produces non-existent channel base URLs (`https://repo.anaconda.com/`
+  and `https://repo.anaconda.com/conda-forge`, respectively).  The existing conda PURL definition
```
```suggestion
  and `https://repo.anaconda.com/conda-forge`, respectively).  The existing (v1.0.1) conda PURL definition
```

cep-XXXX.md
chenghlee: I added these examples as in-line with the relevant sections:

* A PURL illustrating a conflict between the build and build number qualifiers
* An invalid PURL using the undefined channel qualifier

I left out "An invalid PURL containing both `version` component and `vers` qualifier" in this revision because we don't have a great definition for what a conda VERS looks like; that work has been deferred to the next round of CEPs.

jaimergp (approved):

jaimergp:
Dear @conda/steering-council, 

The vote for this CEP has started. It will be open for two weeks, until **September 16th, 2026, 23:59 Anywhere on Earth**. This time period has been chosen to make it eligible for time-out rules. As an Enhancement Proposal vote, it requires 60% affirmative votes to pass. 

- Uwe Korn (@xhochy) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Cheng H. Lee (@chenghlee) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Filipe Fernandes (@ocefpaf) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Michael Sarahan (@msarahan) 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- John Kirkham (@jakirkham) 
    - [ ] yes 
    - [ ] no 
    - [ ] abstain 
- Jannis Leidel (@jezdez) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Wolf Vollprecht (@wolfv) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Bas Zalmstra (@baszalmstra) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Hind Montassif (@Hind-M) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain 
- Pavel Zwerschke (@pavelzw) 
    - [x] yes 
    - [ ] no 
    - [ ] abstain

jaimergp:
@conda/steering-council mid period reminder; cc @mariusvniekerk, @marcelotrevisani, @msarahan, @jakirkham

jaimergp:

* SC size: 13
* Votes: 11 (84.6% of SC)
  * Yes: 11 (100% of votes)
  * No: 0
  * Abstain: 0
  
@conda/steering-council, this will be minted as CEP 49.

PR #168, opened 2026-05-20
CEP XXXX: add recipe source attestation validation

wolfv:
Adding a CEP to validate the _source_ section against know sigstore providers.

## Checklist for submitter

- [ ] I am submitting a new CEP: [Put your title here](#link-to-markdown-preview-in-branch).
  - [ ] I am using the CEP template by creating a copy `cep-0000.md` named `cep-XXXX.md` in the root level.

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -85,8 +85,15 @@ The `attestation` mapping has the following fields:
 The `attestation` field MUST NOT appear on `git:` or `path:` sources in
 this CEP. (Future CEPs MAY extend it to other source types.)
 
+When `attestation` is present, the source MUST also declare a `sha256`
+hash: the in-toto subject formats used in practice identify artifacts by
+SHA-256, and a weaker hash such as `md5` next to an attestation would
+undermine the binding.
+
 A recipe MAY omit the `attestation` field; absence means the builder
 SHALL NOT perform any source attestation verification for that source.
+The field is an additive, backwards-compatible extension of the
+[CEP 14] recipe schema and does not change the schema version.
 
 ### Publisher Identities
 
@@ -115,9 +122,12 @@ publishers:
 
 An `identity` value that is a URL (starts with `https://`) is matched
 against URI SANs using **repository-boundary prefix matching** (see
-below). Any other `identity` value is matched for exact string equality
-against the certificate's SAN (e.g. an email SAN). The `issuer` is
-always compared for exact string equality with the certificate's issuer
+below); this comparison is case-insensitive, since the providers defined
+here treat owner and repository names as case-insensitive while the SAN
+carries whatever casing the provider canonicalized. Any other `identity`
+value is matched for exact, case-sensitive string equality against the
+certificate's SAN (e.g. an email SAN). The `issuer` is always compared
+for exact, case-sensitive string equality with the certificate's issuer
 extension.
 
 The mapping form can express identities that are not `owner/repo` URLs —
@@ -141,6 +151,17 @@ A shorthand publisher identity is a string of the form:
 | `owner`    | The owner / organization on that provider. MUST NOT be empty.                          |
 | `repo`     | The repository on that provider. MUST NOT be empty.                                    |
 | `ref`      | Optional ref constraint (e.g. `refs/tags/v1.0`). Reserved; see **Future Work** below.  |
+
+A shorthand is parsed by splitting at the first `:` (provider), the
+first subsequent `/` (owner), and the first `@` (ref), if any. The
+`<repo>` component MAY therefore itself contain `/` separators to
+address nested namespaces — e.g. GitLab subgroups:
+`gitlab:group/subgroup/project` expands to the identity
+`https://gitlab.com/group/subgroup/project`.
+
+The `ref` semantics are not defined by this CEP. Builders that do not
+implement ref matching MUST fail with an error when a `ref` is present,
+rather than silently ignoring the constraint.
 
 Each shorthand is exactly equivalent to an explicit mapping:
 
@@ -182,11 +203,15 @@ This rule MUST be applied by all conformant builders.
 
 ### Bundle Discovery
 
+[CEP 14] allows `url:` to be a list of equivalent mirror URLs; the
+*download URL* below refers to the entry from which the builder
+actually retrieved the archive.
+
 The bundle URLs are determined as follows, in order:
 
 1. If `bundle_url` is set in the recipe, the builder MUST fetch every
    listed URL (a plain string is equivalent to a single-element list).
-2. Otherwise, if the source URL host is `pypi.org`, `pypi.io`, or
+2. Otherwise, if the download URL's host is `pypi.org`, `pypi.io`, or
    `files.pythonhosted.org`, the builder MUST construct a PyPI
    Integrity API URL of the form:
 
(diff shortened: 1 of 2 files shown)


cep-xxxx-source-attestations.md
```diff
@@ -0,0 +1,299 @@
+# CEP XXXX - Source Attestation Verification in Recipes
+
+<table>
+<tr><td> Title </td><td> Source Attestation Verification in Recipes </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> May 20, 2026</td></tr>
+<tr><td> Updated </td><td> May 20, 2026</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
```
jaimergp:
```suggestion
<tr><td> Discussion </td><td> https://github.com/conda/ceps/pull/168 </td></tr>
```

cep-xxxx-source-attestations.md
```diff
@@ -0,0 +1,299 @@
...
+<tr><td> Implementation </td><td> https://github.com/prefix-dev/rattler-build (experimental) </td></tr>
+<tr><td> Requires </td><td> CEP 14 (recipe format) </td></tr>
```
jaimergp:
```suggestion
```

cep-xxxx-source-attestations.md
```diff
@@ -0,0 +1,299 @@
...
+```text
+<provider>:<owner>/<repo>[@<ref>]
+```
+
+| Component  | Meaning                                                                                |
+| ---------- | -------------------------------------------------------------------------------------- |
+| `provider` | The identity provider. This CEP defines `github` and `gitlab`. Others MAY be added.    |
+| `owner`    | The owner / organization on that provider. MUST NOT be empty.                          |
+| `repo`     | The repository on that provider. MUST NOT be empty.                                    |
+| `ref`      | Optional ref constraint (e.g. `refs/tags/v1.0`). Reserved; see **Future Work** below.  |
```
jaimergp: What about using some subset of PURLs here?

cep-XXXX.md
```diff
@@ -0,0 +1,299 @@
...
+
+A recipe MAY omit the `attestation` field; absence means the builder
+SHALL NOT perform any source attestation verification for that source.
+
+### Publisher Identity Grammar
+
+A publisher identity is a string of the form:
+
+```text
+<provider>:<owner>/<repo>[@<ref>]
```
facutuesca, line 151: I think this is too restrictive. For example, on PyPI, packages uploaded with attestations signed by Google Cloud have subjects that are emails, and not `owner/repo`:
```
  Subject Alternative Name (critical):
    email:
    - 919436158236-compute@developer.gserviceaccount.com
```
[src](https://search.sigstore.dev/?logIndex=200170367)

cep-XXXX.md
```diff
@@ -0,0 +1,299 @@
...
+| `ref`      | Optional ref constraint (e.g. `refs/tags/v1.0`). Reserved; see **Future Work** below.  |
+
+For the providers defined here, the publisher identity maps to the
+Sigstore certificate's Subject Alternative Name (SAN) and OIDC issuer
+as follows:
+
+| Provider | Identity prefix                          | OIDC issuer                                          |
+| -------- | ---------------------------------------- | ---------------------------------------------------- |
+| `github` | `https://github.com/<owner>/<repo>`      | `https://token.actions.githubusercontent.com`        |
+| `gitlab` | `https://gitlab.com/<owner>/<repo>`      | `https://gitlab.com`                                 |
```
facutuesca, line 177: With the upcoming PyPI support for self-hosted GitLab and GitHub instances, there will be attestations that are technically from `github` and `gitlab`, but will have different OIDC issuers. Something to keep in mind here, with regards to naming the providers

cep-xxxx-source-attestations.md
```diff
@@ -0,0 +1,299 @@
...
+
+Builders that encounter an unknown `provider` MUST fail with an error
+rather than silently skipping verification.
+
+### Bundle Discovery
+
+The bundle URL is determined as follows, in order:
+
+1. If `bundle_url` is set in the recipe, the builder MUST use that URL.
+2. Otherwise, if the source URL host is `pypi.io` or
```
facutuesca: why `.io` instead of `.org`?

cep-XXXX.md
```diff
@@ -0,0 +1,299 @@
...
+
+   ```text
+   https://pypi.org/integrity/<project>/<version>/<filename>/provenance
+   ```
+
+   where `<project>` is the [PEP 503][PEP 503]-normalized project name
+   (lowercase, `[-_.]` collapsed to `-`) and `<version>` and `<filename>`
+   are extracted from the source URL's filename.
+3. Otherwise, the builder MUST fail with an error reporting that no
+   `bundle_url` is set and one cannot be auto-derived.
```
facutuesca, line 243: > Otherwise, the builders MUST fail 

this is in contradiction with the statement below:

> Builders MAY recognize additional auto-derived hosts

If builders may recognize additional auto-derived hosts, they will not fail for those hosts, even if the bundle_url is not set in the recipe (1) and the source URL host is not pypi (2).

facutuesca: maybe it makes sense to say "builders SHOULD fail" instead?

cep-XXXX.md
```diff
@@ -0,0 +1,299 @@
...
+- `S == P`, or
+- `S` starts with `P` *and* the character immediately following `P` in
+  `S` is `/` or `@`.
+
+This rule MUST be applied by all conformant builders.
+
+### Predicate Semantics Are Out of Scope
+
+This CEP does not define, restrict, or interpret the predicate types
+that appear in the in-toto statement. In particular:
```
facutuesca, line 314: why not lock the predicate type too? The builders don't need to "interpret it", but they could check if it matches the type specified in the lockfile.

My thinking here is that in this CEP (as it's currently written), the verification process checks that a given artifact has a valid signature from an identity specified in the lockfile. But since these signatures have semantics attached (specified by their predicate type), accepting _any_ signature from that identity as valid might be too much.

So, locking the predicate type, we can ensure during veriication that the semantics of the signature have not changed since they were locked, without actually interpreting them.

cep-XXXX.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Examples
+
+### PyPI source with auto-derived bundle URL
+
+```yaml
+source:
+  url: https://files.pythonhosted.org/packages/ab/cd/flask-3.1.1.tar.gz
+  sha256: "6489f1..."
+  attestation:
```
facutuesca, line 408: currently, this CEP only allows for one attestation url/sounce per source. One thing to consider is if in the future we might want to have multiple attestations sources (such as PyPI _and_ GitHub) per source

cep-XXXX.md
wolfv: Do you think something like `provider:<identifier>` would work where the tool, or the CEP, should define a list of supported provider & identifier pairs? Or should we turn this into a true dictionary supporting all OID fields and match on those? Do you know of any prior art for this?

facutuesca: I wonder if we should support this shorthand form to begin with. It adds complexity by having multiple ways of representing the same identity, and by having different rules on how to parse the shorthand versions (i.e `gihub`'s `<owner>/<repo>[@<ref>]` vs whatever we come up with the Google Cloud identity).

If we decide to go with it, we should make the matching depend on the issuer. For example, the paragraph that says:
> An `identity` value that is a URL (starts with `https://`) is matched
against URI SANs using **repository-boundary prefix matching**

we shouldn't do this in general, since there's no guarantee that a subject that contains a URL is referring to a repository. Rather, we can specify it for known issuers where we know it to be the case (github, gitlab), but otherwise we should do exact matching.

facutuesca: (the above also applies to the explicit mapping form)

PR #170, opened 2026-05-29
CEP 44: fix invalid character range in group name regex

Hofer-Julian:
Inside a regex character class (the square brackets [...]), the hyphen - is a special character used to define a range (like a-z or 0-9).

By placing it at the very end, it is interpreted as character instead

### DIFF
--- a/cep-0044.md
+++ b/cep-0044.md
@@ -36,7 +36,7 @@ The chosen keyword in CEP 29 `MatchSpec`, keywords is `extras` given its popular
 
 ## Specification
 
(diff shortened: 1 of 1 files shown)


wolfv:
Works for me - I guess I'd have a slight preference to escape the `\-` to avoid this ambiguity. But both is fine

wolfv (approved):

wolfv:
Can you add a little note to the changelog at the bottom?

Hofer-Julian:
> Can you add a little note to the changelog at the bottom?

Done!

chenghlee (approved):

PR #179, opened 2026-07-09
CEP XXXX: Prefix placeholder offsets in `paths.json`

chrisburr:
Opening this CEP as requested in https://github.com/conda/rattler/pull/2565. AI was used to help write it but I've reviewed it extremely carefully.

## Checklist for submitter

- [x] I am submitting a new CEP: [Prefix placeholder offsets in `paths.json`](https://github.com/chrisburr/ceps/blob/paths-offsets/cep-XXXX.md).

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,412 @@
+# CEP XXXX - Prefix placeholder offsets in `paths.json`
+
+<table>
+<tr><td> Title </td><td> Prefix placeholder offsets in <code>paths.json</code> </td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Chris Burr &lt;christopher.burr@cern.ch&gt; </td></tr>
+<tr><td> Created </td><td> Jul 9, 2026 </td></tr>
+<tr><td> Updated </td><td> Jul 9, 2026 </td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> https://github.com/conda/rattler/pull/2565 </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+> "SHOULD NOT", "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this
+> document are to be interpreted as described in [RFC2119][RFC2119] when, and
+> only when, they appear in all capitals, as shown here.
+
+[RFC2119]: https://datatracker.ietf.org/doc/html/rfc2119
+
+## Abstract
+
+This CEP adds two optional fields to the path entries of `./info/paths.json`
+(see [CEP 34](./cep-0034.md)): `offsets`, which records the byte positions at
+which the prefix placeholder can be replaced by plain substitution, and
+`shebang_length`, which marks a leading shebang line that installers transform
+under their own rules. Together they allow installers to perform prefix
+replacement without searching file contents, and to compute the exact size of
+a file after replacement without rendering it.
+
+## Motivation
+
+### Linking performance
+
+For every path entry with a `prefix_placeholder`, installers must read the
+complete file contents to search for occurrences of the placeholder before
+writing the patched copy into the target environment. The build tool already
+located every occurrence when it created the package: `conda-build` and
+`rattler-build` scan each file for the build prefix in order to decide whether
+to record a `prefix_placeholder` at all. That information is discarded when
+the package is written, and every installation recomputes it.
+
+With the occurrences recorded in `paths.json`, replacement becomes a plain
+copy with substitutions spliced in at known positions. For binary files, where
+replacement preserves the file length, an installer can go further and patch
+only the affected byte ranges of a copy-on-write clone of the cached file
+(e.g. created with reflink) instead of streaming a full copy.
+
+### Virtual filesystems
+
+An on-demand virtual filesystem that materializes conda environments lazily
+(such as the RattlerVFS work in [conda/rattler#2032][rattler-2032]) needs to
+answer `stat()` calls, i.e. report the final size of a file, before the file
+contents are ever read, and needs to serve reads of arbitrary ranges of the
+patched file directly from the original package data.
+
+Both needs reduce to knowing, without reading the file, where the placeholder
+occurs and how the file length changes. For binary files replacement preserves
+the length. For text files the length changes by a fixed amount per
+occurrence, except for the shebang line, which installers rewrite under
+special rules. `offsets` and `shebang_length` provide exactly the missing
+information.
+
+[rattler-2032]: https://github.com/conda/rattler/pull/2032
+
+## Specification
+
+This CEP amends the _path entry_ schema of `./info/paths.json` defined in
+[CEP 34](./cep-0034.md). Two optional keys are added to each path entry:
+
+- `offsets: list[int] | list[list[int]]`. Optional. The byte offsets at which
+  `prefix_placeholder` occurs in the file contents as stored in the package
+  (i.e. before any replacement). It MUST NOT be present unless
+  `prefix_placeholder` is present. Its shape depends on `file_mode`, as
+  described below.
+- `shebang_length: int`. Optional. The length in bytes of the file's first
+  line, including the terminating newline. Its presence is REQUIRED under the
+  conditions described below.
+
+Throughout this CEP, an _occurrence_ is a leftmost, non-overlapping match of
+the UTF-8 encoding of `prefix_placeholder` in the file contents, i.e. the
+matches found by a single left-to-right search: the same semantics as the
+search-based replacement in existing installers.
+
+`file_mode` normatively determines which of the two shapes of `offsets`
+applies. Implementations MUST NOT infer the shape from the JSON structure,
+since an empty list is valid for the text form and indistinguishable from an
+empty binary form.
+
+`paths_version` remains `1`.
+
+### Text offsets
+
+When `file_mode` is `text`, `offsets` MUST be a flat list of integers, e.g.
+`[10, 45]`. Each value is the offset, in bytes from the beginning of the file,
+at which one occurrence of the placeholder begins.
+
+The list MUST contain every occurrence in the file except those inside the
+shebang region (see `shebang_length` below). It MUST be sorted in strictly
+increasing order and occurrences MUST NOT overlap: each value MUST be greater
+than or equal to the previous value plus the byte length of the placeholder.
+It MAY be empty only when `shebang_length` is present and every occurrence
+falls within the shebang region.
+
+### Binary offsets
+
+When `file_mode` is `binary`, `offsets` MUST be a list of lists of integers,
+e.g. `[[64, 96], [200, 240, 300]]`. Replacement in binary files operates on
+NUL-terminated strings (C strings), because the padding that keeps the file
+length unchanged must stay within the string that contains the occurrence.
+Each inner list therefore describes one C string that contains at least one
+occurrence of the placeholder:
+
+- All values except the last are the offsets, in bytes from the beginning of
+  the file, at which an occurrence of the placeholder begins. They MUST be
+  sorted in strictly increasing order and MUST NOT overlap, as for text
+  offsets.
+- The last value is the offset of the NUL byte (`0x00`) that terminates the
+  C string, or the file size when no NUL byte follows the final occurrence.
+  Every occurrence offset `o` in the same inner list MUST satisfy
+  `o + len(prefix_placeholder) <= nul_offset`.
+
+The list MUST NOT be empty and each inner list MUST contain at least two
+values. The inner lists MUST be sorted by their first value, the byte ranges
+they describe MUST NOT overlap, and together they MUST cover every occurrence
+of the placeholder in the file.
+
+### `shebang_length`
+
+`shebang_length` MUST be present if and only if all of the following hold:
+
+- `offsets` is present,
+- `file_mode` is `text`, and
+- the file starts with the bytes `#!`.
+
+Note that its presence does not depend on whether the first line contains the
+placeholder (see Rationale).
+
+Its value MUST be `i + 1`, where `i` is the offset of the first newline byte
+(`0x0A`) in the file, or the size of the whole file when it contains no
+newline. A carriage return (`0x0D`) preceding the newline lies inside the
+region.
+
+The first `shebang_length` bytes of the file form the _shebang region_.
+Occurrences inside the region MUST NOT be listed in `offsets`, so every value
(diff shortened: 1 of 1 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,412 @@
...
+Build tools MAY include `offsets` and `shebang_length` when writing
+`./info/paths.json`. When present, the fields MUST be exhaustive and
+consistent with the file contents as described above: every occurrence of the
+placeholder MUST either be listed in `offsets` or fall within the shebang
+region, such that an installer relying on the recorded positions produces
+exactly the bytes its own search-based replacement would produce. Tools that
+modify the contents of files in an existing package (for example repackaging
+or transmutation tools) MUST update the fields or remove them.
+
+The fields describe occurrences of the UTF-8 encoding of the placeholder
```
chrisburr: This seems to be a messy edge case that I hadn't considered before an LLM noticed.

In Conda, non-UTF-8 placeholders are supported since https://github.com/conda/conda/pull/9946. In rattler this isn't the case and I didn't find any discussion around this. I did find this mamba issue that seems not to have been implemented: https://github.com/mamba-org/mamba/issues/2215

Now I've seen the feature, I think the non-UTF-8 encodings feature was a mistake to merge in Conda and at the least should have gone through a CEP (it was created before the CEP procedure but merged long after it was implemented). If it is going to be an officially supported feature of the conda format I think the encodings need to be standardised so this paths-offsets feature can be amended accordingly.

baszalmstra: If we store information about each occurrence in the file anyway, we might as well attach some additional metadata like the encoding.

For instance, another shortcoming of the current approach is listed in this issue: https://github.com/conda/rattler/issues/2342. Perhaps by adding more information, we could tackle these cases as well. (See, for instance, this proposed solution: https://github.com/conda/rattler/pull/2503)

chrisburr: What do you think of this?

`offsets` becomes a dictionary:

```json
{
  "_path": "lib/libexample.so",
  "path_type": "hardlink",
  "file_mode": "binary",
  "prefix_placeholder": "/opt/placeholder",
  "offsets": {
    "utf-8": [[64, 96], [200, 240, 300]]
  },
  "sha256": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592",
  "size_in_bytes": 4096
}
```

The allowed keys in offsets become the [`POPULAR_ENCODINGS` in conda](https://github.com/conda/conda/blob/9bf8c853352bebce196ee093aaf6300a2ccc20f0/conda/core/portability.py#L39). Probabaly define that the ranges `MUST NOT` overlap. In future special case handling for rust/go/c++ style strings could also be added with a different key (e.g. `utf-8-c++`).

baszalmstra: Yeah that could work! Or: 

```
{
  "_path": "lib/libexample.so",
  "path_type": "hardlink",
  "file_mode": "binary",
  "prefix_placeholder": "/opt/placeholder",
  "offsets": {
    { "ranges": [[64, 96], [200, 240, 300]], "encoding": "utf-8" }
  },
  "sha256": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592",
  "size_in_bytes": 4096
}
```

I think mine is easier to read, but yours is easier to parse.

While we are introducing "types" of replacements, we could also introduce one specifically for binary and one for text (e.g. `bin-utf-8` or `text-utf-8`). That would make typing easier, maybe.

chrisburr: > While we are introducing "types" of replacements, we could also introduce one specifically for binary and one for text (e.g. `bin-utf-8` or `text-utf-8`). That would make typing easier, maybe.

I think this still needs to be discriminated on `"file_mode": "binary"` so I don't see the benefit of this but I might have misunderstood? Unless you're leaning towards being able to support both text and binary substitution in the same file?

baszalmstra: > Unless you're leaning towards being able to support both text and binary substitution in the same file?

That was my thinking indeed. I don't see a use case for this now, but it would make it very easy to support that.

chrisburr: I don't think this kind of semantic change can rely on an optional key. I think it would need to be a Paths v2 format with new behaviour that makes the offsets mandatory and packages unavailable to older installers. I would even consider advocating for a new virtual package so the builds are ignored by older installers that don't support the feature.

I also think the utility of mixing binary/text substitution is hypothetical, there are known common-ish use cases where non-UTF8 placeholders and non-null terminated strings are needed. I don't think I've ever come across a need for mixing text and binary mode, and even if there was I wonder if it would be better served with a patch to the sources to be more Conda friendly.

cep-XXXX.md
```diff
@@ -0,0 +1,412 @@
...
+  conditions described below.
+
+Throughout this CEP, an _occurrence_ is a leftmost, non-overlapping match of
+the UTF-8 encoding of `prefix_placeholder` in the file contents, i.e. the
+matches found by a single left-to-right search: the same semantics as the
+search-based replacement in existing installers.
+
+`file_mode` normatively determines which of the two shapes of `offsets`
+applies. Implementations MUST NOT infer the shape from the JSON structure,
+since an empty list is valid for the text form and indistinguishable from an
```
jaimergp: Why would we have an empty list? Should that value be forbidden? edit: ok, I see below.

cep-XXXX.md
chrisburr: After further thought I've prepared CEP (https://github.com/conda/ceps/pull/180) to document the status quo which I think there is precedent for in past CEPs? It would probably then be good to have a follow up to decide what should actually be supported with prefix replacement as I clearly had a more simplistic view than what is actually in Conda these days.

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+substitution, and `shebang_length`, which marks a leading shebang line that
+installers transform under their own rules. Together they allow installers to
+perform prefix replacement without searching file contents, and to compute
+the exact size of a file after replacement without rendering it.
+
+## Motivation
+
+### Linking performance
+
+For every path entry with a `prefix_placeholder`, installers must read the
```
jaimergp, line 35:
```suggestion
For every `paths.json` entry with a `prefix_placeholder`, installers must read the
```

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+### Virtual filesystems
+
+An on-demand virtual filesystem that materializes conda environments lazily
+(such as the RattlerVFS work in [conda/rattler#2032][rattler-2032]) needs to
+answer `stat()` calls, i.e. report the final size of a file, before the file
+contents are ever read, and needs to serve reads of arbitrary ranges of the
+patched file directly from the original package data.
+
+Both needs reduce to knowing, without reading the file, where the placeholder
+occurs and how the file length changes. For binary files replacement preserves
```
jaimergp, line 58:
```suggestion
occurs and how the file size changes. For binary files replacement preserves
```

And same elsewhere. I find it less ambiguous to talk about size rather than length, because length makes me think of text and encodings which need to be defined.

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+[CEP 34](./cep-0034.md). The replacement behavior itself is documented in
+[conda/ceps#180][ceps-180]; this CEP does not change what replacement
+produces, only how occurrences are located. Two optional keys are added to
+each path entry:
+
+- `offsets: list[dict]`. Optional. A list of _offset groups_, each recording
+  where the placeholder occurs in the file contents under one encoding, as
+  stored in the package (i.e. before any replacement). It MUST NOT be present
+  unless `prefix_placeholder` is present.
+- `shebang_length: int`. Optional. The length in bytes of the file's first
```
jaimergp, line 78: How is a line defined under different encodings? For ASCII/UTF-8 is "until 0x0A is found", but for UTF-16 and above, that byte may be part of a bigger block.

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+
+- `offsets` is present,
+- `file_mode` is `text`, and
+- the file starts with the bytes `#!`.
+
+Note that its presence does not depend on whether the first line contains the
+placeholder (see Rationale).
+
+Its value MUST be `i + 1`, where `i` is the offset of the first newline byte
+(`0x0A`) in the file, or the size of the whole file when it contains no
```
jaimergp, line 154: Same here, what about UTF-16 and above?

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+`shebang_length` MUST be present if and only if all of the following hold:
+
+- `offsets` is present,
+- `file_mode` is `text`, and
+- the file starts with the bytes `#!`.
+
+Note that its presence does not depend on whether the first line contains the
+placeholder (see Rationale).
+
+Its value MUST be `i + 1`, where `i` is the offset of the first newline byte
```
jaimergp, line 153:
```suggestion
Its value MUST be `i + 1`, where `i` is the 0-indexed offset of the first newline byte
```

I assume?

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+MUST NOT carry `offsets`. In particular, producers MUST omit the fields for
+Windows launcher executables with an appended script archive (distlib
+"pyzzer" launchers), which conda and libmamba patch with a length-changing
+shebang rewrite.
+
+### Installer behavior
+
+Installers MAY use `offsets` to locate placeholder occurrences instead of
+searching the file contents. When the field is absent, installers MUST locate
+occurrences by searching, as they do today.
```
jaimergp, line 197:
```suggestion
occurrences by searching, as per #180.
```

Clearer this way?

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+  under the same encoding. Within each affected C string, occurrences are
+  replaced, the remaining bytes of the string shift towards the start
+  accordingly, and the gap before the original NUL terminator (or the end of
+  file, for an unterminated final string) is filled with zero bytes, so the
+  overall file length is unchanged.
+
+If the shape of a group's `ranges` does not match `file_mode`, a group has an
+unrecognized `encoding` or unrecognized members, or the values are
+inconsistent with the file (out of range, unsorted, overlapping, or the
+encoded placeholder is not present at a recorded position), installers SHOULD
```
jaimergp, line 231: "Presence" is ambiguous here. Being a substring in the range would qualify as present. I'd pose the range MUST exactly match the placeholder and any other case is considered a faulty range that should fall back or error out.

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+4. a shebang file with no trailing newline, where `shebang_length` equals the
+   file size;
+5. a file whose only occurrence is in the shebang line, where `offsets` is the
+   empty list;
+6. multiple occurrences within one shebang line;
+7. a shebang line longer than the kernel limit that contains no occurrence of
+   the placeholder;
+8. a binary file whose final C string is unterminated at the end of the file;
+9. a binary file with occurrences under more than one encoding, installed by
+   an installer that covers all of them and by one that covers only UTF-8.
```
jaimergp, line 278: Let's move these to an Appendix if they are informative.

cep-XXXX.md
jaimergp, line 78: Shall we use `shebang_bytes` for clarity?

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+  line, including the terminating newline. Its presence is REQUIRED under the
+  conditions described below.
+
+Each offset group MUST be a dictionary with exactly two keys:
+
+- `encoding: str`. One of `utf-8`, `utf-16-le`, `utf-16-be`, `utf-32-le`, or
+  `utf-32-be`: the encodings replaced by existing installers.
+- `ranges: list[int] | list[list[int]]`. The byte offsets of the occurrences
+  under that encoding. Its shape depends on `file_mode`, as described below,
+  and it MUST NOT be empty.
```
jaimergp, line 88: I'd fold this under the first item of the list above, as subitems. Or even better, introduce these sections:

```
## Specification

### Schema additions

#### `offsets`

##### Text offsets

##### Binary offsets

#### `shebang_length`

### Tooling behavior

...
```

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+user whose patched shebang stays under the kernel limit and breaks only for
+those with long target prefixes: the same latent, environment-dependent
+failure class that placeholder padding in `conda-build` exists to suppress,
+and one that is invisible to tests that do not use an over-long prefix. With
+the chosen layout the same mistake leaves the placeholder in the first line
+and fails on the first execution. Installers would also have to filter every
+listed offset against the region boundary on every install, instead of
+splicing the list uniformly.
+
+### Making text offsets relative to the end of the shebang line
```
jaimergp, line 423: Do we need to explicitly state that changes in the shebang length as a byproduct of replacement MUST result in the tool adjusting that difference for the absolute offsets?

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+> only when, they appear in all capitals, as shown here.
+
+[RFC2119]: https://datatracker.ietf.org/doc/html/rfc2119
+
+## Abstract
+
+This CEP adds two optional fields to the path entries of `./info/paths.json`
+(see [CEP 34](./cep-0034.md)): `offsets`, which records, per encoding, the
+byte positions at which the prefix placeholder can be replaced by plain
+substitution, and `shebang_length`, which marks a leading shebang line that
```
jaimergp, line 26:
```suggestion
substitution, and `shebang_length`, which marks the byte size of a leading shebang line that
```

or similar

cep-XXXX.md
```diff
@@ -0,0 +1,499 @@
...
+explicit upgrade error rather than corrupting files. conda and libmamba
+enforce `paths_version` today; rattler does not, and that gap should be
+closed well before any bump is proposed (see [conda/ceps#180][ceps-180]).
+
+## References
+
+- [CEP 34 - Contents of conda packages](./cep-0034.md)
+- [conda/ceps#180 - Prefix replacement at install time (draft)][ceps-180]
+- [conda/rattler#2032 - Implementation offsets in paths.json (original PR)](https://github.com/conda/rattler/pull/2032)
+- [conda/rattler#2565 - Implementation offsets in paths.json (follow-up PR)](https://github.com/conda/rattler/pull/2565)
```
jaimergp, line 483:
```suggestion
- [conda/rattler#2032 - Implementation of offsets in paths.json (original PR)](https://github.com/conda/rattler/pull/2032)
- [conda/rattler#2565 - Implementation of offsets in paths.json (follow-up PR)](https://github.com/conda/rattler/pull/2565)
```

?

jaimergp (changes requested):
Pretty solid. I consulted Gemini 3.6 Flash, reviewed its output and condensed the feedback in smaller suggestions. I only have some suggestions for clarity around edge cases; and a (non-blocking) proposal to reorganise the sections.

PR #183, opened 2026-07-24
Add CEP: subdirs and virtual packages for iOS and Android

wolfv:
Defines the ios-*/iossimulator-*/android-* subdirs and the __ios and __android virtual packages, mapping the PyPI wheel-tag model (PEP 730, PEP 738) onto conda's subdir + virtual-package machinery.

## Checklist for submitter

- [ ] I am submitting a new CEP: [Put your title here](#link-to-markdown-preview-in-branch).
  - [ ] I am using the CEP template by creating a copy `cep-0000.md` named `cep-XXXX.md` in the root level.

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -56,9 +56,7 @@ Two facts about these platforms shape the design:
 
 ### Subdirectories
 
-This CEP defines the following subdirs. Each subdir is single-architecture by construction and uses a
-single `<os>-<arch>` separator dash, so that tools which split a subdir on the first (and only) `-`
-continue to work unchanged.
+This CEP defines the following subdirs. As per CEP 26, each subdir is single-architecture and follows the `<os>-<arch>` syntax.
 
 #### iOS
 
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,171 @@
+# CEP XXXX - Subdirectories and virtual packages for iOS and Android
+
+<table>
+<tr><td> Title </td><td> Subdirectories and virtual packages for iOS and Android </td>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> Jul 24, 2026</td></tr>
+<tr><td> Updated </td><td> Jul 24, 2026</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> https://github.com/conda/rattler (feat/ios-android-subdirs) </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+  "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+  described in [RFC2119][RFC2119] when, and only when, they appear in all capitals, as shown here.
+>
+> More specifically, violations of a MUST or MUST NOT rule MUST result in an error. Violations of the
+  rules specified by any of the other all-capital terms MAY result in a warning, at discretion of the
+  implementation.
+
+## Abstract
+
+This CEP defines a set of conda subdirectories (`subdir`) for building and distributing packages that
+target Apple's iOS and Google's Android, together with two new virtual packages, `__ios` and
+`__android`, that carry the minimum-OS-version compatibility axis. The design maps the PyPI wheel-tag
+model for these platforms ([PEP 730] for iOS, [PEP 738] for Android) onto conda's existing
+subdir + virtual-package machinery, so that version compatibility is resolved by the ordinary solver
+in the same way as `__osx` and `__glibc` today. The goal is a single, cross-implementation set of
+names so that conda, mamba, rattler, and the wider ecosystem agree on how mobile packages are
+described.
+
+## Motivation
+
+conda has no standard way to describe packages built for mobile operating systems. Both Python
+(via [PEP 730] and [PEP 738]) and the wider native toolchain ecosystem now support iOS and Android as
+first-class cross-compilation targets, and tools such as pixi and rattler-build are able to produce
+packages for them. Without an agreed naming scheme, each implementation would be free to invent its
+own subdir tokens (`ios_arm64` vs `ios-arm64`, `ios-sim-arm64` vs `iossimulator-arm64`) and its own
+way of encoding the minimum-OS version, producing packages that cannot be described consistently or
+consumed across tools. That fragmentation is exactly the class of problem this CEP exists to prevent.
+
+Two facts about these platforms shape the design:
+
+1. **They are cross-compilation-only targets.** conda tooling does not run *on* iOS or Android as a
+   build host; packages are always produced by cross-compiling from another platform. The subdir and
+   virtual packages must therefore be expressible for a *target* platform, independent of the host
+   the solver runs on.
+
+2. **PyPI already encodes three axes in one wheel tag.** A wheel tag such as
+   `ios_13_0_arm64_iphoneos` packs together the CPU architecture (`arm64`), the ABI / platform variant
+   (device `iphoneos` vs simulator `iphonesimulator`), and the minimum OS version (`13.0`). conda
+   traditionally splits these axes across the *subdir* (architecture and ABI) and *virtual packages*
+   (OS-version compatibility). This CEP applies the same split rather than inventing a monolithic tag.
+
+## Specification
+
+### Subdirectories
+
+This CEP defines the following subdirs. Each subdir is single-architecture by construction and uses a
+single `<os>-<arch>` separator dash, so that tools which split a subdir on the first (and only) `-`
+continue to work unchanged.
+
+#### iOS
+
+| Subdir               | Architecture | Platform variant              |
+| -------------------- | ------------ | ----------------------------- |
+| `ios-arm64`          | `arm64`      | Device (`iphoneos`)           |
+| `iossimulator-arm64` | `arm64`      | Simulator (`iphonesimulator`) |
+| `iossimulator-64`    | `x86_64`     | Simulator (`iphonesimulator`) |
+
+The device/simulator distinction is a genuine ABI split: a binary built for the simulator cannot run
+on a device and vice versa, even for the same architecture. Rather than introduce a second dash
+(which would break subdir-splitting tools), the variant is folded into the OS token: device builds use
+`ios`, simulator builds use `iossimulator`. There is no `ios-64` (Apple ships no x86_64 iOS *device*).
+
+#### Android
+
+| Subdir              | Architecture | Android ABI     |
+| ------------------- | ------------ | --------------- |
+| `android-aarch64`   | `aarch64`    | `arm64-v8a`     |
+| `android-armv7a`    | `armv7a`     | `armeabi-v7a`   |
+| `android-64`        | `x86_64`     | `x86_64`        |
+| `android-32`        | `x86`        | `x86`           |
+
+The `armv7a` architecture token denotes Android's `armeabi-v7a` ABI. It is deliberately distinct from
+`armv7l` (used by `linux-armv7l`): both are 32-bit ARMv7, but `armeabi-v7a` uses Android's softfp
+calling convention and links against Bionic rather than glibc, so binaries are not interchangeable
+between the two.
+
+#### Android is not Linux
+
+Although Android runs a Linux kernel, `android-*` subdirs MUST NOT be treated as Linux subdirs.
+`linux-*` packages express their C-library requirement through the `__glibc` virtual package, and
+Android links against Bionic, which cannot satisfy a glibc constraint. Treating Android as Linux
+would allow the solver to install `linux-*` packages whose libc requirement is silently violated.
+iOS and Android subdirs ARE Unix subdirs for the purpose of the `__unix` virtual package.
+
+### Virtual packages
+
+Two new virtual packages carry the minimum-OS-version axis that the subdir intentionally omits.
+
+#### The `__ios` virtual package
+
+- **Name:** `__ios`
+- **Version:** the *minimum supported iOS version* (the deployment target), e.g. `13.0`. This mirrors
+  the `ios_<major>_<minor>` component of a PEP 730 wheel tag.
+- A package that requires iOS 13.0 or newer depends on `__ios >=13.0`, exactly as an `osx-*` package
+  depends on `__osx >=<version>`.
+
+#### The `__android` virtual package
+
+- **Name:** `__android`
+- **Version:** the *minimum supported Android API level*, encoded as a version, e.g. `21`. This
+  mirrors the API-level component of a PEP 738 wheel tag such as `android_21_arm64_v8a`.
+- A package that requires API level 21 or newer depends on `__android >=21`.
+
+#### Provisioning
+
+When solving *for* an `ios-*`/`iossimulator-*` or `android-*` target subdir, an implementation MUST
+make the corresponding virtual package available:
+
+- For an `ios-*`/`iossimulator-*` target, provide `__ios` (and `__unix`).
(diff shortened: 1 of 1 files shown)


cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+calling convention and links against Bionic rather than glibc, so binaries are not interchangeable
+between the two.
+
+#### Android is not Linux
+
+Although Android runs a Linux kernel, `android-*` subdirs MUST NOT be treated as Linux subdirs.
+`linux-*` packages express their C-library requirement through the `__glibc` virtual package, and
+Android links against Bionic, which cannot satisfy a glibc constraint. Treating Android as Linux
+would allow the solver to install `linux-*` packages whose libc requirement is silently violated.
+iOS and Android subdirs ARE Unix subdirs for the purpose of the `__unix` virtual package.
```
isuruf: This feels wrong. This seems like a different `stdlib`. For eg:

```yaml
c_stdlib:
  - sysroot               # maybe rename to sysroot_glibc?
  - sysroot_bionic
```

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+names so that conda, mamba, rattler, and the wider ecosystem agree on how mobile packages are
+described.
+
+## Motivation
+
+conda has no standard way to describe packages built for mobile operating systems. Both Python
+(via [PEP 730] and [PEP 738]) and the wider native toolchain ecosystem now support iOS and Android as
+first-class target platforms, and tools such as pixi and rattler-build are able to produce packages
+for them. Without an agreed naming scheme, each implementation would be free to invent its
+own subdir tokens (`ios_arm64` vs `ios-arm64`, `ios-sim-arm64` vs `iossimulator-arm64`) and its own
```
jaimergp: Some of these examples are illegal as per CEP-26:

> Channel subdir names MUST either be the literal `noarch` or a string following the syntax
`{os}-{arch}`, where `{os}` and `{arch}` MUST only consist of lowercase ASCII letters and numbers.
Non-`noarch` subdirs MUST match this regex: `^[a-z0-9]+-[a-z0-9]+$`.


`ios_arm64` and `ios-sim-arm64`, in particular.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+way of encoding the minimum-OS version, producing packages that cannot be described consistently or
+consumed across tools. That fragmentation is exactly the class of problem this CEP exists to prevent.
+
+Two facts about these platforms shape the design:
+
+1. **The subdir describes a target platform.** These names identify the platform a package is *for*,
+   independently of the host that resolves or builds it. In practice packages are often produced by
```
```suggestion
   independently of the host that resolves or builds it. In practice packages for new platforms are often produced by
```

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+   cross-compiling from another host, but this CEP does not assume that — an implementation may just
+   as well run natively on the platform.
+
+2. **PyPI already encodes three axes in one wheel tag.** A wheel tag such as
+   `ios_13_0_arm64_iphoneos` packs together the CPU architecture (`arm64`), the ABI / platform variant
+   (device `iphoneos` vs simulator `iphonesimulator`), and the minimum OS version (`13.0`). conda
+   traditionally splits these axes across the *subdir* (architecture and ABI) and *virtual packages*
+   (OS-version compatibility). This CEP applies the same split rather than inventing a monolithic tag.
```
jaimergp: I think these two items are superfluous and already implied by the existing CEP corpus.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Specification
+
+### Subdirectories
+
+This CEP defines the following subdirs. Each subdir is single-architecture by construction and uses a
+single `<os>-<arch>` separator dash, so that tools which split a subdir on the first (and only) `-`
+continue to work unchanged.
```
jaimergp: This is already defined by CEP 26, so this is not part of this specification. Instead, you can say:

```suggestion
This CEP defines the following subdirs. As per CEP 26, each subdir is single-architecture and follows the `<os>-<arch>` syntax.
```

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+| Subdir               | Architecture | Platform variant              |
+| -------------------- | ------------ | ----------------------------- |
+| `ios-arm64`          | `arm64`      | Device (`iphoneos`)           |
+| `iossimulator-arm64` | `arm64`      | Simulator (`iphonesimulator`) |
+| `iossimulator-64`    | `x86_64`     | Simulator (`iphonesimulator`) |
+
+The device/simulator distinction is a genuine ABI split: a binary built for the simulator cannot run
+on a device and vice versa, even for the same architecture. Rather than introduce a second dash
+(which would break subdir-splitting tools), the variant is folded into the OS token: device builds use
+`ios`, simulator builds use `iossimulator`. There is no `ios-64` (Apple ships no x86_64 iOS *device*).
```
jaimergp: CEP 26 prevents double dashes, no need to argue that here.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+
+| Subdir              | Architecture | Android ABI     |
+| ------------------- | ------------ | --------------- |
+| `android-aarch64`   | `aarch64`    | `arm64-v8a`     |
+| `android-armv7a`    | `armv7a`     | `armeabi-v7a`   |
+| `android-64`        | `x86_64`     | `x86_64`        |
+| `android-32`        | `x86`        | `x86`           |
+
+The `armv7a` architecture token denotes Android's `armeabi-v7a` ABI. It is deliberately distinct from
+`armv7l` (used by `linux-armv7l`): both are 32-bit ARMv7, but `armeabi-v7a` uses Android's softfp
```
jaimergp: `armv7l` is not included in the table, so why mention it here? If anything, you can add it in Rejected Ideas (on why it is not included), but this is not part of the specification.

cep-XXXX.md
jaimergp: This should also be moved to Rejected Ideas.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+
+- For an `ios-*`/`iossimulator-*` target, provide `__ios` (and `__unix`).
+- For an `android-*` target, provide `__android` (and `__unix`).
+- `__glibc` MUST NOT be provided for `android-*` targets.
+
+The version MAY be detected from the host when an implementation runs natively on the platform.
+Otherwise (for example when cross-compiling) it SHOULD be taken from an explicit override:
+
+- `__ios` version: from the `CONDA_OVERRIDE_IOS` environment variable, otherwise a build-tool-supplied
+  default, otherwise `0` (matches any requirement).
```
jaimergp: This is not true. If `__ios==0` it will prevent any constraint from being satisfied.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+| Wheel-tag axis         | PyPI example             | conda location                        |
+| ---------------------- | ------------------------ | ------------------------------------- |
+| CPU architecture       | `arm64`, `x86_64`        | subdir arch (`-arm64`, `-64`)         |
+| ABI / platform variant | `iphoneos` / `arm64-v8a` | subdir os token (`iossimulator`, ...) |
+| Minimum OS version     | `ios_13_0`, `android_21` | `__ios` / `__android` virtual package |
+
+## Rationale and alternatives
+
+- **A single dash per subdir.** An alternative was to encode the simulator variant as a third dash
+  (`ios-simulator-arm64`). This breaks the widespread assumption that a subdir is `<os>-<arch>` and
```
jaimergp: Again, not an assumption, this is _the_ specification. No need to state it here.

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
...
+  splits on a single `-`. Folding the variant into the OS token (`iossimulator`) preserves that
+  assumption.
+- **Version in the subdir vs. a virtual package.** Encoding the minimum OS version in the subdir (as
+  PyPI does in the wheel tag) would multiply the number of subdirs and move compatibility resolution
+  out of the solver. Using `__ios`/`__android` keeps a small, fixed set of subdirs and reuses the
+  existing, well-understood `__osx`/`__glibc` mechanism.
+- **Reusing `linux-*` for Android.** Rejected: Bionic is not glibc, and reuse would let the solver
+  violate libc requirements silently (see [Android is not Linux](#android-is-not-linux)).
+
```
jaimergp, line 116: I'm missing some bits mentioned in [PEP 730's Rationale](https://peps.python.org/pep-0730/#development-landscape) here, namely:

- Why two ABIs for iOS, and why not simply `osx`
- Why no iPadOS
- Why `ios` and `iossimulator` instead of the classic `iphoneos` vendor string

cep-XXXX.md
```diff
@@ -0,0 +1,172 @@
+# CEP XXXX - Subdirectories and virtual packages for iOS and Android
+
+<table>
+<tr><td> Title </td><td> Subdirectories and virtual packages for iOS and Android </td>
```
```suggestion
# CEP XXXX - Subdirs and virtual packages for iOS and Android

<table>
<tr><td> Title </td><td> Subdirs and virtual packages for iOS and Android </td>
```

I don't think we refer to platforms as "subdirectories". "Subdir" has taken its own meaning in our ecosystem on its own.

jaimergp (changes requested):
A few editorial comments. My main observation is that the current draft includes redundant information already expressed in prior CEPs, so ends up explaining and justifying a few things that are not needed. Trimming those out will improve the readability and focus of the CEP.

Also, the Specification section should be as terse as possible. A couple passages there belong better in Rejected Ideas.

The ideas are otherwise sound and well reasoned, and it makes sense to cover this omission in our ecosystem. Thanks for bringing it up!

cep-XXXX.md
wolfv: Addressed in 9e16e31. The rationale now explains that device and simulator are incompatible ABIs even on arm64 and that macOS is a distinct platform despite also using Darwin; that iPadOS has no separate development ABI; and that conda folds PEP 730’s separate OS/ABI fields into `ios` and `iossimulator`, avoiding an iPhone-only name while conforming to CEP 26.

isuruf:
`linux-64` is actually `x86_64-linux-gnu` and we need a subdir for `x86_64-linux-android`. Using a `android-64` seems wrong here since lots of build scripts uses pattern matching like `case linux-*` or `linux` as a preprocessing selector.

Maybe subdirs should be extended to three terms instead.

wolfv:
I think `linux_android` would look ugly but it would possibly work OK with the current tools. Personally I don't mind `android-xxx`...

isuruf:
I'm suggesting something like `x86_64-linux-android` as the subdir and let `linux-64` be an alias for `x86_64-linux-gnu`

cep-XXXX.md
```diff
@@ -0,0 +1,129 @@
...
+| `ios-arm64`          | `arm64`      | Device (`iphoneos`)           |
+| `iossimulator-arm64` | `arm64`      | Simulator (`iphonesimulator`) |
+| `iossimulator-64`    | `x86_64`     | Simulator (`iphonesimulator`) |
+
+#### Android
+
+| Subdir            | Architecture | Android ABI   |
+| ----------------- | ------------ | ------------- |
+| `android-aarch64` | `aarch64`    | `arm64-v8a`   |
+| `android-armv7a`  | `armv7a`     | `armeabi-v7a` |
```
freakboy3742, line 55: Is there any benefit to diverging from the Android ABI naming here? At least in principle, there could be other 64-bit ARM ABIs in future; is there some benefit to using `aarch64` instead of `arm64-v8a` and `armv7a` instead of `armeabi-v7a`?

freakboy3742 (commented):
I'm the author of PEP 730 and CPython iOS/Android platform delegate; I'll admit to not being fully across the implications of this specification, but the details described don't appear too surprising based on what knowledge I do have. 

Tagging @mhsmith (PEP 738 author) for visibility.

cep-XXXX.md
mhsmith: This follows the architecture naming of the [existing conda platforms](https://github.com/conda/conda/blob/e5cf1f1172c6bc7511497148b7c6e6b88f1762d2/conda/base/constants.py#L87-L106) (hence x86 is simply "32" and "64"), which is probably more useful than following the Android names. In hindsight, I should probably have done the same when establishing the Android wheel tag format, and made it match the Linux architecture names, which would have saved me a bit of work in auditwheel.

There could theoretically be other ABI variants of ARM64 in the future, but I think Google will probably handle them the same way as they did with NEON instructions on ARM32: declare that they're required on all devices newer than a certain Android version. In that case, there would be no need for a separate subdir; packages that are affected can simply require a minimum Android version.

[ARM32 discussion moved below]

cep-XXXX.md
```diff
@@ -0,0 +1,129 @@
...
+  solver. `__ios` and `__android` reuse conda's existing version-compatibility mechanism.
+
+## Rejected ideas
+
+- **`ios-simulator-*` subdirs.** A second dash does not conform to CEP 26. The simulator variant is
+  therefore folded into the OS token.
+- **An `ios-64` subdir.** Apple does not provide an x86_64 iOS device ABI.
+- **Using `armv7l` for Android.** Android's `armeabi-v7a` uses the softfp calling convention and
+  Bionic, unlike the ABI represented by `linux-armv7l`. The `armv7a` token keeps these incompatible
+  targets distinct.
```
mhsmith, line 111: Even if it does use the softfp calling convention, it is still armv7l (ARM version 7, little-endian), and the kernel reports it as such. The incompatible targets can be expressed by having different OS names: `android-armv7l` vs `linux-armv7l`.

In any case, the Android 32-bit platforms aren't officially supported by Python, so I'd suggest leaving them out for now.

mhsmith:
> I'm suggesting something like `x86_64-linux-android` as the subdir and let `linux-64` be an alias for `x86_64-linux-gnu`

As [discussed above](https://github.com/conda/ceps/pull/183#discussion_r3650046799), the `{os}-{arch}` format is already established by CEP 26, and I don't think this is a good enough reason to reopen that. Any build scripts with code like `case linux-*` will need to be updated to accommodate Android, in the same way as Python code that checks for `sys.platform == "linux"`.

cep-XXXX.md
wolfv: Thanks! As @mhsmith says, the `<arch>` token follows conda's existing subdirs (`linux-aarch64`, `linux-armv7l`, `linux-64`, `osx-arm64`) rather than Android's ABI names. I added a Rationale bullet spelling this out. It also notes that `arm64-v8a` and `armeabi-v7a` would be illegal under CEP 26, and that a future ARM64 ABI variant would most likely come with a minimum Android version, which `__android` already expresses.

cep-XXXX.md
wolfv: Thanks, you are right that this is still ARMv7 little-endian, so I renamed it to `android-armv7l` and let the OS token carry the ABI distinction. I kept the 32-bit subdirs rather than dropping them: conda channels ship C, C++ and Rust libraries, not only CPython, and `armeabi-v7a` is still common on Android TV, Wear OS and other low-end devices. The spec now notes that build tools MAY skip the 32-bit subdirs, and mentions that CPython does not support them.

cep-XXXX.md
wolfv: I considered this. Modeling Bionic as another `c_stdlib` inside `linux-*` would put binaries that cannot run on the same device into one subdir, and correct resolution would then rely on every package carrying an exact stdlib constraint. Android also differs in dynamic linker, userspace ABI, and platform APIs, so it gets its own OS token, the same way PEP 738 treats it as a platform separate from Linux. This is captured under "Using `linux-*` for Android" in Rejected Ideas.

wolfv:
Following up on the `x86_64-linux-android` suggestion: I recorded it in Rejected Ideas. The `<os>-<arch>` syntax is fixed by CEP 26, and as @mhsmith notes, scripts matching `linux-*` need to learn about Android regardless of the spelling. The latest commit also renames `android-armv7a` to `android-armv7l` and adds rationale for the architecture token names. @jaimergp, I believe all your editorial comments are addressed now, so this should be ready for another look.

PR #184, opened 2026-07-24
migrate to pixi-based pre-commit

pavelzw:
equivalent for https://github.com/conda/governance/pull/380.

- no fetching of random GitHub repositories which contain loose pinning that result in pip installations without lockfiles. Instead, manage the whole supply chain through `pixi.lock`
- no need for system-wide pre-commit installation, just run `pixi run lint`
- no dependence on pre-commit.ci. Instead, simple and locally reproducible CI through GitHub actions and pixi

`pre-commit.ci - pr` failing makes sense because after this PR, we'll run the checks with the `CI / Lint (pull_request)` job.

### DIFF
--- a/.github/workflows/bump-lockfile.yml
+++ b/.github/workflows/bump-lockfile.yml
@@ -1,0 +1,52 @@
+name: Update lockfiles
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 1 * * # monthly
+jobs:
+  pixi-update:
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
+        with:
+          run-install: false
+      - name: Update lockfiles
+        run: |
+          set -o pipefail
+          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+      - name: Create pull request using GH CLI
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          # 1. Configure git user
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+
+          # 2. Check if pixi.lock has actually changed
+          if [ -z "$(git status --porcelain pixi.lock)" ]; then
+            echo "No changes detected in pixi.lock. Skipping PR creation."
+            exit 0
+          fi
+
+          # 3. Stage, commit, and push changes
+          git checkout -b update-pixi
+          git add pixi.lock
+          git commit -m "Update pixi lockfile"
+          git push origin update-pixi --force
+
+          # 4. Create the Pull Request if needed
+          if gh pr view update-pixi &>/dev/null; then
+              echo "PR already exists. Updating existing PR branch suffices."
+              exit 0
+          fi
+          gh pr create \
+            --title "Update pixi lockfile" \
+            --body-file diff.md \
+            --base main \
+            --head update-pixi \
+            --label "pixi"
--- a/.github/workflows/bump-lockfile.yml
+++ b/.github/workflows/bump-lockfile.yml
@@ -13,12 +13,12 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Set up pixi
         uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
-        with:
-          run-install: false
       - name: Update lockfiles
         run: |
           set -o pipefail
-          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+          # use pixi-diff-to-markdown from previous pixi.lock
+          pixi install
+          pixi update --json --no-install | pixi run --as-is pixi-diff-to-markdown >> diff.md
       - name: Create pull request using GH CLI
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/bump-lockfile.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          # 1. Configure git user
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+
+          # 2. Check if pixi.lock has actually changed
+          if [ -z "$(git status --porcelain pixi.lock)" ]; then
+            echo "No changes detected in pixi.lock. Skipping PR creation."
+            exit 0
+          fi
```
jaimergp, line 33: We can do this first, before the `git config` calls, can't we?

.github/workflows/bump-lockfile.yml
```diff
@@ -0,0 +1,52 @@
...
+
+          # 3. Stage, commit, and push changes
+          git checkout -b update-pixi
```
jaimergp, line 36: Branch may exist already locally, right? Or is the default clone in `checkout` restricted enough?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  lint:
+    permissions:
+      contents: read
+    name: Lint
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
```
jaimergp:
```suggestion
    runs-on: ubuntu-slim
```

Given the timeout?

.github/workflows/bump-lockfile.yml
pavelzw: i would expect default clone is `main` only and also only a shallow clone

.github/workflows/bump-lockfile.yml
pavelzw: i don't care, this code was suggested by yourself :p https://github.com/conda/governance/pull/380#discussion_r3527846613

.pre-commit-config.yaml
```diff
@@ -1,70 +1,90 @@
...
-    # see https://pre-commit.com/#meta-hooks
-    hooks:
-      - id: check-hooks-apply
-      - id: check-useless-excludes
-  - repo: https://github.com/codespell-project/codespell
-    # see setup.cfg
-    rev: 2ccb47ff45ad361a21071a7eedda4c37e6ae8c5a  # frozen: v2.4.2
-    hooks:
+        name: check-merge-conflict
+        entry: pixi run check-merge-conflict --assume-in-merge
```
jaimergp, line 29: Are these executables provided by `pre-commit-hooks` or how?

pavelzw: exactly. this is the same that the previous `check-merge-conflict` pre-commit hook was doing: https://github.com/pre-commit/pre-commit-hooks/blob/9377601095448be1e7141feaaaa835714b53f6a5/.pre-commit-hooks.yaml#L70-L75, https://github.com/pre-commit/pre-commit-hooks/blob/3e8a8703264a2f4a69428a0aa4dcb512790b2c8c/setup.cfg#L39

.github/workflows/bump-lockfile.yml
```diff
@@ -0,0 +1,52 @@
...
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: 0 5 1 * * # monthly
+jobs:
+  pixi-update:
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
```
jaimergp:
```suggestion
    runs-on: ubuntu-slim
```

.github/workflows/bump-lockfile.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
+        with:
+          run-install: false
+      - name: Update lockfiles
+        run: |
+          set -o pipefail
+          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
```
jaimergp: Should we make `pixi-diff-to-markdown` a dependency in `pixi.toml` too? The point of the PR is to control dependencies more tightly, so not sure if we want a non-locked `exec` here.

pavelzw: i find it a bit odd as it wouldn't give you the same reproducibility+security guarantees since it would run `pixi-diff-to-markdown` _after_ the pixi environment just got updated which might pull in a new version. the `pixi exec` is explicit with it not being locked.

one can probably work around this by running something like `pixi install && pixi update --json --no-install | pixi run --as-is pixi-diff-to-markdown` but this seems quite complicated to me... but might be worth it, especially since we have a github token with `pull-requests: write` in this workflow.

pavelzw: [`2b1b82b` (this PR)](https://github.com/conda/ceps/pull/184/commits/2b1b82b3dcc659b198ae9535616c6fed45428841)

.github/workflows/bump-lockfile.yml
```diff
@@ -13,12 +13,12 @@ jobs:
...
       - name: Set up pixi
         uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
-        with:
-          run-install: false
       - name: Update lockfiles
         run: |
           set -o pipefail
-          pixi update --json | pixi exec pixi-diff-to-markdown >> diff.md
+          # use pixi-diff-to-markdown from previous pixi.lock
+          pixi install
```
jaimergp: Do we need a double install now? `run-install: false` was removed and this added, so I'm not following.

pavelzw:
```suggestion
```

jaimergp (approved):

jaimergp:
Thanks! Disabled the pre-commit ci app and updated the branch protection rules for the new job.

PR #188, opened 2026-08-19
Add CEP: Channel-provided virtual package plugins

hunger:
## Checklist for submitter

### DIFF
--- a/cep-XXXX.md
+++ b/cep-XXXX.md
@@ -1,0 +1,696 @@
+# CEP XXXX - Channel-provided virtual package detection plugins
+
+<table>
+<tr><td> Title </td><td> Channel-provided virtual package detection plugins </td></tr>
+<tr><td> Status </td><td> Draft </td></tr>
+<tr><td> Author(s) </td><td> Wolf Vollprecht &lt;wolf@prefix.dev&gt;<br/>Tobias Hunger &lt;tobias@prefix.dev&gt;</td></tr>
+<tr><td> Created </td><td> Aug 5, 2026</td></tr>
+<tr><td> Updated </td><td> Aug 12, 2026</td></tr>
+<tr><td> Discussion </td><td> NA </td></tr>
+<tr><td> Implementation </td><td> NA </td></tr>
+<tr><td> Requires </td><td> CEP 26, CEP 30, CEP 36, CEP 42 </td></tr>
+</table>
+
+> The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT",
+> "RECOMMENDED", "NOT RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as
+> described in [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) when, and only when, they appear in all capitals, as shown here.
+
+## Abstract
+
+[CEP 30](./cep-0030.md) standardizes a fixed set of virtual packages and makes detecting them an obligation of the client.
+A client can therefore only offer the names it was written to know about, and a channel shipping packages that depend on any other capability of the system has no way to have that capability detected.
+
+This CEP lets a channel declare, in its `repodata.json`, that a package it serves is a *detection plugin* for one or more virtual packages.
+A client resolving that channel installs the plugin into an isolated environment, runs it, and reads the virtual packages it reports from the plugin's standard output.
+The reported values then take part in the solve exactly as a client-detected virtual package does.
+
+The mechanism is deliberately narrow: a plugin answers only for names its channel advertised, is bounded in time and output, is cached with a mandatory expiry, and can be overridden or suppressed from the environment.
+Virtual package names are assumed to be unique across channels, and a channel introducing one SHOULD build its own name into it.
+It is also, unavoidably, arbitrary code execution, see [Security considerations](#security-considerations).
+
+## Motivation
+
+The set of virtual packages in CEP 30 is closed, and extending it requires a CEP per name (as [CEP 46](./cep-0046.md) did for `__cuda_arch`).
+That is the right process for capabilities the whole ecosystem shares, and the wrong one for everything else:
+
+- **Accelerator and interconnect stacks.** ROCm, oneAPI, Metal, Infiniband, TPUs and NPUs each need version *and* capability detection.
+  Each is relevant to the channels shipping builds against it and to nobody else.
+- **Site-specific capability.** An organization's internal channel may need to know about a license server, a filesystem, a kernel module or a CPU feature that no public channel cares about.
+- **Capabilities that change faster than clients release.** A new GPU generation is a new value for an existing name.
+  Today every client has to ship a new release to know about it, even though the channel that builds for it knows already.
+
+In all three cases whoever builds the packages has the knowledge necessary to get the needed information.
+Code that extracts the data from the machine is a typically small, easy to write, and easy to ship *as a conda package*.
+What is missing is a standard way for a channel to say "run this program to find out", and a defined way to format the answers.
+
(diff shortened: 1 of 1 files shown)


beckermr (commented):
I need to digest this more, but I like it, and it would solve an MPI-related plugin issue for conda-forge that came up with @minrk. 

The CEP needs to require that if a client caches plugins, the client MUST provide a user-facing way to clear the cache.

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+4. MUST capture standard output, which carries the report, separately from standard error, which carries diagnostics. Both streams count against the same output bound.
+
+A plugin MUST exit with status `0` on success.
+A non-zero exit MUST be treated as a failure.
+A plugin that exits `0` having written nothing to standard output MUST also be treated as a failure: writing the report to standard error by mistake is common enough to be worth naming rather than reading as "nothing detected".
+
+A client MUST bound a plugin run:
+
+- **Time.** A client MUST apply a timeout and MUST terminate a plugin that exceeds it.
+  The timeout SHOULD default to **5 seconds**.
```
minrk, line 215: 5 seconds is sensible for most use on owned hardware, but I would expect potential issues on CI, especially on Windows. I'm often amazed by how long `python3 -c "print('hi')"` can take on Window CI, sometimes well over 5 seconds. 5 seconds _will_ timeout sometimes on super slow CI, so I think anything that includes waiting for a process to warm up and produce output and exit shouldn't be this impatient by default. As annoying as it is to wait a long time for a hang, I think it's worse for slow CI to look like solve failures, so I think a 10-30 second default is probably safer.

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+  A client MAY let a user raise it, but MUST NOT allow any timeout longer than **60 seconds**: detection happens before a solve can begin, and an unbounded plugin is an unbounded hang in a tool a user is waiting on.
+- **Output.** A client MUST bound the combined amount of standard output and standard error it will read, and MUST treat exceeding that bound as a failure.
```
minrk, line 217: is that bound specified anywhere? Since exceeding it MUST cause a solve failure, it seems plugin authors need to know what number they cannot exceed. Seems like a default needs to be specified, at least, for the same reason a default is suggested for the timeout.

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+  The bound SHOULD scale with the number of virtual packages the plugin was registered for rather than being a single fixed number.
+
+Diagnostics a plugin wrote to standard error before the output bound was exceeded MUST be reported when a run fails.
+This is to enable users to diagnose and report issues they run into, while still preventing an unbounded diagnostic stream from exhausting client memory.
+
+A failed plugin MUST NOT abort the solve on its own.
+The names it was registered for are simply not available, which the solver then reports in the ordinary way for an unsatisfiable dependency.
+A client MUST report the failure.
```
minrk, line 225: maybe be more specific, since there are two possible antecedents for "the failure" (solve and plugin): this could be read as the client must report the failure (to solve), but I think should state that it must report every failure to execute a plugin, including diagnostic info, even if the solve succeeds.

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+Rejecting unknown keys would buy no safety: the plugin is arbitrary code the client has already run.
+
+Because the report is keyed by name, a duplicate verdict cannot be expressed, and clients therefore need no rule for one.
+
+### The contract
+
+The registration is a promise in both directions, and a client MUST check it before anything reaches the solver:
+
+- A plugin MUST give a verdict for **every** name its channel registered it for.
+  A name registered but absent from `virtual_packages` is a **contract violation**, not an absent capability — that is what `null` is for.
```
minrk, line 278: This does not state what the consequence of the contract failure is if it doesn't report all expected packages. It states that it must report it, but doesn't state what to do with the "incomplete" output. Does it JUST add a report of the failure and proceed with the solve with the reported virtual packages, ignore all virtual packages due to the contract violation, or abort the solve?

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+- A plugin MUST NOT report a name its channel did **not** register it for.
```
minrk, line 279: Thinking about upgrading, these present some challenges when it comes to changes in plugin outputs over time.

What happens, for example, when a plugin is updated and adds (or removes) a virtual package output? If the environment is cached from a previous run with an outdated plugin version, it may output less than all virtual packages. Here, this MUST be treated as a failure (still unspecified effect on the solve, I think), and the user-facing resolution is to update the plugin package, but it is not in a user-controlled env. What is the user's recourse? Seems like an explicit "flush plugin cache" or similar.

Should the channel be allowed to express version constraints on plugins in the repodata? e.g. plugin `virtual-package-x>=2.0: [__a, __b]`? Or is it assumed that plugins are all always up-to-date?

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+
+A `repodata.json` file MAY include a `virtual_package_plugins` key in its `info` dictionary.
+If present, it MUST be a dictionary mapping a **package name** to a **non-empty array of virtual package names**:
+
+```json
+{
+  "info": {
+    "subdir": "linux-64",
+    "virtual_package_plugins": {
+      "rocm-detect": ["__rocm"],
```
minrk, line 64: Since the solver looks for virtual package names, which then indicate which plugin packages it needs, should this index be reversed? It seems the client usually want to get a plugin for a requested virtual package, not the other way around.

I guess either way, both this and its reverse mapping are always going to be needed, anyway.

minrk (commented):
I like this, and will be thrilled if we can finally get our external mpi plugin working!

travishathaway:
Just FYI, there's been some more discussion about this over on conda-forge's Zulip for people discovering it only via GitHub: [#general > Virtual package plugin CEP](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Virtual.20package.20plugin.20CEP/with/620287162)

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+      "cuda-detect": ["__cuda", "__cuda_arch"]
+    }
+  }
+}
```
carterbox: I know it's not valid JSON, but I would add comments to make this example faster to grok.


```suggestion
{
  "info": {
    "subdir": "linux-64",
    "virtual_package_plugins": {
      "rocm-detect": ["__rocm"],  // names a package which provides __rocm
      "cuda-detect": ["__cuda", "__cuda_arch"]
    }
  }
}
```

cep-XXXX.md
```diff
@@ -0,0 +1,700 @@
...
+  Values that are not valid virtual package names MUST be ignored.
+  Plugins left speaking for an empty set of virtual packages, because every name they declared was
+  ignored, MUST themselves be ignored.
+- One plugin MUST register between 1 and 16 (inclusive on both ends) virtual packages.
+  This limits the amount of data one plugin can feed into a client.
+  The count is over the names the plugin declares, before any invalid name is ignored.
+  A plugin declaring a number of virtual packages outside that range is an error, and a client
+  encountering one MUST treat it as such.
+- Every name in every array MUST be a valid virtual package name (see [Virtual package names](#virtual-package-names)).
+  The virtual package name SHOULD contain the channel name, to make it less likely other channels will end up using the same virtual package name.
```
carterbox: You name this as best practice, but didn't do it in your example?

hunger:
We were asked to split this CEP into a plugin part and the channel-registration parts. @Hofer-Julian is preparing the split CEP proposal right now and we plan to handle one CEP per PR. I will close this proposal to make room for the new ones.

Thank you all for your contributions, we will have your feedback here reflected in the new proposals.

PR #191, opened 2026-09-18
CEP XXXX: Virtual Package Detector Protocol

Hofer-Julian:
This CEP defines how clients install and run virtual package detectors and use their results.

Together with the companion [Channel Registration of Virtual Package Detectors proposal](https://github.com/conda/ceps/pull/192), this supersedes [#188](https://github.com/conda/ceps/pull/188).

This CEP doesn't describe how detectors are discovered. The companion proposal adds channel discovery, priority and consent, and depends on this protocol.

## Checklist for submitter

- [x] I am submitting a new CEP: [Virtual Package Detector Protocol](https://github.com/Hofer-Julian/ceps/blob/virtual-package-detector-protocol/cep-XXXX.md).

- [ ] The CEP title header has been updated with the assigned number.


cep-XXXX.md
```diff
@@ -0,0 +1,249 @@
...
+Before execution or result-cache lookup, clients MUST resolve against current repodata and compute the [environment digest](#environment-digest), a fingerprint of all resolved package records.
+Clients MUST reuse a detector environment only if its digest matches; otherwise they must install the newly resolved packages.
+
+The **detector environment** contains the detector and its resolved dependencies.
+Installation MUST follow CEP 32 and [CEP 34](./cep-0034.md), with these restrictions:
+
+- The environment MUST be dedicated to detection, separate from the target environment, the client's environment and user working environments. The detector MUST NOT become a target-environment dependency.
+- Clients MUST NOT run any resolved package's pre-link, post-link or pre-unlink scripts.
+- Clients MAY skip byte compilation of `noarch: python` packages ([CEP 20](./cep-0020.md)); if performed, it SHOULD be bounded like activation.
+- Clients MUST NOT run detectors from incomplete installations and MUST prevent concurrent installations of the same digest from corrupting each other.
```
travishathaway, line 86: > Clients MUST NOT run detectors from incomplete installations

I'm curious when this situation is likely to arise? Is this the for corrupted environments?

cep-XXXX.md
```diff
@@ -0,0 +1,249 @@
...
+The detector MUST be resolvable for the **host platform**, the machine running the client (CEP 30's native platform), as defined by the registration source.
+
+### Resolution and installation
+
+Clients MUST resolve a [CEP 29](./cep-0029.md) MatchSpec containing only the detector name for the host platform, as defined by the registration source.
+A source MAY require a channel qualifier but MUST NOT add version or build constraints. Ordinary solver preferences determine the build.
+The only virtual packages available MUST be the client's own CEP 30 virtual packages, honoring `CONDA_OVERRIDE_*`. Detector results MUST NOT participate.
+
+Before execution or result-cache lookup, clients MUST resolve against current repodata and compute the [environment digest](#environment-digest), a fingerprint of all resolved package records.
+Clients MUST reuse a detector environment only if its digest matches; otherwise they must install the newly resolved packages.
```
travishathaway, line 78: Does this mean that clients are constantly checking for updates to these virtual package packages?

Hofer-Julian: There is a later section about caching :)

cep-XXXX.md
Hofer-Julian: Yeah, I was doubting if I should mention this at all.

isuruf:
Can you describe the relationship between this and the conda virtual plugin hookspec? See https://github.com/conda/conda/blob/main/conda/plugins/types.py#L187

Hofer-Julian:
> Can you describe the relationship between this and the conda virtual plugin hookspec? See https://github.com/conda/conda/blob/main/conda/plugins/types.py#L187

Do you mean here in the comments? Or do you think it makes sense to include it in the CEP?

The problem with conda plugins is that it only works for conda clients written in Python (which is only conda itself). Also the UX isn't great IMO. You have to remember to install the plugin, while the approach outlined in these two CEP proposals would just work

travishathaway:
> You have to remember to install the plugin, while the approach outlined in these two CEP proposals would just work

Just want to go on record stating that I think this is one of the best things about this proposal 🫶 ➕.

